PR #35, opened 2022-01-24
Inherit from dbt-spark for backward compatibility

### LINKED ISSUE #32: dbt-databricks adapter plugin  inherits from the dbt-spark plugin
### Describe the feature

Teams using dbt-spark and dbt-labs/spark-utils intending to migrate to using dbt-databricks cant do so because adapter name is now `databricks` instead of spark, so macros that are prefixed with `spark__` won't automatically be picked up for use.

The dbt-databricks adapter plugin depends on / inherits from the dbt-spark plugin. This would have the effect of telling dbt, "For adapter-dispatched macros, look for databricks__, then spark__, then default__

### Who will this benefit?
What kind of use case will this feature be useful for? Please be specific and provide examples, this will help us prioritize properly.

teams wanting to use dbt-databricks with dbt-labs/spark-utils macros

superdupershant:
Resolves #32
Fixes dbt-labs/spark-utils#23

Integration tests will be added to dbt-labs/spark_utils to ensure compatibility with dbt-databricks.

### Description

Inherit from dbt-spark in order to maintain compatibility with existing spark-utils and other macros.

### Checklist

- [ x] I have run this code in development and it appears to resolve the stated issue

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 databricks-sql-connector>=0.9.0
+dbt-spark ~= 1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 databricks-sql-connector>=0.9.0
-dbt-spark ~= 1.0.0
+dbt-spark~=1.0.0
--- a/dbt/adapters/databricks/__init__.py
+++ b/dbt/adapters/databricks/__init__.py
@@ -10,4 +10,6 @@ from dbt.include import databricks
 Plugin = AdapterPlugin(
     adapter=DatabricksAdapter,
     credentials=DatabricksCredentials,
-    include_path=databricks.PACKAGE_PATH)
+    include_path=databricks.PACKAGE_PATH,
+    dependencies=['spark']
+)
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1 +1,2 @@
 databricks-sql-connector>=0.9.0
+dbt-spark ~= 1.0.0
```
ueshin, line 2: Just a nit: maybe we want the definition format to be consistent.

```
databricks-sql-connector >= 0.9.0
dbt-spark ~= 1.0.0
```

or

```
databricks-sql-connector>=0.9.0
dbt-spark~=1.0.0
```

ueshin (commented):
Otherwise, LGTM.
Let me try something with this change in my local as well.

jtcohen6 (approved):
Looks right to me!

This change also enables you to remove methods + macros that are duplicated from / identical to those in `dbt-spark`, as you're able/interested, or inherit new functionality going forward. Let's keep talking and thinking through the right separation of concerns

ueshin (approved):

ueshin:
Thanks! merging.

PR #37, opened 2022-01-26
Close the connection properly

### LINKED ISSUE #34: ResourceWarning: Unclosed SSLSocket
### Describe the bug
Finishing commands like dbt run result in a message with an unclosed socket

### Steps To Reproduce
Run a command like dbt run

### Expected behavior
Close the socket?

 ResourceWarning: unclosed <ssl.SSLSocket fd=9, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('10.1.0.237', 47582), raddr=('40.74.30.80', 443)>

### System information
```
installed version: 1.0.0
   latest version: 1.0.0

Up to date!

Plugins:
  - databricks: 1.0.0
  - postgres: 1.0.0
  - redshift: 1.0.0
  - bigquery: 1.0.0
  - snowflake: 1.0.0
```

Ubuntu 21.10 in WSL2 (Windows 10)

Python 3.9.7

ueshin:
resolves #34

### Description

Closes the connection properly.

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -184,6 +184,12 @@ class DatabricksConnectionManager(SQLConnectionManager):
     def add_commit_query(self, *args, **kwargs):
         logger.debug("NotImplemented: add_commit_query")
 
+    def commit(self, *args, **kwargs):
+        logger.debug("NotImplemented: commit")
+
+    def rollback(self, *args, **kwargs):
+        logger.debug("NotImplemented: rollback")
+
     @classmethod
     def validate_creds(cls, creds, required):
         for key in required:
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -13,6 +13,11 @@ from dbt.events import AdapterLogger
 from dbt.utils import DECIMALS
 
 from databricks import sql as dbsql
+from databricks.sql.client import (
+    Connection as DatabricksSQLConnection,
+    Cursor as DatabricksSQLCursor,
+)
+from databricks.sql.exc import OperationalError
 
 logger = AdapterLogger("Databricks")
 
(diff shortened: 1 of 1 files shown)


dbt/adapters/databricks/connections.py
```diff
@@ -65,43 +70,45 @@ class DatabricksSQLConnectionWrapper(object):
...
-            # Handle bad response in the pyhive lib when
-            # the connection is cancelled
             try:
                 self._cursor.close()
-            except EnvironmentError as exc:
+            except OperationalError as exc:
                 logger.debug(
                     "Exception while closing cursor: {}".format(exc)
                 )
+        self._conn.close()
```
ueshin, line 101: I'm wondering whether this is also needed in dbt-spark.

ueshin: @jtcohen6 Is it fine in dbt-spark? I suspect it has the same issue without any notice.

jtcohen6: Came here because I was seeing the same unclosed socket error as in #34. Tried the fix in this PR, worked like a charm! Then I saw I was tagged...

This isn't an issue we've noticed with the `dbt-spark` connector, but it's not so surprising to me, different cursors/connectors behave differently. Even with the change in #35, which will allow `dbt-databricks` to inherit (and avoid reimplementing) common functionality from `dbt-spark`, I would expect these methods to be points of departure

ueshin: Got it. Thanks for confirming!

superdupershant (approved):
Nice cleanup!

ueshin:
Thanks! merging.

dbt/adapters/databricks/connections.py
danielhaviv: @jtcohen6 I can confirm I'm experiencing an issue with unclosed connections with dbt-spark and the fix is needed there as well

jtcohen6: @danielhaviv Thanks for opening the issue!

PR #40, opened 2022-01-28
Cleanup unnecessary codes by the inheritance of dbt-spark any more.

ueshin:
### Description

Cleans up codes and macros that are unnecessary after the inheritance of `dbt-spark` any more.


dbt/include/databricks/macros/materializations/snapshot.sql
```diff
@@ -48,25 +16,6 @@
...
-        {% for column in columns %}
-          `{{ column.name }}` {{ column.data_type }} {{- ',' if not loop.last -}}
-        {% endfor %}
-      );
-    {% endcall %}
-    {% endif %}
-{% endmacro %}
-
-
 {% materialization snapshot, adapter='databricks' %}
```
jtcohen6, line 19: It looks like you discovered this yourselves, but just for the sake of stating clearly:

Materializations don't natively support the same sort of "parent" adapter inheritance as other macros. I'm going to open an issue in `dbt-core` that addresses this. In the meantime, you can "mock" that inheritance (and avoid duplicated code) via:
```sql
{% materialization snapshot, adapter='databricks' %}
    {# this actually calls the spark version of the snapshot materialization #}
    {% set relations = materialization_snapshot_spark() %}
    {{ return(relations) }}
{% endmaterialization %}
```
This isn't a capability we document or recommend for end users, but [we use it in `dbt-snowflake` as well](https://github.com/dbt-labs/dbt-snowflake/blob/main/dbt/include/snowflake/macros/materializations/snapshot.sql). Any better approach we implement in a new `dbt-core` minor version will aim to include a non-breaking upgrade path / backwards compatibility for this syntax.

jtcohen6 (commented):
Very exciting to see this PR!

dbt/include/databricks/macros/materializations/snapshot.sql
ueshin: Thanks for letting me know. That's great to know.
Actually for this materialization, there are some updates for dbt-databricks from dbt-spark, so I think we still need to redefine here.

superdupershant (approved):

ueshin:
Thanks! merging.

PR #50, opened 2022-02-10
Bump up databricks-sql-connector to 1.0.1 and use the Cursor APIs.

ueshin:
### Description

Bumps up databricks-sql-connector to `1.0.1` and use the Cursor APIs.

- `Cursor.schemas()` to retrieve a list of schemas.

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -4,7 +4,7 @@ import re
 import time
 from typing import Any, Callable, ClassVar, Dict, Iterator, List, Optional, Sequence, Tuple
 
-import agate
+from agate import Table
 
 import dbt.exceptions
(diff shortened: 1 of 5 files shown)


dbt/adapters/databricks/connections.py
```diff
@@ -2,13 +2,17 @@
 from dataclasses import dataclass, field
 import re
 import time
-from typing import Any, ClassVar, Dict, Iterator, List, Optional, Sequence, Tuple
+from typing import Any, Callable, ClassVar, Dict, Iterator, List, Optional, Sequence, Tuple
+
+import agate
```
superdupershant, line 7: nit: up to you on style but `from agate import Table` to indicate we're only using a single type from the module.

superdupershant (approved):
Beautiful!

PR #55, opened 2022-02-19
Fix version to follow dbt-core.

ueshin:
### Description

Fixes version to follow dbt-core.


ueshin:
Need to rethink versioning, but I'd merge this now to unblock other PRs.

PR #57, opened 2022-02-23
Get the session_properties map to work.

superdupershant:
### Description

Allow session properties to be set as part of the profile configuration

### Checklist
Will update the README and docs in a separate commit.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -63,7 +63,6 @@ jobs:
 
       - name: Install python dependencies
         run: |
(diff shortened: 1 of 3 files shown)


ueshin (approved):
This is awesome!
Do we want to enable ansi mode by default or just mention in the doc?

superdupershant:
> This is awesome! Do we want to enable ansi mode by default or just mention in the doc?
Just in the doc for now.

ueshin:
I also checked the setting is passed to the cluster manually.
I'd merge this. Thanks!

PR #51, opened 2022-02-11
Setup integration tests on GitHub Actions.

ueshin:
### Description

Setups integration tests on GitHub Actions.

The integration tests work when the PR is merged, and scheduled time (5am UTC).

PR #58, opened 2022-03-04
Add integration test profiles against UC clusters.

ueshin:
### Description

Adds integration test profiles against UC clusters.

The http-path should be set by environment variables:

- `DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH`
- `DBT_DATABRICKS_UC_ENDPOINT_HTTP_PATH`

The initial catalog can be set by:

- `DBT_DATABRICKS_UC_INITIAL_CATALOG`

or "main" if not exist.

To run the tests:

- `tox -e integration-databricks-uc-cluster`
- `tox -e integration-databricks-uc-sql-endpoint`

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -10,6 +10,7 @@ import unittest
 from contextlib import contextmanager
 from datetime import datetime
 from functools import wraps
+from typing import Dict, Optional
 
 import pytest
 import yaml
@@ -144,93 +145,62 @@ class DBTIntegrationTest(unittest.TestCase):
     prefix = f'test{_runtime}{_randint:04}'
     setup_alternate_db = False
 
-    def databricks_cluster_profile(self):
-        return {
+    def _build_databricks_cluster_profile(
+        self,
+        http_path: str,
+        catalog: Optional[str] = None,
+        session_properties: Dict[str, str] = None,
+    ):
+        profile = {
             'config': {
                 'send_anonymous_usage_stats': False
             },
             'test': {
                 'outputs': {
-                    'cluster': {
+                    'dev': {
                         'type': 'databricks',
                         'host': os.getenv('DBT_DATABRICKS_HOST_NAME'),
-                        'http_path': os.getenv('DBT_DATABRICKS_CLUSTER_HTTP_PATH'),
+                        'http_path': http_path,
                         'token': os.getenv('DBT_DATABRICKS_TOKEN'),
                         'schema': self.unique_schema()
                     },
                 },
-                'target': 'cluster'
+                'target': 'dev'
             }
         }
+        if catalog is not None:
+            # TODO: catalog should be set as 'catalog' or 'database'
+            #       instead of using 'session_properties'
+            # profile['test']['outputs']['dev']['catalog'] = catalog
+            if session_properties is not None:
+                session_properties['databricks.catalog'] = catalog
+            else:
+                session_properties = {'databricks.catalog': catalog}
+        if session_properties is not None:
+            profile['test']['outputs']['dev']['session_properties'] = session_properties
+        return profile
+
+    def databricks_cluster_profile(self):
+        return self._build_databricks_cluster_profile(
+            http_path=os.getenv('DBT_DATABRICKS_CLUSTER_HTTP_PATH'),
+        )
 
     def databricks_uc_cluster_profile(self):
-        return {
-            'config': {
-                'send_anonymous_usage_stats': False
-            },
-            'test': {
-                'outputs': {
-                    'uc_cluster': {
-                        'type': 'databricks',
-                        'host': os.getenv('DBT_DATABRICKS_HOST_NAME'),
-                        'http_path': os.getenv('DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH'),
-                        'token': os.getenv('DBT_DATABRICKS_TOKEN'),
-                        'schema': self.unique_schema(),
-                        # TODO: catalog should be set as 'catalog' or 'database'
-                        #       instead of using 'session_properties'
-                        # 'catalog': os.getenv('DBT_DATABRICKS_UC_DEFAULT_CATALOG', 'main'),
-                        'session_properties': {
-                            'databricks.catalog': os.getenv('DBT_DATABRICKS_UC_DEFAULT_CATALOG', 'main'),
-                        },
-                    },
-                },
-                'target': 'uc_cluster'
-            }
-        }
+        return self._build_databricks_cluster_profile(
+            http_path=os.getenv('DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH'),
+            catalog=os.getenv('DBT_DATABRICKS_UC_INITIAL_CATALOG', 'main'),
+        )
 
     def databricks_sql_endpoint_profile(self):
-        return {
-            'config': {
-                'send_anonymous_usage_stats': False
-            },
-            'test': {
-                'outputs': {
-                    'endpoint': {
-                        'type': 'databricks',
-                        'host': os.getenv('DBT_DATABRICKS_HOST_NAME'),
-                        'http_path': os.getenv('DBT_DATABRICKS_ENDPOINT_HTTP_PATH'),
-                        'token': os.getenv('DBT_DATABRICKS_TOKEN'),
-                        'schema': self.unique_schema()
-                    },
-                },
-                'target': 'endpoint'
-            }
-        }
+        return self._build_databricks_cluster_profile(
+            http_path=os.getenv('DBT_DATABRICKS_ENDPOINT_HTTP_PATH'),
+        )
 
     def databricks_uc_sql_endpoint_profile(self):
-        return {
-            'config': {
-                'send_anonymous_usage_stats': False
-            },
-            'test': {
-                'outputs': {
-                    'uc_endpoint': {
-                        'type': 'databricks',
-                        'host': os.getenv('DBT_DATABRICKS_HOST_NAME'),
-                        'http_path': os.getenv('DBT_DATABRICKS_UC_ENDPOINT_HTTP_PATH'),
-                        'token': os.getenv('DBT_DATABRICKS_TOKEN'),
-                        'schema': self.unique_schema(),
-                        # TODO: catalog should be set as 'catalog' or 'database'
-                        #       instead of using 'session_properties'
-                        # 'catalog': os.getenv('DBT_DATABRICKS_UC_DEFAULT_CATALOG', 'main'),
-                        'session_properties': {
-                            'databricks.catalog': os.getenv('DBT_DATABRICKS_UC_DEFAULT_CATALOG', 'main'),
-                        },
-                    },
-                },
-                'target': 'uc_endpoint'
-            }
-        }
+        return self._build_databricks_cluster_profile(
+            http_path=os.getenv('DBT_DATABRICKS_UC_ENDPOINT_HTTP_PATH'),
+            catalog=os.getenv('DBT_DATABRICKS_UC_INITIAL_CATALOG', 'main'),
+        )
 
     @property
     def packages_config(self):
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -182,23 +182,31 @@ class DBTIntegrationTest(unittest.TestCase):
 
     def databricks_cluster_profile(self):
         return self._build_databricks_cluster_profile(
-            http_path=os.getenv('DBT_DATABRICKS_CLUSTER_HTTP_PATH'),
+            http_path=os.getenv(
+                "DBT_DATABRICKS_CLUSTER_HTTP_PATH", os.getenv("DBT_DATABRICKS_HTTP_PATH")
+            ),
         )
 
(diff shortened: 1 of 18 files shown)


tests/integration/base.py
```diff
@@ -150,15 +151,40 @@ def databricks_cluster_profile(self):
...
                         'token': os.getenv('DBT_DATABRICKS_TOKEN'),
                         'schema': self.unique_schema()
                     },
                 },
-                'target': 'dbsql'
+                'target': 'cluster'
+            }
+        }
+
+    def databricks_uc_cluster_profile(self):
```
allisonwang-db, line 166: I wonder if we can combine the UC cluster profile with the regular cluster profile by simply giving it a cluster HTTP path that points it to a UC-enabled cluster (with initial catalog set to different values).

allisonwang-db, line 166: I wonder if we can combine the UC cluster profile with the regular cluster profile by simply giving it a cluster HTTP path that points it to a UC-enabled cluster during the integration test (with the initial catalog set to different values).

tests/integration/base.py
```diff
@@ -150,15 +151,40 @@ def databricks_cluster_profile(self):
...
+                        'type': 'databricks',
+                        'host': os.getenv('DBT_DATABRICKS_HOST_NAME'),
+                        'http_path': os.getenv('DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH'),
+                        'token': os.getenv('DBT_DATABRICKS_TOKEN'),
+                        'schema': self.unique_schema(),
+                        # TODO: catalog should be set as 'catalog' or 'database'
+                        #       instead of using 'session_properties'
+                        # 'catalog': os.getenv('DBT_DATABRICKS_UC_DEFAULT_CATALOG', 'main'),
+                        'session_properties': {
+                            'databricks.catalog': os.getenv('DBT_DATABRICKS_UC_DEFAULT_CATALOG', 'main'),
```
allisonwang-db, line 183: We can change the default catalog by changing this session config? I thought we can only 1) set the static config for the initial catalog name before cluster startup or  2) use `USE CATALOG main` before running the query.

Also just curious is there any function we can call to validate cluster configuration before the integration test?  For example for UC profiles, we must have `SET spark.databricks.unityCatalog.enabled` returns `true`.

allisonwang-db, line 183: We can set the default catalog by changing this session config? I though we need to use "USE CATALOG main`?

tests/integration/base.py
ueshin: I'm afraid I don't think we should do that.
The UC cluster and non-UC cluster have some different behaviors, e.g., the UC cluster doesn't allow to write non-delta tables.
Having another profile makes switching tests easier.

tests/integration/base.py
ueshin: > We can change the default catalog by changing this session config? I thought we can only 1) set the static config for the initial catalog name before cluster startup or  2) use `USE CATALOG [catalog_name]` before running the query.

Yes, `databricks.catalog` in `session_properties` will change the current catalog similar to `USE CATALOG [catalog_name]`

> Also just curious is there any function we can call to validate cluster configuration before the integration test?  For example for UC profiles, we must have `SET spark.databricks.unityCatalog.enabled` returns `true`.

That's a good point. I'll add the tests.

ueshin: Added a test for `databricks_uc_cluster` profile: https://github.com/databricks/dbt-databricks/blob/d500f7c28409836f21b8e3a0fd40bd309fdce82f/tests/integration/current_catalog/test_current_catalog.py#L13-L19
The `databricks_uc_sql_endpoint` profile will raise an error because the UC SQL Endpoint doesn't have such a config.

Another test in the file is checking the current catalog, btw.

tests/integration/base.py
allisonwang-db: Thanks for the explanation. My intention is to reuse the code and the environment variable a bit more here. For example, if in the future we want to test the behavior when ANSI is enabled, then we might need to add more profiles. I am thinking if we can do something like this:
```python
def build_databricks_profile(self, session_properties: Dict[str, str] = None):
      profile = {
          'config': {
              'send_anonymous_usage_stats': False
          },
          'test': {
              'outputs': {
                  'default2': {
                      'type': 'databricks',
                      'host': os.getenv('DBT_DATABRICKS_HOST_NAME'),
                      'http_path': os.getenv('DBT_DATABRICKS_HTTP_PATH'),  # this can be cluster or SQL endpoint HTTP path
                      'token': os.getenv('DBT_DATABRICKS_TOKEN'),
                      'schema': self.unique_schema()
                  },
              },
              'target': 'default2'
          }
      }
     if session_properties is not None:
        ... 
     return profile

def databricks_uc_cluster_profile(self):
    return self.build_databricks_profile(session_properties=dict(...))
```
Or we can also pass in `http_path` as a param.

tests/integration/current_catalog/models/expected.sql
```diff
@@ -0,0 +1 @@
+SELECT '{{ env_var('DBT_DATABRICKS_UC_DEFAULT_CATALOG', 'main') }}' as cat
```
allisonwang-db, line 1: How about naming this `DBT_DATABRICKS_UC_INITIAL_CATALOG`?

tests/integration/current_catalog/test_current_catalog.py
```diff
@@ -0,0 +1,32 @@
...
+    def schema(self):
+        return "current_catalog"
+
+    @property
+    def models(self):
+        return "models"
+
+    def unity_catalog_enabled(self):
+        results = self.run_sql(
+            'SET spark.databricks.unityCatalog.enabled',
```
allisonwang-db, line 15: super nit: use double quote

allisonwang-db (commented):
Looks good! Just curious what are the integration tests that currently don't work with UC?

tests/integration/base.py
ueshin: Refactoring like `build_databricks_profile` sounds good.
I'll modify that.

However, reusing the environment `DBT_DATABRICKS_HTTP_PATH` is not a good idea, IMO.

Thinking of run the integration tests in the local environment manually, we will have to modify the environment variable as well as the profile name.

For example, we want to run `integration-databricks-cluster` and `integration-databricks-uc-cluster`, then:

```
export DBT_DATABRICKS_HTTP_PATH=<http-path for non-uc cluster>
tox -e integration-databricks-cluster
export DBT_DATABRICKS_HTTP_PATH=<http-path for uc cluster>
tox -e integration-databricks-uc-cluster
```

I'd prefer to set the environment variables in advance like in a profile file, then run the tox command in a row:

in a profile file:

```
export DBT_DATABRICKS_CLUSTER_HTTP_PATH=<http-path for non-uc cluster>
export DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH=<http-path for uc cluster>
```

then run tox commands:

```
tox -e integration-databricks-cluster
tox -e integration-databricks-uc-cluster
```

We don't need to take care of the environment variables anymore.

ueshin:
> Just curious what are the integration tests that currently don't work with UC?

`TestParquetAppend`, `TestParquetInsertOverwrite`, and `TestStoreFailuresParquet`.

tests/integration/current_catalog/test_current_catalog.py
ueshin: We need to apply black for test files once https://github.com/dbt-labs/dbt-spark/pull/297 is merged.
I intentionally left them to check the diff with the upstream test cases easily.

tests/integration/base.py
ueshin: Ah, reusing `DBT_DATABRICKS_HTTP_PATH` for delegation should be fine.
e.g., `os.getenv('DBT_DATABRICKS_CLUSTER_HTTP_PATH', os.getenv('DBT_DATABRICKS_HTTP_PATH'))`
Actually I'm going to do that at #51.
We can do that here, too.

ueshin: Let me apply the changes here.

tests/integration/base.py
```diff
@@ -143,43 +145,70 @@ class DBTIntegrationTest(unittest.TestCase):
...
-                'target': 'dbsql'
+                'target': 'dev'
             }
         }
+        if catalog is not None:
+            # TODO: catalog should be set as 'catalog' or 'database'
+            #       instead of using 'session_properties'
+            # profile['test']['outputs']['dev']['catalog'] = catalog
+            if session_properties is not None:
+                session_properties['databricks.catalog'] = catalog
```
allisonwang-db, line 176: What if the `session_properties` already have the 'databricks.catalog' field? How about throwing an exception for now if the catalog field is specified? For example:
```
if catalog is not None:
  throw Exception(f"Specifying catalog name is currently not supported. Please use session property 'databricks.catalog' = {catalog} instead.")
```

tests/integration/base.py
```diff
@@ -143,43 +145,70 @@ class DBTIntegrationTest(unittest.TestCase):
...
+    def databricks_cluster_profile(self):
+        return self._build_databricks_cluster_profile(
+            http_path=os.getenv(
+                "DBT_DATABRICKS_CLUSTER_HTTP_PATH", os.getenv("DBT_DATABRICKS_HTTP_PATH")
+            ),
+        )
+
+    def databricks_uc_cluster_profile(self):
+        return self._build_databricks_cluster_profile(
+            http_path=os.getenv(
```
allisonwang-db, line 192: I see this appears in multiple places. How about let's define them in a separate `config.py` file for all the configurations? For example
```python
# tests/integration/config.py

DBT_DATABRICKS_DEFAULT_HTTP_PATH = os.getenv("DBT_DATABRICKS_HTTP_PATH")
# A databricks cluster HTTP path with Unity Catalog enabled.
DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH = os.getenv("DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH", DBT_DATABRICKS_DEFAULT_HTTP_PATH)
...
```

tests/integration/base.py
```diff
@@ -143,43 +145,70 @@ class DBTIntegrationTest(unittest.TestCase):
...
+                "DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH", os.getenv("DBT_DATABRICKS_HTTP_PATH")
+            ),
+            catalog=os.getenv('DBT_DATABRICKS_UC_INITIAL_CATALOG', 'main'),
```
allisonwang-db, line 195: ditto

allisonwang-db (approved):
LGTM! Left a few comments on style.

tests/integration/base.py
ueshin: This is a temporary implementation.
The `catalog` should be in the profile as `catalog` or `database` as in the TODO comment.
I don't think we need to do it here because we don't have a profile with `session_properties` specified so far.
(We should rather remove the argument and add it later when it's needed?)

> What if the `session_properties` already have the 'databricks.catalog' field?

That's a good point. We should handle the case when we have the actual implementation. cc @superdupershant

tests/integration/base.py
ueshin: I'm afraid I don't think it's worth doing it.
We will see `os.getenv( ... )` anyway and I don't want to separate the test configurations here and there.
Actually there are already many places we need to modify to add a profile.

Maybe I can modify a bit more to deduplicate code. Let me see.

ueshin: Hmm, we could do something like:

```diff
@@ -147,7 +147,7 @@ class DBTIntegrationTest(unittest.TestCase):

     def _build_databricks_cluster_profile(
         self,
-        http_path: str,
+        http_path: Optional[str],
         catalog: Optional[str] = None,
         session_properties: Dict[str, str] = None,
     ):
@@ -160,7 +160,11 @@ class DBTIntegrationTest(unittest.TestCase):
                     'dev': {
                         'type': 'databricks',
                         'host': os.getenv('DBT_DATABRICKS_HOST_NAME'),
-                        'http_path': http_path,
+                        'http_path': (
+                            http_path
+                            if http_path is not None
+                            else os.getenv('DBT_DATABRICKS_HTTP_PATH')
+                        ),
                         'token': os.getenv('DBT_DATABRICKS_TOKEN'),
                         'schema': self.unique_schema()
                     },
@@ -182,9 +186,7 @@ class DBTIntegrationTest(unittest.TestCase):

     def databricks_cluster_profile(self):
         return self._build_databricks_cluster_profile(
-            http_path=os.getenv(
-                "DBT_DATABRICKS_CLUSTER_HTTP_PATH", os.getenv("DBT_DATABRICKS_HTTP_PATH")
-            ),
+            http_path=os.getenv("DBT_DATABRICKS_CLUSTER_HTTP_PATH")
         )

     def databricks_uc_cluster_profile(self):
```

but I'd not prefer this because it's not very clear that `http_path` will be delegated to `os.getenv('DBT_DATABRICKS_HTTP_PATH')` when the env var doesn't exist.

WDYT?

tests/integration/base.py
allisonwang-db: Sounds good!

PR #61, opened 2022-03-15
Make tox integration tests take additional arguments for pytest command.

ueshin:
### Description

Makes tox integration tests take additional arguments for `pytest` command.

For example:

```
tox -e integration-databricks-sql-endpoint -- -k test_dbt_empty
```


ueshin:
I think this is just good to go. I’d merge this now.

PR #63, opened 2022-03-23
Fix the dependencies to see 1.0.latest branches.

ueshin:
### Description

Fixes the dependencies to see `1.0.latest` branches.

### DIFF
--- a/dev_requirements.txt
+++ b/dev_requirements.txt
@@ -1,6 +1,10 @@
 # install latest changes in dbt-core
 # TODO: how to automate switching from develop to version branches?
-git+https://github.com/dbt-labs/dbt.git#egg=dbt-core&subdirectory=core
+git+https://github.com/dbt-labs/dbt.git@1.0.latest#egg=dbt-core&subdirectory=core
+
+# install latest changes in dbt-spark
+# TODO: how to automate switching from develop to version branches?
+git+https://github.com/dbt-labs/dbt-spark.git@1.0.latest#egg=dbt-spark
 
 freezegun==0.3.9
 pytest==6.0.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,2 @@
 databricks-sql-connector>=1.0.1
+dbt-spark~=1.0.0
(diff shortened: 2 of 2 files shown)


dev_requirements.txt
```diff
@@ -1,6 +1,10 @@
 # install latest changes in dbt-core
 # TODO: how to automate switching from develop to version branches?
-git+https://github.com/dbt-labs/dbt.git#egg=dbt-core&subdirectory=core
+git+https://github.com/dbt-labs/dbt.git@1.0.latest#egg=dbt-core&subdirectory=core
+
+# install latest changes in dbt-spark
+# TODO: how to automate switching from develop to version branches?
+git+https://github.com/dbt-labs/dbt-spark.git@1.0.latest#egg=dbt-spark
```
allisonwang-db, line 7: Just want to make sure this `1.0.latest` means installing the latest patch version (`1.0.x`)? Does it also include unreleased changes?

I am curious why `dbt-spark` is a *dev* requirement? When users install `dbt-databricks` package, do they also need to also install `dbt-spark`?

ueshin: It also includes unreleased changes in their repository https://github.com/dbt-labs/dbt-spark/tree/1.0.latest.

> I am curious why `dbt-spark` is a *dev* requirement?

That's a good question. Let me try to keep it in both `requirements.txt` and `dev_requirements.txt`.

> When users install `dbt-databricks` package, do they also need to also install `dbt-spark`?

Users usually use `pip` command to install the package, then it will also install `dbt-spark` according to `setup.py` settings. https://github.com/databricks/dbt-databricks/blob/7cc4f9d66baeb40591d2607619b14dbb313b2c14/setup.py#L57-L60

ueshin: > Let me try to keep it in both `requirements.txt` and `dev_requirements.txt`.

Seems like it's working. Let me revert the change in `requirements.txt`.

allisonwang-db (approved):
LGTM

ueshin:
Thanks! merging.

PR #69, opened 2022-03-30
Add database and schema in the log when executing schemas API.

ueshin:
### Description

Adds `database` and `schema` in the log when executing schemas API.

For example, in `dbt.log`:

```
... [debug] [ThreadPool]: On list_schemas: GetSchemas(database=None, schema=None)
```


ueshin:
Thanks! merging.

PR #70, opened 2022-03-31
Port testing framework changes.

ueshin:
### Description

Port testing framework changes from dbt-labs/dbt-spark#299 and dbt-labs/dbt-spark#314.


jtcohen6:
@ueshin You beat me to it! I was planning to open an issue for the testing framework migration in the `dbt-databricks` repo today :)

Draft docs (https://github.com/dbt-labs/docs.getdbt.com/pull/1263) for using the new framework: https://deploy-preview-1263--docs-getdbt-com.netlify.app/docs/contributing/testing-a-new-adapter

We've also got work in progress (https://github.com/dbt-labs/dbt-core/pull/4958) to convert and consolidate the test case that you ported in https://github.com/databricks/dbt-databricks/pull/62, so that you can inherit + run these test cases with a _lot_ less copy-pasted code.

ueshin:
@jtcohen6 Yes, we are keeping an eye on `dbt-spark` changes so that we don't loose the compatibility. :)

> We've also got work in progress (https://github.com/dbt-labs/dbt-core/pull/4958) to convert and consolidate the test case that you ported in https://github.com/databricks/dbt-databricks/pull/62, so that you can inherit + run these test cases with a _lot_ less copy-pasted code.

That's really good to know! Looking forward to it being merged.

tox.ini
```diff
@@ -29,7 +29,7 @@ deps =
 
 [testenv:integration-databricks-cluster]
 basepython = python3
-commands = /bin/bash -c '{envpython} -m pytest -v tests/specs/databricks-cluster.dbtspec {posargs}; ret=$?; [ $ret = 5 ] && exit 0 || exit $ret'
+commands = /bin/bash -c '{envpython} -m pytest -v --profile databricks_cluster tests/functional/adapter/test_basic.py {posargs}; ret=$?; [ $ret = 5 ] && exit 0 || exit $ret'
```
allisonwang-db, line 32: I like the `--profile` option! I am curious if we can change the integration tests to use this too.

allisonwang-db (approved):

tox.ini
ueshin: I think we can, in a separate PR, or we can contribute to `dbt-spark` if they want.

ueshin: Or https://github.com/dbt-labs/dbt-core/pull/4958 will provide the testing framework even for custom integration tests?

ueshin:
Thanks! merging.

tox.ini
jtcohen6: @ueshin Correct, our thinking is that this framework will provide the foundation for:
- basic test cases (minimum expectations for any dbt adapter)
- optional additional test cases for more advanced functionality (we'll be gradually adding, e.g. https://github.com/dbt-labs/dbt-core/pull/4958)
- your own custom integration/functional tests, which can tap into the same reusable utilities

As far as the `--profile` option in particular: It looks like this is something we added in `dbt-spark` (https://github.com/dbt-labs/dbt-spark/blob/main/tests/conftest.py), rather than the test utilities built into `dbt-core` (since the tests there only use a single Postgres profile). @gshank Do you see merit in standardizing this for all plugin maintainers, by moving `pytest_addoption` and `dbt_profile_target` into `dbt-core`? Or must they be defined in each plugin's own `conftest.py`?

jtcohen6: As far as the `--profile` option in particular: It looks like this is something we added in `dbt-spark` (https://github.com/dbt-labs/dbt-spark/blob/main/tests/conftest.py), rather than the test utilities built into `dbt-core` (since the tests there only use a single Postgres profile). @gshank Do you see merit in standardizing this for all plugin maintainers, by moving `pytest_addoption` and `dbt_profile_target` into `dbt-core`? Or must they be defined in each plugin's own `conftest.py`?

PR #71, opened 2022-04-01
Support Delta constraints

allisonwang-db:
### Description
This PR adds support for [Delta Constraints](https://docs.databricks.com/delta/delta-constraints.html) with dbt table models and incremental models. It supports two types of constraints: model level [Check constraint](https://docs.databricks.com/delta/delta-constraints.html#check-constraint) and column level [not_null](https://docs.databricks.com/delta/delta-constraints.html#not-null-constraint) constraint. 

Users can specify model-level and column-level constraints under the `meta` field of the model config. For example:
```yaml
# schema.yml
models:
  - name: my_model
    meta:
      constraints:
        - name: id_greater_than_zero
          condition: id > 0	
    columns:
      - name: id
      - name: name
        meta:
          constraint: not_null
```
Constraints specified will be created if the `persist_constraints` config is enabled (default: false):
```sql
-- my_model.sql
{{ config(
    materailized='table',
    persist_constraints=True
) }}
...
```

Note, Delta constraints are only available in Databricks Runtime 7.4 and above. So DBR 7.3 LTS won't support this feature.

### Checklist


allisonwang-db:
cc @jtcohen6

tests/integration/constraint_test/test_constraints.py
```diff
@@ -0,0 +1,129 @@
...
+    @property
+    def project_config(self):
+        return {
+            "config-version": 2,
+            "models": {
+                "test": {
+                    "+persist_docs": {
+                        "relation": True,
+                        "columns": True,
+                    },
```
ueshin, line 25: I'm just wondering if we need to enable them? I'm feeling it's ok to enable automatically if there is such configs in `schema.yml`.
I also feel a bit weird that the config named `persist_docs` enables the constraints.

jtcohen6: > I'm just wondering if we need to enable them? I'm feeling it's ok to enable automatically if there is such configs in `schema.yml`.

Fair, I think there's good arguments to be made in either direction. From my perspective, `meta` is usually for user's own metadata / annotation purposes, with no _functional_ effect in dbt, until/unless they opt into that via an extension macro/package or custom config.

> I also feel a bit weird that the config named `persist_docs` enables the constraints.

Agreed. Even if the mechanism to add these constraints is very very similar to `persist_docs`, I'd rather see this enabled via another config. Users (and therefore adapter plugins) can define whichever configs they like. All that would require is another macro call within the materialization, right alongside the calls to `persist_docs`:

https://github.com/databricks/dbt-databricks/blob/5d44dc64315a1a906c24b3223b074de4da60460f/dbt/include/databricks/macros/materializations/table.sql#L24-L26

https://github.com/databricks/dbt-databricks/blob/5d44dc64315a1a906c24b3223b074de4da60460f/dbt/include/databricks/macros/materializations/incremental/incremental.sql#L43-L45

allisonwang-db: Make sense! I will use a different config for constraints.

ueshin: We might also need it in `snapshot.sql`? https://github.com/databricks/dbt-databricks/blob/6dd2a8f7fcdafba56d788b42118fb43abe41a59e/dbt/include/databricks/macros/materializations/snapshot.sql#L96-L98

ueshin (approved):
LGTM.

> Note, Delta constraints are only available in Databricks Runtime 7.4 and above. So DBR 7.3 LTS won't support this feature.

If it doesn't support this feature with DBR 7.3 LTS, we should mention it in README to say:

- The `dbt-databricks` adapter has been tested against `Databricks SQL` and `Databricks runtime releases 9.1 LTS` and later.

or

- Delta constraints feature only works with `Databricks runtime releases 9.1 LTS` and later.

@superdupershant Which would you recommend, or other option?

ueshin:
Thanks! merging.

PR #72, opened 2022-04-01
Make internal macros use macro dispatch pattern

ueshin:
### Description

Makes internal macros use macro dispatch pattern for backward compatibility for users who override the equivalent macros in their projects.


jtcohen6:
Two thoughts:
- In cases where these macros are line-for-line identical with their counterparts in `dbt-spark`, I don't think you actually need to redefine them here. You can still call them within other macros in this plugin, including in materializations. The reason: All macros defined in plugins and in `dbt-core`'s global project are saved in the "internal"/"global" `dbt` namespace.
- In cases where you do need to redefine these macros, e.g. `file_format_clause` should return `delta` by default rather than `parquet`, we should use explicit dispatching ([docs](https://docs.getdbt.com/reference/dbt-jinja-functions/dispatch)) to deterministically define inheritance order. To that end, I think the right solution is to add `dispatch` calls and `spark__` prefixes in `dbt-spark`, so that you can then add `databricks__` prefixes in `dbt-databricks`. The preference order goes: Macros in the user's own project; built-in macros prefixed `databricks__`, then prefixed `spark__`, finally followed by macros prefixed `default__`. Could I ask you to open that as an issue/PR in `dbt-spark`? If the CLA is a blocker to contributing actual code, let's talk more and find a way to make that not so

dev_requirements.txt
```diff
@@ -5,7 +5,7 @@ git+https://github.com/dbt-labs/dbt-core.git#egg=dbt-tests-adapter&subdirectory=
 
 # install latest changes in dbt-spark
 # TODO: how to automate switching from develop to version branches?
-git+https://github.com/dbt-labs/dbt-spark.git#egg=dbt-spark
+git+https://github.com/dbt-labs/dbt-spark.git@d4c1498703b0c1c2a1a0e9913a46675f27a15b47#egg=dbt-spark
```
ueshin, line 8: This line should be reverted once https://github.com/dbt-labs/dbt-spark/pull/320 is merged.

dbt/include/databricks/macros/adapters.sql
```diff
@@ -1,19 +1,11 @@
-{% macro dbt_databricks_file_format_clause() %}
+{% macro databricks__file_format_clause() %}
```
allisonwang-db, line 1: Just curious, if a user has re-defined the `file_format_clause` in an existing project, for example
```
{% macro file_format_clause() %}
...
{% endmacro %}
```
Will this change (and the upstream change in dbt-spark) break the existing pipeline when the user upgrades the adapter version?

tox.ini
```diff
@@ -9,8 +9,8 @@ commands = /bin/bash -c '$(which black) --config black.ini --check .'
            /bin/bash -c '$(which mypy) --config-file mypy.ini --namespace-packages --explicit-package-bases dbt tests'
 passenv = DBT_* PYTEST_ADDOPTS
 deps =
-     -r{toxinidir}/requirements.txt
      -r{toxinidir}/dev_requirements.txt
+     -r{toxinidir}/requirements.txt
```
allisonwang-db, line 13: This is needed because we need to install the latest dbt-spark version first?

allisonwang-db (approved):
Looks good!

dbt/include/databricks/macros/adapters.sql
ueshin: According to @jtcohen6's explanation in https://github.com/databricks/dbt-databricks/pull/72#issuecomment-1086867013, the user's `file_format_clause` will be used.

> 1. Macros in the user's own project — they can always override the builtins, and they can do it by defining a macro named any of these: `file_format_clause`, `databricks__file_format_clause`, `spark__file_format_clause`, `default__file_format_clause`

tox.ini
ueshin: Yes.

ueshin:
Thanks! merging.

PR #73, opened 2022-04-06
[1.0] Make internal macros use macro dispatch pattern

ueshin:
### Description

Backport of #72.

Makes internal macros use macro dispatch pattern for backward compatibility for users who override the equivalent macros in their projects.
The old name macros are still available during 1.0.x releases, but will not be available in 1.1.0 release.

This fix will be available with `dbt-spark>=1.0.1`.


setup.py
```diff
@@ -40,7 +40,7 @@ def _get_plugin_version():
 
 package_name = "dbt-databricks"
 package_version = _get_plugin_version()
-dbt_core_version = "1.0.0"
+dbt_spark_version = "1.0.0"
```
ueshin, line 43: `dbt_spark_version` should be upgraded to `1.0.1rc1` or upper once `dbt-spark` bumps up the version, also in `requirements.txt`.

ueshin:
The tests "verify packages" will be fixed once `dbt-spark==1.0.1rc1` is published to pypi.

dbt/include/databricks/macros/adapters.sql
```diff
@@ -139,3 +87,30 @@
     {% endfor %}
   {% endif %}
 {% endmacro %}
+
+
+{# backward compatitibily #}
+{% macro dbt_databricks_file_format_clause() %}
```
allisonwang-db, line 93: I wonder if it's possible to add a test for this compatibility.

ueshin: The main macros such as `databricks__create_table_as` are still using the old name macros and there is a unit test `tests/unit/test_macros.py`.

jtcohen6: > The old name macros are still available during 1.0.x releases, but will not be available in 1.1.0 release.

This approach makes sense to me!

allisonwang-db (approved):

PR #81, opened 2022-04-18
Fix UC base tests to use test catalogs.

ueshin:
### Description

Fixes UC base tests to use test catalogs.

When porting testing framework, the catalog settings were lost.

### DIFF
--- a/tests/integration/base.py
+++ b/tests/integration/base.py
@@ -30,6 +30,9 @@ from dbt.events.functions import (
 from dbt.events import AdapterLogger
 from dbt.contracts.graph.manifest import Manifest
 
+from ..utils import build_databricks_cluster_profile
+
+
 logger = AdapterLogger("Databricks")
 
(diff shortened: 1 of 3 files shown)


tests/integration/base.py
```diff
@@ -30,6 +30,9 @@
 from dbt.events import AdapterLogger
 from dbt.contracts.graph.manifest import Manifest
 
+from ..utils import build_databricks_cluster_profile
```
allisonwang-db, line 33: super nit: how about using absolute import?

allisonwang-db (approved):

tests/integration/base.py
ueshin: Good point!
Unfortunately, `pytest` doesn't recognize the absolute import for some reason.
I'll investigate it later, but let me merge this now to unblock other investigations.
Thanks.

ueshin:
Thanks! merging.

PR #85, opened 2022-04-21
Set the upper bound on databricks-sql-connector.

ueshin:
### Description

Sets the upper bound on `databricks-sql-connector`.

`databricks-sql-connector 2.0.0` has been released, but it has a dependency issue.


allisonwang-db (approved):

ueshin:
Thanks! merging.

PR #87, opened 2022-04-30
Show full error info in debug log for databricks-sql-connector 2.0

ueshin:
### Description

Shows full error info in debug log for `databricks-sql-connector 2.0`.

The way to hold error info was changed in `databricks-sql-connector 2.0`.
We should show the full error info in the debug log.


ueshin:
Let me just merge this now. Thanks.

PR #89, opened 2022-05-09
first try at catalog support

superdupershant:
resolves #

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -38,6 +38,10 @@ class DatabricksCredentials(Credentials):
     session_properties: Optional[Dict[str, Any]] = None
     retry_all: bool = False
 
+    _ALIASES = {
+        'catalog': 'database',
+    }
+
     @classmethod
     def __pre_deserialize__(cls, data: Dict[Any, Any]) -> Dict[Any, Any]:
         data = super().__pre_deserialize__(data)
@@ -46,15 +50,11 @@ class DatabricksCredentials(Credentials):
         return data
 
     def __post_init__(self) -> None:
-        # spark classifies database and schema as the same thing
-        if self.database is not None and self.database != self.schema:
+        if self.database is not None and not(self.database.strip()):
             raise dbt.exceptions.RuntimeException(
-                f"    schema: {self.schema} \n"
                 f"    database: {self.database} \n"
-                f"On Spark, database must be omitted or have the same value as"
-                f" schema."
+                f"Invalid catalog name."
             )
-        self.database = None
 
     @property
     def type(self) -> str:
@@ -65,7 +65,7 @@ class DatabricksCredentials(Credentials):
         return self.host
 
     def _connection_keys(self) -> Tuple[str, ...]:
-        return "host", "port", "http_path", "schema"
+        return ("host", "database", "http_path", "schema", "session_properties")
 
 
 class DatabricksSQLConnectionWrapper(object):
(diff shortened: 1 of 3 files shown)


dbt/adapters/databricks/relation.py
```diff
@@ -1,12 +1,22 @@
...
 
-from dbt.exceptions import RuntimeException
-
+from dbt.adapters.base.relation import Policy
 from dbt.adapters.spark.relation import SparkRelation
 
 
+@dataclass
+class DatabricksIncludePolicy(Policy):
+    database: bool = False  # TODO: should be True
```
ueshin, line 9: If this is `True`, some integration tests fail. Need further investigation in the following PRs.

dbt/adapters/databricks/connections.py
```diff
@@ -47,15 +53,22 @@ def __pre_deserialize__(cls, data: Dict[Any, Any]) -> Dict[Any, Any]:
...
-        # spark classifies database and schema as the same thing
-        if self.database is not None and self.database != self.schema:
+        session_properties = self.session_properties or {}
+        if CATALOG_KEY_IN_SESSION_PROPERTIES in session_properties:
+            if self.database is None:
+                self.database = session_properties[CATALOG_KEY_IN_SESSION_PROPERTIES]
+                del session_properties[CATALOG_KEY_IN_SESSION_PROPERTIES]
+            else:
+                raise dbt.exceptions.RuntimeException(
+                    f"`catalog` or `database` and `{CATALOG_KEY_IN_SESSION_PROPERTIES}` "
```
allisonwang-db, line 63: Maybe we can add a test case for this? 

And what if I have a config like this:
```
{
  "database": "db",
  "catalog": "ct",
}
```
Will it throw exceptions?

dbt/adapters/databricks/connections.py
```diff
@@ -47,15 +53,22 @@ def __pre_deserialize__(cls, data: Dict[Any, Any]) -> Dict[Any, Any]:
...
+                    "in `session_properties` cannot both be set"
+                )
+        self.session_properties = session_properties
+
+        if self.database is not None and not (self.database.strip()):
```
allisonwang-db, line 68: Is this checking that the catalog name cannot be empty? How about changing the error message to be: `Invalid catalog name: {self.database}`?

tests/unit/test_adapter.py
```diff
@@ -78,6 +120,65 @@ def databricks_sql_connector_connect(
             )
             self.assertIsNone(connection.credentials.database)
 
+    def test_databricks_sql_connector_catalog_connection(self):
+        config = self._get_target_databricks_sql_connector_catalog(self.project_cfg)
+        adapter = DatabricksAdapter(config)
+
+        if LooseVersion(dbsql.__version__) < "2.0":
```
allisonwang-db, line 127: Can we mock the dbsql version in the test so that we can test both cases?

dbt/adapters/databricks/connections.py
ueshin: Yes, it throws an exception. `Got duplicate keys: (catalog) all map to "database"`.
Added the test and modified the error message here to follow it.

tests/unit/test_adapter.py
```diff
@@ -44,39 +45,192 @@ def _get_target_databricks_sql_connector(self, project):
...
+                        }
+                    },
+                    "target": "test",
+                },
+            )
+
+    def test_database_and_catalog_settings(self):
+        with self.assertRaisesRegexp(
+            dbt.exceptions.DbtProfileError,
+            'Got duplicate keys: \\(catalog\\) all map to "database"',
```
allisonwang-db, line 97: How about "Cannot specify both `catalog` and `database`"?

allisonwang-db (approved):
Looks good!

tests/unit/test_adapter.py
ueshin: That's from `dbt-spark` or `dbt-core`. I think we should follow the error message.

allisonwang-db: Sounds good!

ueshin:
Thanks! merging.

PR #90, opened 2022-05-10
Reject databricks-sql-connector 2.0.0

ueshin:
### Description

Reject databricks-sql-connector 2.0.0 as it had a dependency issue.


ueshin:
As this is trivial, I'd merge this now. Thanks.

PR #93, opened 2022-05-11
Fix DBTIntegrationTest to be more robust

ueshin:
### Description

Fixes `DBTIntegrationTest` to be more robust.

- Make `DBTIntegrationTest.prefix` `@property` to not share the same prefix in sub classes.
- Drop the schema with the suffix `_dbt_test__audit` created when `dbt test --store-failures` is run.


allisonwang-db (approved):
Good catch!

ueshin:
Thanks! merging.

PR #94, opened 2022-05-12
Use 3-level namespace when catalog is set.

ueshin:
### Description

Uses 3-level namespace when catalog is set.

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -6,11 +6,12 @@ from agate import Table
 from dbt.contracts.connection import AdapterResponse
 from dbt.adapters.base import AdapterConfig
 from dbt.adapters.base.relation import BaseRelation
-from dbt.adapters.databricks import DatabricksConnectionManager
-from dbt.adapters.databricks.relation import DatabricksRelation
-from dbt.adapters.databricks.column import DatabricksColumn
-
 from dbt.adapters.spark.impl import SparkAdapter
+
+from dbt.adapters.databricks.column import DatabricksColumn
+from dbt.adapters.databricks.connections import DatabricksConnectionManager
+from dbt.adapters.databricks.relation import DatabricksRelation
+from dbt.adapters.databricks.utils import undefined_proof
 
 
 @dataclass
@@ -25,6 +26,7 @@ class DatabricksConfig(AdapterConfig):
     tblproperties: Optional[Dict[str, str]] = None
 
 
+@undefined_proof
 class DatabricksAdapter(SparkAdapter):
 
     Relation = DatabricksRelation
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -120,11 +120,6 @@ class DatabricksAdapter(SparkAdapter):
 
         return relations
 
-    def get_relation(
-        self, database: Optional[str], schema: str, identifier: str
-    ) -> Optional[BaseRelation]:
-        return super(SparkAdapter, self).get_relation(database, schema, identifier)
-
     def parse_describe_extended(
         self, relation: DatabricksRelation, raw_rows: List[Row]
     ) -> List[DatabricksColumn]:
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -1,17 +1,29 @@
 from dataclasses import dataclass
-from typing import Optional, List, Dict, Tuple, Union
+import re
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
-from agate import Table
+from agate import Row, Table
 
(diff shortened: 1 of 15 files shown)


ewengillies:
thrilled to see work in this direction, I was just looking for this feature 3 days ago and happy to see its already in a PR.  Is there anything I can do to help?  Thanks :)

ueshin:
Let me separate some changes to another PR.

ueshin:
Now this is based on #98.

dbt/adapters/databricks/impl.py
```diff
@@ -67,3 +80,113 @@ def execute(
         finally:
             if staging_table is not None:
                 self.drop_relation(staging_table)
+
+    def list_relations_without_caching(
```
allisonwang-db, line 84: It would be nice to add a new method in dbt-spark that returns the catalog field to avoid duplicate code in the future (but we need to know the release schedule for dbt-spark).

dbt/adapters/databricks/impl.py
```diff
@@ -67,3 +80,113 @@ def execute(
...
+                type=rel_type,
+                information=information,
+                is_delta=is_delta,
+                is_hudi=is_hudi,
+            )
+            relations.append(relation)
+
+        return relations
+
+    def get_relation(
```
allisonwang-db, line 123: Seems like we don't need to override this method?
https://github.com/dbt-labs/dbt-spark/blob/bc9fc0baaacce28706d0542391704d9ea971cee4/dbt/adapters/spark/impl.py#L162-L168

dbt/include/databricks/macros/adapters.sql
```diff
@@ -143,3 +143,23 @@
...
+        {{ return(custom_database_name) }}
+    {%- endif -%}
+{%- endmacro %}
+
+{% macro databricks__make_temp_relation(base_relation, suffix) %}
+    {% set tmp_identifier = base_relation.identifier ~ suffix %}
+    {% set tmp_relation = base_relation.incorporate(path = {
+        "identifier": tmp_identifier,
+        "schema": None,
+        "database": None
```
allisonwang-db, line 161: If we don't specify `"database": None` here, what's the default value for it?

dbt/include/databricks/macros/materializations/snapshot.sql
```diff
@@ -58,7 +73,7 @@
 
       {{ adapter.valid_snapshot_target(target_relation) }}
 
-      {% set staging_table = spark_build_snapshot_staging_table(strategy, sql, target_relation) %}
+      {% set staging_table = databricks_build_snapshot_staging_table(strategy, sql, target_relation) %}
```
allisonwang-db, line 76: Not urgent but should we use the dispatch pattern here to make it backward compatible?

tests/integration/persist_constraints/test_persist_constraints.py
```diff
@@ -89,27 +91,25 @@ def test_incremental_constraints(self):
...
-
         # Insert a row into the seed model with an invalid id.
         self.run_sql_file("insert_invalid_id.sql")
         self.run_and_check_failure(
             model_name,
             err_msg="CHECK constraint id_greater_than_zero",
         )
         self.check_staging_table_cleaned()
-        self.run_sql(f"delete from {schema}.seed where id = 0")
+        self.run_sql("delete from {database_schema}.seed where id = 0")
```
allisonwang-db, line 101: Interesting, so if we use `{database_schema}` in the query then dbt will automatically change it to use the target database and target schema?

dbt/adapters/databricks/impl.py
ueshin: Yes, we can ask `dbt-spark` to make the change.
Let's have them here and work on it in a separate PR for now.

dbt/include/databricks/macros/adapters.sql
ueshin: The `tmp_relation` will hold the `database` of `base_relation`, that results with `database.identifier` for the temp relation name if `base_relation` has `database` field.

tests/integration/persist_constraints/test_persist_constraints.py
ueshin: Actually I made some changes in `tests/integration/base.py` to handle `database_schema`.

dbt/include/databricks/macros/adapters.sql
allisonwang-db: Make sense. Thanks for the explanation!

tests/integration/persist_constraints/test_persist_constraints.py
allisonwang-db: :+1: it's in `transform_sql`

allisonwang-db (approved):

ueshin:
Thanks! merging.

PR #96, opened 2022-05-17
Fix snapshot to always drop the staging table

ueshin:
### Description

We introduced Delta constraints at #71 and now that snapshot query could fail when it's violating the constraints.
In that case, the temporary view keeps existing because snapshot uses a permanent view and drop it later.

We should always drop them.

### DIFF
--- a/dbt/include/databricks/macros/materializations/snapshot.sql
+++ b/dbt/include/databricks/macros/materializations/snapshot.sql
@@ -50,6 +50,10 @@
       {% set build_sql = build_snapshot_table(strategy, model['compiled_sql']) %}
       {% set final_sql = create_table_as(False, target_relation, build_sql) %}
 
+      {% call statement('main') %}
+          {{ final_sql }}
+      {% endcall %}
+
   {% else %}
 
       {{ adapter.valid_snapshot_target(target_relation) }}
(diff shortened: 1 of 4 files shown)


dbt/include/databricks/macros/materializations/snapshot.sql
```diff
@@ -102,10 +106,6 @@
 
   {{ adapter.commit() }}
 
-  {% if staging_table is defined %}
-      {% do post_snapshot(staging_table) %}
```
allisonwang-db, line 106: The default implementation of `post_snapshot` seems to be a noop 
https://github.com/dbt-labs/dbt-core/blob/e7218d3e99837f0139fb7ecd367d3bdf1135a961/core/dbt/include/global_project/macros/materializations/snapshots/helpers.sql#L17-L23
And dbt-spark doesn't seem to implement it either. Where is the logic of post_snapshot defined?

allisonwang-db (approved):

dbt/include/databricks/macros/materializations/snapshot.sql
ueshin: `dbt-spark` does have the implementation. 

https://github.com/dbt-labs/dbt-spark/blob/main/dbt/include/spark/macros/materializations/snapshot.sql#L52-L54

allisonwang-db: :+1: This doesn't show up in Github's search result..

ueshin:
Thanks! merging.

PR #97, opened 2022-05-20
Use packaging.version instead of distutils.version

ueshin:
### Description

Uses `packaging.version` instead of `distutils.version` to suppress warnings.

```
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
```

`package` module is safe to use as it's included in the `dbt-core`'s dependencies.

https://github.com/dbt-labs/dbt-core/blob/e50678c91424142aa1a4995ee1d358dac6a6fdf8/core/setup.py#L65


allisonwang-db (approved):

ueshin:
Thanks! merging.

PR #98, opened 2022-05-23
Block taking jinja2.runtime.Undefined into DatabricksAdapter

ueshin:
### Description

Blocks taking `jinja2.runtime.Undefined` into `DatabricksAdapter`.

When calling the adapter method from macros, `jinja2.runtime.Undefined` can be passed to `DatabricksAdapter` that could cause unexpected behavior.

For example,

https://github.com/databricks/dbt-databricks/blob/b8152b697b003e02fb52e86f59b2bde4b492cc0a/dbt/include/databricks/macros/materializations/snapshot.sql#L33

The `model.database` is not defined and `jinja2.runtime.Undefined` will be used to call `adapter.check_schema_exists`.

PR #99, opened 2022-05-23
[1.0] Block taking jinja2.runtime.Undefined into DatabricksAdapter.

ueshin:
### Description

Backport of #98 

Blocks taking `jinja2.runtime.Undefined` into `DatabricksAdapter`.

When calling the adapter method from macros, `jinja2.runtime.Undefined` can be passed to `DatabricksAdapter` that could cause unexpected behavior.

For example,

https://github.com/databricks/dbt-databricks/blob/b8152b697b003e02fb52e86f59b2bde4b492cc0a/dbt/include/databricks/macros/materializations/snapshot.sql#L33

The `model.database` is not defined and `jinja2.runtime.Undefined` will be used to call `adapter.check_schema_exists`.


ueshin:
Thanks! merging.

PR #100, opened 2022-05-23
Avoid using Cursor.schema API when database is None

ueshin:
### Description

Avoids using `Cursor.schema` API when `database` is `None`.

`Cursor.schema` API tries to collect schema from all catalogs when `database` is `None`, but we should collect from only the current catalog.

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -6,11 +6,12 @@ from agate import Table
 from dbt.contracts.connection import AdapterResponse
 from dbt.adapters.base import AdapterConfig
 from dbt.adapters.base.relation import BaseRelation
-from dbt.adapters.databricks import DatabricksConnectionManager
-from dbt.adapters.databricks.relation import DatabricksRelation
-from dbt.adapters.databricks.column import DatabricksColumn
-
 from dbt.adapters.spark.impl import SparkAdapter
+
+from dbt.adapters.databricks.column import DatabricksColumn
+from dbt.adapters.databricks.connections import DatabricksConnectionManager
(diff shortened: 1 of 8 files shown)


ueshin:
This is based on #98.

dbt/adapters/databricks/impl.py
```diff
@@ -39,13 +39,15 @@ class DatabricksAdapter(SparkAdapter):
 
     def list_schemas(self, database: Optional[str]) -> List[str]:
         """Get a list of existing schemas in database."""
```
allisonwang-db, line 41: Let's also update this comment to specify the behavior when `database` is None?

dbt/adapters/databricks/impl.py
```diff
@@ -39,13 +39,15 @@ class DatabricksAdapter(SparkAdapter):
...
-        results = self.connections.list_schemas(database=database)
+        if database is not None:
+            results = self.connections.list_schemas(database=database)
+        else:
+            results = self.execute_macro(LIST_SCHEMAS_MACRO_NAME, kwargs={"database": database})
         return [row[0] for row in results]
 
     def check_schema_exists(self, database: Optional[str], schema: str) -> bool:
         """Check if a schema exists."""
```
allisonwang-db, line 49: ditto

allisonwang-db (approved):

dbt/adapters/databricks/impl.py
ueshin: Maybe we don't need here as it call `list_schemas` now?

allisonwang-db: Sounds good.

ueshin:
Thanks! merging.

PR #105, opened 2022-05-26
Support multi-catalog

### LINKED ISSUE #95: Support for Databricks CATALOG as a DATABASE in DBT compilations
### Describe the feature

Unity Catalog for Databricks now supports a three-level namespace via `CATALOG` in analogy to how most SQL dialects support `DATABASES`: https://docs.databricks.com/data-governance/unity-catalog/queries.html#three-level-namespace-notation

 

It would be excellent if DBT was able to treat `DATABASE` as `CATALOG` when compiling SQL for databricks.  This is done in analogy to custom databases in DBT.  In fact, the BigQuery connector seems to already support a similar configuration via `project` configuration: https://docs.getdbt.com/docs/building-a-dbt-project/building-models/using-custom-databases

 

We would have to not use `DATABASE` or custom databases at all with Databricks, or we would have to do so using a pre-build hook that issues a `USE CATALOG` statement.

 

### Additional context

PRs #89 and #94 are already working on this, I added this in support of this work!

 

### Who will this benefit?

Anyone who wants to use custom "databases" feature of DBT in databricks.

 

Would love to help any way I can.

ueshin:
resolves #95

### Description

Supports multi-catalog.

Enables `catalog` or `database` config to use a different catalog for models.

- model `alternative_catalog`

```sql
{{ config(
    catalog = 'alternative',
    materialized = 'table'
) }}

select * from {{ ref('seed') }}
```

Also enables to run cross catalog queries.

```sql
select * from {{ ref('seed') }}
union all select * from {{ ref('alternative_catalog')}}
```

Note: mixing Unity Catalog and Hive metastore tables is not recommened:

> org.apache.spark.sql.AnalysisException: Non-Unity-Catalog object `hive_metastore`.`schema_1`.`table_a` can't be referenced in Unity Catalog objects

> org.apache.spark.sql.AnalysisException: Create a persistent view that references both unity catalog and Hive metastore objects is not supported in Unity Catalog


ueshin:
This is based on #94.

dbt/adapters/databricks/impl.py
```diff
@@ -86,7 +94,9 @@ def list_relations_without_caching(
     ) -> List[DatabricksRelation]:
         kwargs = {"schema_relation": schema_relation}
         try:
-            results = self.execute_macro(LIST_RELATIONS_MACRO_NAME, kwargs=kwargs)
+            # The catalog for `show table extended` needs to be the current catalog.
```
allisonwang-db, line 97: We cannot pass the catalog name in the `SHOW TABLE EXTENDED` command? Something like `SHOW TABLE EXTENDED IN catalog.db LIKE '*'` where `catalog` is not the current catalog.

dbt/adapters/databricks/impl.py
```diff
@@ -185,3 +210,19 @@ def _get_columns_for_catalog(self, relation: DatabricksRelation) -> Iterable[Dic
             as_dict["column_name"] = as_dict.pop("column", None)
             as_dict["column_type"] = as_dict.pop("dtype")
             yield as_dict
+
+    @contextmanager
+    def _catalog(self, catalog: Optional[str]) -> Iterator[None]:
+        current_catalog: Optional[str] = None
+        try:
+            if catalog is not None:
+                current_catalog = self.execute_macro(CURRENT_CATALOG_MACRO_NAME)[0][0]
```
allisonwang-db, line 219: Just curious is the result of current_catalog cached somewhere?

dbt/adapters/databricks/impl.py
ueshin: Unfortunately, it's not supported. We need to set the current catalog the same as the catalog in the statement.

```
>>> with dbsql.connect(host, path, token, catalog="main") as conn:
...   with conn.cursor() as cursor:
...     cursor.execute("SHOW TABLE EXTENDED IN alternative.default LIKE '*'")
...     print(cursor.fetchall())
...
Traceback (most recent call last):
...
databricks.sql.exc.ServerOperationError: Error running query: org.apache.spark.sql.AnalysisException: Fail to execute the command as the target schema `alternative.default` is not in the current catalog. Please set the current catalog with 'USE CATALOG alternative' first.
```

dbt/adapters/databricks/impl.py
ueshin: No, it's not cached.

dbt/adapters/databricks/impl.py
allisonwang-db:
```suggestion
            # The catalog for `show table extended` needs to match the current catalog.
```

dbt/adapters/databricks/impl.py
```diff
@@ -185,3 +210,19 @@ def _get_columns_for_catalog(self, relation: DatabricksRelation) -> Iterable[Dic
             as_dict["column_name"] = as_dict.pop("column", None)
             as_dict["column_type"] = as_dict.pop("dtype")
             yield as_dict
+
+    @contextmanager
+    def _catalog(self, catalog: Optional[str]) -> Iterator[None]:
```
allisonwang-db, line 215: nit: maybe we can add a comment here?

dbt/include/databricks/macros/catalog.sql
```diff
@@ -0,0 +1,20 @@
...
+  {% do return(load_result('current_catalog').table) %}
+{% endmacro %}
+
+{% macro use_catalog(catalog) -%}
+  {{ return(adapter.dispatch('use_catalog', 'dbt')(catalog)) }}
+{% endmacro %}
+
+{% macro databricks__use_catalog(catalog) -%}
+  {% call statement() %}
+    use catalog `{{ catalog }}`
```
allisonwang-db, line 18: Should we use `adapter.quote` here?

tests/integration/multi_catalog/test_multi_catalog.py
```diff
@@ -0,0 +1,100 @@
...
+            kwargs=dict(database=self.alternative_database),
+        )
+        self.assertEqual(len(results), 2)
+
+        catalog = self.run_dbt(["docs", "generate"])
+        assert len(catalog.nodes) == 5
+
+
+class TestMultiCatalogTableModels(TestMultiCatalog):
+    @use_profile("databricks_uc_cluster")
```
allisonwang-db, line 58: How would the test behave if we run it against a non-uc cluster?

allisonwang-db (commented):
Looks good! Added a few comments.

tests/integration/multi_catalog/test_multi_catalog.py
ueshin: It will just be skipped.

allisonwang-db (approved):
Looks good!

ueshin:
Thanks! merging.

PR #106, opened 2022-06-02
Include the value of environment variable "DBT_INVOCATION_ENV" to the user agent entry.

ueshin:
### Description

Include the value of environment variable "DBT_INVOCATION_ENV" to the user agent entry.

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -1,5 +1,6 @@
 from contextlib import contextmanager
 from dataclasses import dataclass
+import os
 import re
 import time
 from typing import Any, Callable, ClassVar, Dict, Iterator, List, Optional, Sequence, Tuple
@@ -14,6 +15,7 @@ from dbt.contracts.connection import Connection, ConnectionState
 from dbt.events import AdapterLogger
 from dbt.events.functions import fire_event
 from dbt.events.types import ConnectionUsed, SQLQuery, SQLQueryStatus
+from dbt.tracking import DBT_INVOCATION_ENV
 from dbt.utils import DECIMALS
 
 from dbt.adapters.spark.connections import SparkConnectionManager, _is_retryable_error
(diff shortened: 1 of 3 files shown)


tests/unit/test_adapter.py
```diff
@@ -116,35 +117,55 @@ def test_database_and_catalog_settings(self):
...
+                self.assertEqual(
+                    _user_agent_entry,
+                    f"dbt-databricks/{__version__.version} ({dbt_invocation_env})",
+                )
+
+            return connect
 
-        self._test_databricks_sql_connector_connection("1.0.2", connect)
+        self._test_databricks_sql_connector_connection("1.0.2", connect_func("manual"))
+        with envvar(DBT_INVOCATION_ENV, "dbt_invocation_env"):
```
allisonwang-db, line 137: Can we use `mock.patch.dict` to patch `os.environ` here?

ueshin: Could you show me how to use it? Thanks.

dbt/adapters/databricks/connections.py
```diff
@@ -256,7 +258,8 @@ def open(cls, connection: Connection) -> Connection:
                 cls.validate_creds(creds, required_fields)
 
                 dbt_databricks_version = __version__.version
-                user_agent_entry = f"dbt-databricks/{dbt_databricks_version}"
+                dbt_invocation_env = os.getenv(DBT_INVOCATION_ENV) or "manual"
+                user_agent_entry = f"dbt-databricks/{dbt_databricks_version} ({dbt_invocation_env})"
```
allisonwang-db, line 262: Looks like in dbt-spark we don't have the parentheses around `dbt_invocation_env`.

allisonwang-db (approved):

dbt/adapters/databricks/connections.py
ueshin: Actually dbt-spark doesn't have this parameter (yet?).

dbt/adapters/databricks/connections.py
```diff
@@ -256,7 +258,8 @@ def open(cls, connection: Connection) -> Connection:
                 cls.validate_creds(creds, required_fields)
 
                 dbt_databricks_version = __version__.version
-                user_agent_entry = f"dbt-databricks/{dbt_databricks_version}"
+                dbt_invocation_env = os.getenv(DBT_INVOCATION_ENV) or "manual"
```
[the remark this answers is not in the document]
ueshin: @jtcohen6 Thanks for letting us know!
I agree to handle it in the analytical system, so I'd just leave it as is for now.

PR #107, opened 2022-06-03
Fix the user agent entry to avoid nested parens.

ueshin:
### Description

Fixes the user agent entry to avoid nested parens.

The full user agent was like `PyDatabricksSqlConnector/2.0.2 (dbt-databricks/1.2.0 (dbt-cloud))`, but we should avoid nested parens.

The full user agent after the fix will be like `PyDatabricksSqlConnector/2.0.2 (dbt-databricks/1.2.0; dbt-cloud)`.


allisonwang-db (approved):
🚢

ueshin:
Thanks! merging.

PR #116, opened 2022-06-22
Add setting of UC to the project template

ueshin:
### Description

Adds setting of UC to the project template:

- When using UC:

```
$ dbt init
...
[1] use Unity Catalog
[2] not use Unity Catalog
Desired unity catalog option (enter a number): 1
catalog (initial catalog): cat
...
```

- When not using UC:

```
$ dbt init
...
[1] use Unity Catalog
[2] not use Unity Catalog
Desired unity catalog option (enter a number): 2
...
```

### DIFF
--- a/dbt/include/databricks/profile_template.yml
+++ b/dbt/include/databricks/profile_template.yml
@@ -8,6 +8,12 @@ prompts:
   token:
     hint: 'dapiXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
     hide_input: true
+  _choose_unity_catalog:
+    'use Unity Catalog':
+      catalog:
+        hint: 'initial catalog'
+    'not use Unity Catalog':
+      _fixed_catalog: null
   schema:
     hint: 'default schema that dbt will build objects in'
   threads:


dbt/include/databricks/profile_template.yml
```diff
@@ -8,6 +8,12 @@ prompts:
   token:
     hint: 'dapiXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
     hide_input: true
+  _choose_unity_catalog:
+    'use Unity Catalog':
+      catalog:
```
allisonwang-db, line 13: What if the user leaves it empty? Will it use current_catalog()?

allisonwang-db (approved):

dbt/include/databricks/profile_template.yml
ueshin: The prompt will ask again:

```
Desired unity catalog option (enter a number): 1
catalog (initial catalog):
catalog (initial catalog):
catalog (initial catalog):
catalog (initial catalog):
catalog (initial catalog):
```

If the user explicitly set it empty by editing the profile file, dbt command will fail:

```
  Credentials in profile "xxx", target "dev" invalid: Runtime Error
    Invalid catalog name : .
```

ueshin:
Thanks! merging.

PR #134, opened 2022-07-18
Add new basic tests TestDocsGenerateDatabricks and TestDocsGenReferencesDatabricks

ueshin:
### Description

Adds new basic tests `TestDocsGenerateDatabricks` and `TestDocsGenReferencesDatabricks` based on `BaseDocsGenerate` and `BaseDocsGenReferences` respectively as per an item from #127.


tests/functional/adapter/test_basic.py
```diff
@@ -50,3 +58,58 @@ class TestSnapshotTimestampDatabricks(BaseSnapshotTimestamp):
...
+
+
+class TestDocsGenReferencesDatabricks(BaseDocsGenReferences):
+    @pytest.fixture(scope="class")
+    def model_stats(self, is_uc):
+        if is_uc:
+            return no_stats()
+        else:
+            return {
+                "bytes": {
```
allisonwang-db, line 85: I am wondering why do we need "bytes" and "has_stats" here vs no stats for UC?

ueshin: This is an expected value of the stats for `ephemeral_summary` in the test.
Seems like currently non-UC clusters return the stats somehow.

allisonwang-db (approved):

ueshin:
Thanks! merging.

PR #136, opened 2022-07-22
Avoid using default as a temporary fix for dbt-labs/dbt-core#5359.

ueshin:
### Description

Avoids using default as a temporary fix for dbt-labs/dbt-core#5359.

This is a temporary fix and dbt-labs/dbt-spark#394 should be ported later.


ueshin:
Let me merge this now to unblock other issues.

PR #137, opened 2022-07-22
Apply connection retry refactor, add defaults with exponential backoff

ueshin:
### Description

Applies connection retry refactor, add defaults with exponential backoff as per an item from #127.

- Refactors retry logic to use `retry_connection` from core.
- Adds consistent defaults with other adpaters:
  - if `connect_retries` and `connect_timeout` are not specified: 1 retry after 1s
  - if `connect_retries` is specified but `connect_timeout` is not, it will use exponential backoff


xg1990:
Thanks a lot! Our team is eagerly waiting for this feature!

ueshin:
Thanks! merging.

PR #142, opened 2022-07-27
Fix DatabricksCredentials._connection_keys

ueshin:
### Description

Fixes `DatabricksCredentials._connection_keys`.

- `catalog` can't be used; should use `database`
- `database` can't be removed.

Otherwise, `Undefined` object comes from somewhere:

```
TypeError: can not serialize 'Undefined' object
```

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -98,9 +98,7 @@ class DatabricksCredentials(Credentials):
         return self.host
 
     def _connection_keys(self) -> Tuple[str, ...]:
-        connection_keys = ["host", "http_path", "schema"]
-        if self.database:
-            connection_keys.insert(2, "catalog")
+        connection_keys = ["host", "http_path", "database", "schema"]
         if self.session_properties:
             connection_keys.append("session_properties")
         return tuple(connection_keys)


dbt/adapters/databricks/connections.py
```diff
@@ -98,9 +98,7 @@ def unique_field(self) -> str:
         return self.host
 
     def _connection_keys(self) -> Tuple[str, ...]:
-        connection_keys = ["host", "http_path", "schema"]
-        if self.database:
```
allisonwang-db, line 102: The connection_keys seems to be used here: https://github.com/dbt-labs/dbt-core/blob/8f50eee330ea34dd6aaba40993a01a380373fa32/core/dbt/contracts/connection.py#L144-L153

@jtcohen6 do you know if we can use `catalog` instead of `database` as a connection key here? 
```
Connection:
  host: your.databrickshost.com
  http_path: /sql/your/http/path
  database: catalog_name  <----- change this to "catalog: catalog_name"
  schema: schema_name
```

allisonwang-db (approved):

ueshin:
Let me merge this for now to unblock testing for 1.2.0 release. Thanks.

dbt/adapters/databricks/connections.py
ueshin: I submitted another PR with a hacky way (#145).

jtcohen6: Ah - I don't think there's a good way to do this, based on the way we've implemented aliases (always resolves to the "true" name within dbt). I think the hacky way is good enough for now

PR #145, opened 2022-07-27
Retry DatabricksCredentials._connection_keys to show only meaningful configs

ueshin:
### Description

Retry: `DatabricksCredentials._connection_keys` to show only meaningful configs (#138).

Assuming `DatabricksCredentials.connection_info(self, *, with_aliases: bool = False)` is called from only:

- `Profile` with `with_aliases=True`
- `DebugTask` without `with_aliases` (`False` by default)

Thus, if `with_aliases` is `True`, `DatabricksCredentials._connection_keys` should return the same tuple as the current one for the internal use; otherwise it can use `catalog` to show in the debug output.

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -120,6 +120,14 @@ class DatabricksCredentials(Credentials):
                 yield key, as_dict[key]
 
     def _connection_keys(self, *, with_aliases: bool = False) -> Tuple[str, ...]:
+        # Assuming `DatabricksCredentials.connection_info(self, *, with_aliases: bool = False)`
+        # is called from only:
+        #
+        # - `Profile` with `with_aliases=True`
+        # - `DebugTask` without `with_aliases` (`False` by default)
+        #
+        # Thus, if `with_aliases` is `True`, `DatabricksCredentials._connection_keys` should return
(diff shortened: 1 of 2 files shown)


ueshin:
@jtcohen6 Is there an official way to achieve something like this?

We want to show `catalog` instead of `database` as a connection config key when `dbt debug` because `database` could confuse Unity **Catalog** users.

dbt/adapters/databricks/connections.py
```diff
@@ -97,8 +109,22 @@ def type(self) -> str:
...
+        aliases: List[str] = []
+        if with_aliases:
+            aliases = [k for k, v in self._ALIASES.items() if v in connection_keys]
+        for key in itertools.chain(self._connection_keys(with_aliases=with_aliases), aliases):
+            if key in as_dict:
+                yield key, as_dict[key]
+
+    def _connection_keys(self, *, with_aliases: bool = False) -> Tuple[str, ...]:
+        connection_keys = ["host", "http_path", "schema"]
+        if with_aliases:
```
allisonwang-db, line 124: Can we add a short comment here? Similar to the PR description.

allisonwang-db (approved):

ueshin:
Thanks! merging.

PR #149, opened 2022-08-05
Use insert overwrite for the first batch of seed

### LINKED ISSUE #114: Data is duplicated on reloading seeds that are using an external table
### Describe the bug

When seeds are configured as an external table, the data is getting duplicated on reload. 

### Steps To Reproduce

I create a repo with a demo of the issue: https://github.com/dejan/dbt-demo-inconsistent-seeds but here is a short instruction on how to reproduce:

Have seeds/cities.yml:
```csv
id,name
1,berlin
2,paris
```

Have seeds/countries.yml:
```csv
id,name
1,germany
2,france
```

Configure one seed to use managed table (by default) and another one to use external (by setting location_root).

```yml
seeds:
  foo:
    cities:
    countries:
      location_root: "{{ env_var('LOCATION') }}"
```

Run dbt seed *twice*:

    dbt seed && dbt seed

Observe the content in both tables.

```sql
select * from foo.cities
```

|id|name|
|--|----|
|1|berlin|
|2|paris|

```sql
select * from foo.countries
```

|id|name|
|--|----|
|1|germany|
|1|germany|
|2|france|
|2|france|

### Expected behavior
The behavior should be consistent regardless of the table type. The data should be reloaded - ie there should be no duplicates.

### System information
```
Core:
  - installed: 1.1.0
  - latest:    1.1.1 - Update available!

Plugins:
  - spark:      1.1.0 - Up to date!
  - databricks: 1.1.0 - Up to date!
```

**The operating system you're using: OS X Big Sur**

**The output of `python --version`: Python 3.8.10**

ueshin:
resolves #114

### Description

Uses `insert overwrite` for the first batch of seed to help seeds in external tables.

This must be a temporary fix and should follow dbt-labs/dbt-spark#182.

### DIFF
--- a/dbt/include/databricks/macros/materializations/seed.sql
+++ b/dbt/include/databricks/macros/materializations/seed.sql
@@ -2,6 +2,44 @@
   {{ return('%s') }}
 {% endmacro %}
 
+{% macro databricks__load_csv_rows(model, agate_table) %}
+
+  {% set batch_size = get_batch_size() %}
+  {% set column_override = model['config'].get('column_types', {}) %}
+
+  {% set statements = [] %}
+
+  {% for chunk in agate_table.rows | batch(batch_size) %}
+      {% set bindings = [] %}
+
+      {% for row in chunk %}
+          {% do bindings.extend(row) %}
+      {% endfor %}
+
+      {% set sql %}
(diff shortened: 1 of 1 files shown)


zhengruifeng (approved):

allisonwang-db (commented):
Looks good!

ueshin:
Thanks! merging.

PR #152, opened 2022-08-11
Set upper bound for Python

ueshin:
### Description

The dependent library `databricks-sql-connector` doesn't support Python 3.10 yet (databricks/databricks-sql-python#26).
We should also avoid incompatible Python version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,6 @@ setup(
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",
     ],
-    python_requires=">=3.7",
+    python_requires=">=3.7,<3.10",
 )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -68,7 +68,6 @@ def _get_plugin_version():
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
-        "Programming Language :: Python :: 3.10",
     ],
-    python_requires=">=3.7",
+    python_requires=">=3.7,<3.10",
```
ueshin, line 72: This might be good enough without explicitly setting the upper bound. We could just say `dbt-databricks` does not **officially** support Python 3.10 yet.
@allisonwang-db WDYT?

allisonwang-db: Yea this should be fine.

allisonwang-db (approved):
Good catch! Thanks for the fix.

ueshin:
Thanks! merging.

PR #154, opened 2022-08-12
Set upper bound for databricks-sql-connector instead of Python

ueshin:
### Description

Set upper bound for `databricks-sql-connector` when Python 3.10 instead of `Python`.

Note that `databricks-sql-connector` doesn't support Python 3.10 yet, so `dbt-databricks` doesn't either.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
-databricks-sql-connector>=2.0.2
+databricks-sql-connector>=2.0.2; python_version<'3.10'
+databricks-sql-connector>=2.0.2,<2.0.3; python_version=='3.10'
 dbt-spark~=1.3.0a1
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1,2 +1,3 @@
-databricks-sql-connector>=2.0.2
+databricks-sql-connector>=2.0.2; python_version<'3.10'
+databricks-sql-connector>=2.0.2,<2.0.3; python_version=='3.10'
```
allisonwang-db, line 2: Why do we need this line?

allisonwang-db (approved):

requirements.txt
ueshin: It applies the upper bound only when Python 3.10.

ueshin:
Thanks! merging.

PR #163, opened 2022-08-26
Explicitly close cursors

ueshin:
### Description

Explicitly closes cursors.

Sometimes we observe **Query has been timed out due to inactivity** error in the Query History UI, that is likely because the cursors are not closed.

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -1,3 +1,4 @@
+import warnings
 from contextlib import contextmanager
 from dataclasses import dataclass
 import itertools
@@ -22,7 +23,8 @@ from agate import Table
 import dbt.exceptions
 from dbt.adapters.base import Credentials
 from dbt.adapters.databricks import __version__
-from dbt.contracts.connection import Connection, ConnectionState
+from dbt.clients import agate_helper
+from dbt.contracts.connection import AdapterResponse, Connection, ConnectionState
 from dbt.events import AdapterLogger
 from dbt.events.functions import fire_event
 from dbt.events.types import ConnectionUsed, SQLQuery, SQLQueryStatus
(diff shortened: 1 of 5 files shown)


zhengruifeng (approved):

dbt/include/databricks/macros/materializations/seed.sql
```diff
@@ -29,7 +29,7 @@
           {%- endfor %}
       {% endset %}
 
-      {% do adapter.add_query(sql, bindings=bindings, abridge_sql_log=True) %}
+      {% do adapter.add_query(sql, bindings=bindings, abridge_sql_log=True, close_cursor=True) %}
```
allisonwang-db, line 32: Why do we need to pass the `close_cursor` flag for seed?

dbt/adapters/databricks/connections.py
```diff
@@ -210,21 +219,23 @@ def description(
...
+            self._cursor.close()
+            with warnings.catch_warnings():
+                warnings.simplefilter("always")
+                warnings.warn("The cursor was closed by destructor.")
+
 
 class DatabricksConnectionManager(SparkConnectionManager):
     TYPE: ClassVar[str] = "databricks"
 
-    DROP_JAVA_STACKTRACE_REGEX: ClassVar["re.Pattern[str]"] = re.compile(
```
allisonwang-db, line 224: This is deleted because we never used it?

dbt/adapters/databricks/connections.py
```diff
@@ -141,46 +143,53 @@ def _connection_keys(self, *, with_aliases: bool = False) -> Tuple[str, ...]:
...
-
-    def cancel(self) -> None:
-        if self._cursor:
-            try:
-                self._cursor.cancel()
-            except DBSQLError as exc:
-                logger.debug("Exception while cancelling query: {}".format(exc))
-                _log_dbsql_errors(exc)
+    def cursor(self) -> "DatabricksSQLCursorWrapper":
+        return DatabricksSQLCursorWrapper(self._conn.cursor())
```
allisonwang-db, line 155: Should we avoid creating an instance of the wrapper each time we invoke the cursor?

dbt/adapters/databricks/connections.py
ueshin: Yes.

dbt/include/databricks/macros/materializations/seed.sql
ueshin: `adapter.add_query` doesn't close the cursor by default, then the cursor will remain opened.

dbt/adapters/databricks/connections.py
ueshin: How can we manage cursors otherwise?

allisonwang-db: Nvm this method should return a new cursor.

dbt/include/databricks/macros/materializations/seed.sql
allisonwang-db: I see!

allisonwang-db (approved):
Looks good!

ueshin:
Thanks! merging.

PR #169, opened 2022-09-12
Bump version to 1.3.0b0.dev0 and pin upstreams to 1.3.0b2.

ueshin:
### Description

Bumps version to `1.3.0b0.dev0` and pins upstreams to `1.3.0b2`.


ueshin:
Let me merge this now to unblock other PRs.

PR #174, opened 2022-09-16
http_headers should be dict in the profile.

ueshin:
### Description

Currently there is no way to specify `http_headers` in `connection_parameters` properly.

`http_headers` should be `dict` in the profile:

```yml
      ...
      connection_parameters:
        http_headers:
          aaa: xxx
          bbb: yyy
      ...
```

but previously it failed to connect to the cluster.

```
failed to connect: unsupported operand type(s) for +: 'dict' and 'list'
```


zhengruifeng (approved):

ueshin:
Thanks! merging.

PR #178, opened 2022-09-16
Follow enhancement/refactor python submission in dbt-spark

### LINKED ISSUE #182: Remove submit_python_job from Jinja context
### Describe the feature
With the latest change in dbt-core, we should consider remove [this line](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/impl.py#L263) so that user won't have access to `adapter.submit_python_job` in Jinja macros directly. This function is used by [a context member directly in `dbt-core`](https://github.com/dbt-labs/dbt-core/pull/5822/files#diff-d14b7d00e52da5da96bfb3bdef414e9289d923efe0a9cdadd716d792c7199535R1263).

Also something related to this area, we recently refactored the function here a little bit(dbt-core, dbt-spark). Maybe we want to follow the style there so any updates are automatically included

### Additional context
This would help dbt keep being opinionated about when certain operation should happen

ueshin:
Resolves #182

### Description

Follows enhancement/refactor python submission in dbt-labs/dbt-spark#452 and dbt-labs/dbt-core#5822.


zhengruifeng (approved):

dbt/adapters/databricks/impl.py
```diff
@@ -260,86 +250,9 @@ def run_sql_for_tests(
...
-            finally:
-                # Delete the execution context
-                api_client.Context.destroy(cluster_id=cluster_id, context_id=context_id)
-        except HTTPError as e:
-            raise dbt.exceptions.RuntimeException(str(e)) from e
-        finally:
-            api_client.close()
+    @property
+    def python_submission_helpers(self) -> Dict[str, Type[PythonJobHelper]]:
+        return {"commands": CommandApiPythonJobHelper}
```
ChenyuLInx, line 255: Hey @ueshin  sorry for updating things so frequent, but after talking to @lostmygithubaccount we decided to update the submission method to reflect the cluster type that we run python model on instead of the API we use.
Here's the [change on the dbt-spark](https://github.com/dbt-labs/dbt-spark/pull/467/files#diff-af2c2165403b748d8057a7770f80e17d100d0cf8999c4160d57b634ee15d4977R385) side.

I think for here all we need to do is change `commands` to `all_purpose_cluster ` 
```suggestion
        return {"all_purpose_cluster": CommandApiPythonJobHelper}
```

ChenyuLInx (changes requested):

dbt/adapters/databricks/impl.py
ueshin: oh, and maybe `check_credentials` was changed.

Thanks for letting me know. I'll handle it.

ueshin: Made minimum changes. I'll submit a follow-up PR.

ueshin:
Thanks! merging.

PR #180, opened 2022-09-17
Convert df to pyspark DataFrame if it is pandas before writing

### LINKED ISSUE #179: Enable pandas and pandas-on-Spark DataFrames for Databricks
### Describe the feature
When a dbt python model returns a Pandas DataFrame, materialize it as a table in Databricks.

Currently, an error will be raised for Pandas DataFrames (whereas PySpark DataFrames work fine).

Same as https://github.com/dbt-labs/dbt-bigquery/issues/312 and https://github.com/dbt-labs/dbt-spark/issues/468

I will submit a pull request for this 👍

dbeatty10:
resolves #179

### Description

@chamini2 is the actual author of this solution ([here](https://github.com/dbt-labs/dbt-bigquery/pull/301)). I just copied it to [dbt-spark](https://github.com/dbt-labs/dbt-spark/pull/469) and here and confirmed that it works.

### Two options
This pull request is one option for implementation this feature.

The other option is to just inherit the implementation from dbt-spark. e.g., merely delete the `py_write_table` implementation within dbt-databricks entirely once https://github.com/dbt-labs/dbt-spark/pull/469 is merged.

### Testing
I manually confirmed that both options will work. The testing was manual since it won't be covered by automated testing in CI.

### Checklist


dbt/include/databricks/macros/materializations/table.sql
```diff
@@ -42,5 +42,18 @@
...
+import importlib.util
+import pyspark.pandas
+package_name = 'pandas'
+if importlib.util.find_spec(package_name):
+    import pandas
+    # convert to pyspark.sql.dataframe.DataFrame
+    if isinstance(df, pandas.core.frame.DataFrame):
+      df = spark.createDataFrame(df)
+    elif isinstance(df, pyspark.pandas.frame.DataFrame):
+      df = df.to_spark()
```
ueshin, line 56: The following flow would be better?

- if df is pandas DataFrame,
  - if pyspark.pandas is available, convert to pandas-on-Spark DataFrame
  - otherwise convert to Spark DataFrame
- if df is pandas-on-Spark DataFrame, convert to Spark DataFrame

ueshin (commented):
Thanks for working on this!
I left some comments, please take a look.

dbt/include/databricks/macros/materializations/table.sql
```diff
@@ -42,5 +42,18 @@
 # --- Autogenerated dbt materialization code. --- #
 dbt = dbtObj(spark.table)
 df = model(dbt, spark)
+
+# make sure pandas exists
+import importlib.util
+import pyspark.pandas
```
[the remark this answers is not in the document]
dbeatty10: Yep, I will add that check before doing the import.

dbt/include/databricks/macros/materializations/table.sql
dbeatty10: Would that introduce **two** different conversions (rather than one) in the case that the input is a pandas-on-Spark DataFrame? Wouldn't we want only a single conversion if possible?

ueshin: I think that's fine as converting from pandas-on-Spark to Spark DataFrame has no overhead.

```py
if isinstance(df, pd.DataFrame):
    if ps_available:
        df = ps.DataFrame(df)
    else:
        df = spark.createDataFrame(df)

if ps_available and isinstance(df, ps.DataFrame):
    df = df.to_spark()
```

dbeatty10: Thank you for this code example @ueshin !

My impression is that there's only 3 possibilities we care about for the original type of the df:
1. Spark DataFrame (no conversion needed)
1. pandas DataFrame
1. pandas-on-Spark DataFrame

Only the latter two need to be converted to Spark DataFrame.
Would the following flow work?
- if df is a Spark DataFrame, do nothing
- else if `pandas` is available and df is a pandas DataFrame, convert to Spark DataFrame
- else if `pyspark.pandas` is available and df is a pandas-on-Spark DataFrame, convert to Spark DataFrame
- else raise a helpful error message

My understanding is that we only need to import `pandas` and `pyspark.pandas` so that we can use `isinstance()`. Without those imports, then we can't determine if the df has those types using `isinstance()`.

An alternative is something like `type(df) == "pandas.core.frame.DataFrame"`, which has drawbacks like not supporting inheritance.

ueshin: The reason I want it to try to convert pandas DataFrame to pandas-on-Spark DataFrame first is that `ps.DataFrame(pandas_df)` knows how to convert pandas DataFrame better than `spark.createDataFrame(pandas_df)`.

Let's keep using `isinstance()`; otherwise we would miss something that inherits pandas/pandas-on-Spark DataFrame.

dbeatty10: I vastly underestimated the nuances of this PR! Thank you for your patient explanations 🙏 

Incorporating all your feedback here so that we can get all the details right as far upstream as possible:
https://github.com/dbt-labs/dbt-spark/pull/469

In hindsight, I should have created just a single PR there and invited you to be reviewer on it.

ueshin: Now that dbt-labs/dbt-spark#469 looks good to me, we can remove the macro in our side once it gets merged.

dbeatty10:
Closing in favor of https://github.com/dbt-labs/dbt-spark/pull/469. Once that is merged, a new PR here in dbt-databricks can remove the `py_write_table` macro override.

PR #185, opened 2022-09-21
Use builtin timestampadd and timestampdiff functions for dateadd/datediff macros if available.

ueshin:
### Description

Uses builtin `timestampadd` and `timestampdiff` functions for `dateadd`/`datediff` macros if available.

The functions [timestampadd](https://docs.databricks.com/spark/latest/spark-sql/language-manual/functions/timestampadd.html) and [timestampdiff](https://docs.databricks.com/spark/latest/spark-sql/language-manual/functions/timestampdiff.html) have been available since DBR 10.4. Those are equivalent to what `dateadd`/`datediff` macros are trying to achieve.


zhengruifeng (approved):

dbt/adapters/databricks/impl.py
```diff
@@ -59,6 +60,19 @@ class DatabricksAdapter(SparkAdapter):
 
     AdapterSpecificConfigs = DatabricksConfig
 
+    @available.parse(lambda *a, **k: 0)
```
allisonwang-db, line 63: Just curious, why do we need this `@available.parse`?

allisonwang-db (approved):

dbt/adapters/databricks/impl.py
ueshin: Without that, the function will be executed even at parse time that causes `InvalidConnectionException` because connections are not available at parse time.

> A decorator factory to indicate that a method on the adapter will be
        exposed to the database wrapper, and will be stubbed out at parse time
        with the given function.

ueshin:
Thanks! merging.

PR #186, opened 2022-09-21
Manage cursors to avoid relying on the protected member.

ueshin:
### Description

Manages cursors to avoid relying on the protected member.


zhengruifeng (approved):

ueshin:
Thanks! merging.

PR #191, opened 2022-09-26
Remove Koalas related tests.

ueshin:
### Description

Removes Koalas related tests that was added at https://github.com/dbt-labs/dbt-core/commit/e58edaab2deba8bb14396f463489adf0e1e02e6e.

We won't encourage users to use Koalas any more.


ueshin:
Let me merge this now to unblock other PRs. Thanks.

PR #194, opened 2022-09-26
Follow "More flexible cluster configuration".

ueshin:
### Description

Follows "More flexible cluster configuration" at dbt-labs/dbt-spark#467.

- Reuse `dbt-spark`'s implementation
- Remove the dependency on `databricks-cli`
- Internal refactorings


dbt/adapters/databricks/python_submissions.py
```diff
@@ -1,75 +1,65 @@
...
-            raise dbt.exceptions.RuntimeException(str(e)) from e
+class DbtDatabricksAllPurposeClusterPythonJobHelper(
+    DbtDatabricksBasePythonJobHelper, AllPurposeClusterPythonJobHelper
+):
+    def check_credentials(self) -> None:
+        self.credentials.validate_creds()
+        if not self.cluster_id:
+            raise ValueError(
+                "Databricks `http_path` or `cluster_id` of all-purpose cluster is required "
+                "for `all_purpose_cluster` submission method."
```
allisonwang-db, line 64:
```suggestion
                "Databricks `http_path` or `cluster_id` of an all-purpose cluster is required "
                "for the `all_purpose_cluster` submission method."
```

dbt/adapters/databricks/python_submissions.py
```diff
@@ -1,75 +1,65 @@
...
+        self.auth_header.update({"User-Agent": user_agent, **http_headers})
 
-    def submit(self, compiled_code: str) -> None:
-        cluster_id = self.cluster_id
+    @property
+    def cluster_id(self) -> Optional[str]:  # type: ignore[override]
+        return self.parsed_model.get(
+            "cluster_id",
+            self.credentials.extract_cluster_id(
+                self.parsed_model.get("http_path", self.credentials.http_path)
```
allisonwang-db, line 40: Can we specify an `http_path` that is different from the one in dbt profiles for a Python model?

allisonwang-db (approved):
This looks good! Thanks for the refactoring.

dbt/adapters/databricks/python_submissions.py
ueshin: Thanks for the heads-up.
Actually I double-checked it and found there is a mistake. I'll update it.

After the update, yes, we can specify `http_path` in the Python model config that is different from the one in the profile.

ueshin:
Thanks! merging.

PR #196, opened 2022-09-27
Fix the macro for Python models.

ueshin:
### Description

Fixes the macro for Python models.

In `dbt-spark`, `importlib.util.find_spec` is used to check whether the module is available or not, but it raises `ModuleNotFoundError` if it can't find the module.

We should use `try-except` instead.


ueshin:
This PR is intended to be a temporary fix. The macro in `dbt-spark` should be fixed. cc @dbeatty10

dbeatty10 (approved):
👍 Looks good and makes sense @ueshin.

Thank you for raising this originally [here](https://github.com/dbt-labs/dbt-spark/pull/474#discussion_r981644066).

If they don't exist there already, I'll open issues within dbt-spark and dbt-bigquery to report this bug. There's a very good chance we'll use your solution here as-is.

ueshin:
Let me merge this for now to unblock other PRs.
I'll revert this commit once `dbt-spark` side is fixed.
Thanks.

PR #198, opened 2022-09-29
Use search instead of match to extract DBR version.

ueshin:
### Description

Uses `search` instead of `match` to extract DBR version because `match` is sometimes too strict.


allisonwang-db (approved):

ueshin:
Thanks! merging.

PR #206, opened 2022-10-07
Add catalog to the work dir path.

ueshin:
### Description

Adds `catalog` to the work dir path for the case where there are the same name schema in different catalogs.

The work dir will be `"/Shared/dbt_python_model/{schema}/"` without `catalog`, and `"/Shared/dbt_python_model/{catalog}/{schema}/"` with `catalog`.

### DIFF
--- a/dbt/adapters/databricks/python_submissions.py
+++ b/dbt/adapters/databricks/python_submissions.py
@@ -18,6 +18,8 @@ class DbtDatabricksBasePythonJobHelper(BaseDatabricksHelper):
             parsed_model=parsed_model, credentials=credentials  # type: ignore[arg-type]
         )
 
+        self.database = parsed_model.get("database")
+
         user_agent = f"dbt-databricks/{version}"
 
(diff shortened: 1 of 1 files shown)


dbt/adapters/databricks/python_submissions.py
```diff
@@ -18,6 +18,8 @@ def __init__(self, parsed_model: Dict, credentials: DatabricksCredentials) -> No
             parsed_model=parsed_model, credentials=credentials  # type: ignore[arg-type]
         )
 
+        self.database = parsed_model.get("database")
```
allisonwang-db, line 21: Just to confirm: if a user specifies `catalog`, it will still be stored in this database entry right?

allisonwang-db (approved):

dbt/adapters/databricks/python_submissions.py
ueshin: Yes, it will be stored here.

ueshin:
Thanks! merging.

PR #213, opened 2022-10-12
Avoid comparing the query itself.

ueshin:
### Description

Avoids comparing the query itself because the table names are different between non-UC and UC tests.

### DIFF
--- a/tests/functional/adapter/test_ephemeral.py
+++ b/tests/functional/adapter/test_ephemeral.py
@@ -1,6 +1,3 @@
-import os
-import re
-
 from dbt.tests.adapter.ephemeral.test_ephemeral import BaseEphemeralMulti
 from dbt.tests.util import run_dbt, check_relations_equal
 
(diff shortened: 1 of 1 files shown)


zhengruifeng (approved):

ueshin:
Thanks! merging.

tests/functional/adapter/test_ephemeral.py
```diff
@@ -14,25 +11,3 @@ def test_ephemeral_multi(self, project):
...
-        assert os.path.exists("./target/run/test/models/double_dependent.sql")
-        with open("./target/run/test/models/double_dependent.sql", "r") as fp:
-            sql_file = fp.read()
-
-        sql_file = re.sub(r"\d+", "", sql_file)
-        expected_sql = (
-            "create or replace view test_test_ephemeral.double_dependent"
-            "as"
-            "with __dbt__cte__base as ("
-            "select * from test_test_ephemeral.seed"
```
allisonwang-db, line 27: Do you mean the table name here is different between UC and non-UC tests?

ueshin: 2-layer vs. 3-layer, to be exact.

allisonwang-db: 👍  3L namespace for UC tables.

PR #220, opened 2022-10-25
Fix double quoting strings in HTTP headers

KoningJasper:
### Description

Before a header like `{'string': 'some-string'}` would result in the header: `{'string': '"some-string"'}`. This PR fixes this and returns the correct header `{'string': 'some-string'}`.

### Checklist


ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

PR #222, opened 2022-10-31
Raise an exception when schema contains '.'.

ueshin:
### Description

Raises an exception when schema contains `'.'`.

This is a breaking change since `1.4.0` to disallow a workaround using `schema` to set a catalog for the older versions.


zhengruifeng (approved):

allisonwang-db (approved):
Nice!

ueshin:
Thanks! merging.

PR #223, opened 2022-11-02
Fix copy into macro when passing expression_list.

ueshin:
### Description

Fixes copy into macro when passing `expression_list`.

```
Encountered an error while running operation: Runtime Error
  Runtime Error

    [PARSE_SYNTAX_ERROR] Syntax error at or near 'select'(line 5, pos 9)

    == SQL ==
    /* ... */


        copy into test16673467754364144015_copy_into.target_with_expression_list
        from select id, name from '/path/to/source'
    ---------^^^
        fileformat = parquet


        format_options ('mergeSchema' = 'true')
        copy_options ('mergeSchema' = 'true')
```

There should be parenthesis surrounding `select id, name from '/path/to/source'`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
     - Containing a catalog in `schema` is not allowed anymore.
     - Need to explicitly use `catalog` instead.
 
+## dbt-databricks 1.3.2 (Release TBD)
+
+### Fixes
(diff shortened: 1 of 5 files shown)


allisonwang-db (approved):
Thanks for fixing this!

ueshin:
Thanks! merging.

PR #221, opened 2022-10-27
Show and log a warning when schema contains '.'.

ueshin:
### Description

Shows and logs a warning when schema contains `'.'`.

As a side work, trying to help more cases where schema contains `'.'` in `list_relations_without_caching`.

PR #224, opened 2022-11-03
Partially revert to fix the case where schema config contains uppercase letters.

ueshin:
### Description

Partially reverts #221 to fix the case where `schema` config contains uppercase letters.

```
Compilation Error in model test (models/test.sql)
  When searching for a relation, dbt found an approximate match. Instead of guessing
  which relation to use, dbt will move on. Please delete DBT_TEST.test, or rename it to be less ambiguous.
  Searched for: dbt_test.test
  Found: DBT_TEST.test

  > in macro load_cached_relation (macros/adapters/relation.sql)
  > called by macro load_relation (macros/adapters/relation.sql)
  > called by macro materialization_incremental_databricks (macros/materializations/incremental/incremental.sql)
  > called by model test (models/test.sql)
```

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -131,13 +131,13 @@ class DatabricksAdapter(SparkAdapter):
                     f'Invalid value from "show table extended ...", '
                     f"got {len(row)} values, expected 4"
                 )
-            _, name, _, information = row
+            _schema, name, _, information = row
             rel_type = RelationType.View if "Type: VIEW" in information else RelationType.Table
             is_delta = "Provider: delta" in information
             is_hudi = "Provider: hudi" in information
             relation = self.Relation.create(
                 database=schema_relation.database,
-                schema=schema_relation.schema,
+                schema=_schema,
                 identifier=name,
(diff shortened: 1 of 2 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -131,13 +131,13 @@ def list_relations_without_caching(  # type: ignore[override]
                     f'Invalid value from "show table extended ...", '
                     f"got {len(row)} values, expected 4"
                 )
-            _, name, _, information = row
+            _schema, name, _, information = row
```
allisonwang-db, line 134: Can we comment here on why we need the schema returned from the row?

allisonwang-db (approved):

dbt/adapters/databricks/impl.py
ueshin: Sure, left a comment.
Actually we might have a better way to handle this, but let me just revert to avoid the current regression.

allisonwang-db: Sounds good!

ueshin:
Thanks! merging.

PR #231, opened 2022-11-29
Avoid show table extended command.

ueshin:
### Description

Avoids show table extended command.

This is based on dbt-labs/dbt-spark#433.

1. Create a table/view list with `show tables in {{ relation }}` and `show views in {{ relation }}` commands, or `get_tables` API when `catalog` is provided.
2. Retrieve additional information by `describe extended {{ relation }}` command.
    1. `get_relation` with `needs_information=True`
    2. `get_columns_in_relation`


andrefurlan-db (approved):
looks good to me. Only thing is later we may want to use the information schema if that makes sense.

PR #242, opened 2022-12-23
Add "materialized_view" materialization.

ueshin:
### Description

Adds "materialized_view" materialization.

#### Prerequisite

- Only available in UC-enabled SQL Warehouses
- The warehouse type must be `PRO`
- The MVs must be registered in Unity Catalog
- [Delta CDF](https://docs.databricks.com/delta/delta-change-data-feed.html#enable-change-data-feed) (change data feed) must be enabled on the base table

#### How it works

- Create MVs for the first run
- Refresh the MVs for the iteration
- Create or replace the MVs when `--full-refresh` is specified.

For the iteration, even if the MV definition is updated in the dbt project, it won't replace the definition. `--full-refresh` should be used to update the MV definition in the cluster.


andrefurlan-db (commented):
we should try to check at compile-time any limitations/restrictions, or at least warn users, so that we fail fast. This leads to a better user experience overall.

ueshin:
If we want to check at compile-time any limitations/restrictions, we need to know the dependencies and their details, like configs or actual relations.
@jtcohen6 Is it possible to retrieve the info of dependencies from adapter side or macros in adapters?

PR #265, opened 2023-01-31
Ignore case sensitivity in relation matches method

andrefurlan-db:
### Description
If an user has a model name such as models/mf/sf/TNAME.sql, the second dbt run will fail with:
```
When searching for a relation, dbt found an approximate match. Instead of guessing 
12:24:19    which relation to use, dbt will move on. Please delete `cname`.`sname`.`tname`, or rename it to be less ambiguous.
12:24:19    Searched for: `cname`.`sname`.`TNAME`
12:24:19    Found: `cname`.`sname`.`tname`
12:24:19    
12:24:19    > in macro materialization_incremental_databricks (macros/materializations/incremental/incremental.sql)
12:24:19    > called by model TNAME (models/mf/sf/TNAME.sql)
```

This PR overwrites the relation.matches method and removes case sensitive checks.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/relation.py
+++ b/dbt/adapters/databricks/relation.py
@@ -91,8 +91,4 @@ class DatabricksRelation(BaseRelation):
             ):
                 match = False
 
-        if not match:
-            target = self.create(database=database, schema=schema, identifier=identifier)
-            raise ApproximateMatchError(target, self)
-
         return match
--- a/dbt/adapters/databricks/relation.py
+++ b/dbt/adapters/databricks/relation.py
@@ -8,7 +8,7 @@ from dbt.adapters.spark.impl import KEY_TABLE_OWNER, KEY_TABLE_STATISTICS
 
 from dbt.adapters.databricks.utils import remove_undefined
 from dbt.utils import filter_null_values
-from dbt.exceptions import ApproximateMatchError, DbtRuntimeError
+from dbt.exceptions import DbtRuntimeError
 
 KEY_TABLE_PROVIDER = "Provider"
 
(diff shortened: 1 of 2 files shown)


dbt/adapters/databricks/relation.py
```diff
@@ -61,3 +64,35 @@ def owner(self) -> Optional[str]:
...
+
+        for k, v in search.items():
+            if str(self.path.get_lowered_part(k)).strip(self.quote_character) != v.lower().strip(
+                self.quote_character
+            ):
+                match = False  # type: ignore[union-attr]
+
+        if not match:
+            target = self.create(database=database, schema=schema, identifier=identifier)
+            raise ApproximateMatchError(target, self)
```
ueshin, line 96: We don't need this part? Otherwise, it always raises the error when not match.

dbt/adapters/databricks/relation.py
```diff
@@ -61,3 +64,35 @@ def owner(self) -> Optional[str]:
...
+            # nothing was passed in
+            raise DbtRuntimeError("Tried to match relation, but no search path was passed!")
+
+        match = True
+
+        for k, v in search.items():
+            if str(self.path.get_lowered_part(k)).strip(self.quote_character) != v.lower().strip(
+                self.quote_character
+            ):
+                match = False  # type: ignore[union-attr]
```
ueshin, line 92: nit: Do we need `# type: ignore[union-attr]`?

dbt/adapters/databricks/relation.py
andrefurlan-db: Hum, what should be the expected behaviour? I am just following what was there before. Should we just return false?

ueshin: Yes, it should return false.

andrefurlan-db: okay

ueshin (approved):
LGTM, pending tests.

PR #270, opened 2023-02-15
Raise exception on unexpected error of list relations

### LINKED ISSUE #266: Table is overwritten if failed to retrieve tables list
### Describe the bug
We are running Databricks on AWS and using `dbt` with `dbt-databricks`.
The process is quite standard, we don't have any custom macros in our case, just running incremental model that writes to delta table.
Sometimes we receive error on step `show table extended in {{ relation }} like '*'` and it's infrastructure error.
Databricks logs are saying `Command failed because warehouse {warehouse_id} was stopped.`

The problem is that DatabricksAdapter is handling this exception and returns empty list of tables and further logic decides to `create or replace table ...` which leads to table being overwritten and all previous partitions are lost.

Workaround: is to revert table to previous version and run model again, but this requires constant monitoring that all our models are not accidentally overwritten.

### Steps To Reproduce
1) Run dbt model on Databricks AWS
2) Somehow emulate failure on `show table extended in {{ relation }} like '*'` step. Alternative would be to modify macros and raise exception manually:
```
{% macro spark__list_relations_without_caching(relation) %}
  {% call statement('list_relations_without_caching', fetch_result=True) -%}
    show table extended in {{ relation }} like '*'
    {% do exceptions.raise_database_error("Failed retrieve tables from database.") %}
  {% endcall %}

  {% do return(load_result('list_relations_without_caching').table) %}
{% endmacro %}
```

### Expected behavior
Exception should raised and allow process to fail.
Another option, retry few times to retrieve tables list and after couple of failures raise exception.

### System information
```
Core:
  - installed: 1.2.2
  - latest:    1.4.1 - Update available!

Plugins:
  - databricks: 1.2.2 - Update available!
  - spark:      1.2.0 - Update available!
```

macOS Ventura

Python 3.8.12

### Additional context
The problem is in line DatabricksAdapter [line 135](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/impl.py#L135). If replace `return []` with `raise e` for my case that would work, but I am not sure if anyone can rely on this behaviour.
As alternative, maybe create a flag in adapter config and raise exception on flag is True.
Let me know what do you think.

andrefurlan-db:
### Description
Because of a [AWS Glue issue](https://github.com/dbt-labs/dbt-spark/pull/54), it was set to handle any exception and returning empty list of tables. The problem is that further logic decides to create or replace table, which leads to table being overwritten and all previous partitions are lost.

I believe that the problem the previous fix solves is no where near as important to Databricks, and the problem it causes is very bad.

resolves #266

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -132,7 +132,7 @@ class DatabricksAdapter(SparkAdapter):
             else:
                 description = "Error while retrieving information about"
                 logger.debug(f"{description} {schema_relation}: {e.msg}")
-                return []
+                raise e
 
         return [
             self.Relation.create(
@@ -165,7 +165,7 @@ class DatabricksAdapter(SparkAdapter):
             else:
                 description = "Error while retrieving information about"
                 logger.debug(f"{description} {schema_relation.without_identifier()}: {e.msg}")
-                results = []
+                raise e
 
         relations: List[Tuple[DatabricksRelation, str]] = []
         for row in results:


andrefurlan-db:
Not ready at all yet.

susodapop (dismissed):

susodapop:
I think these changes should be committed to the 1.4.latest branch rather than main, but let's expedite now and we can clean it up later.

ueshin (approved):

PR #271, opened 2023-02-15
Specific dbt dependencies

andrefurlan-db:
### Description

Given changes from 1.4 to 1.5 in dbt, we need to be able to maintain specific dependency versions in order to run tests

### Checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,8 +5,8 @@ envlist = linter, unit
 [testenv:linter]
 basepython = python3
 commands =
-  {envpython} -m black --config black.ini --check .
-  {envpython} -m flake8 --select=E,W,F --ignore=E203,W503 --max-line-length=100 .
+  {envpython} -m black --config black.ini --check dbt tests
+  {envpython} -m flake8 --select=E,W,F --ignore=E203,W503 --max-line-length=100 dbt tests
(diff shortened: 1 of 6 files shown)


ueshin (approved):
We might need to use `1.4.*` for our version in https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/__version__.py, too?

Otherwise, LGTM as a temporary fix for the breakage.

tox.ini
```diff
@@ -5,8 +5,8 @@ envlist = linter, unit
 [testenv:linter]
 basepython = python3
 commands =
-  {envpython} -m black --config black.ini --check .
-  {envpython} -m flake8 --select=E,W,F --ignore=E203,W503 --max-line-length=100 .
+  {envpython} -m black --config black.ini --check dbt tests
+  {envpython} -m flake8 --select=E,W,F --ignore=E203,W503 --max-line-length=100 dbt tests
```
susodapop, line 9: Clarifying for posterity: this change is necessary if you develop with a virtualenv in the same directory as the source code. Otherwise `black` and `flake8` will lint and format the site-packages of your Python installation which is both unnecessary and time-consuming. Andre and I learned this the hard way.

susodapop (approved):
LGTM and works on my machine as well.

susodapop (approved):
LGTM.

PR #273, opened 2023-02-15
Fix test_grants to use the error class to check the error.

ueshin:
### Description

Fixes `test_grants` to use the error class to check the error.

There was a change on the error message when the specified privilege doesn't exist.
We should use the error class `INVALID_PARAMETER_VALUE` that is contained in the error message and more stable.


ueshin:
@andrefurlan-db This needs to be backported to `1.4.latest`, `1.3.latest`, and `1.2.latest`.

andrefurlan-db (approved):

PR #274, opened 2023-02-16
Add integration tests 

susodapop:
### Description

This pull request adds a Github Actions Workflow to run our integration tests. It _should_ run automatically on pull requests from branches on the base repository and has access to Github Secrets to gather credentials for running tests. But due to some issues with our cloud workspace setup, the tests will all fail right now. So for this PR we've set the workflow to only run on the `workflow_dispatch` event, which means a user must click a button to start the workflow. Once we square away the workspace setup with the DECO team we'll make a follow-up PR to run these automatically.

### Checklist


andrefurlan-db (approved):
This is a good start.

susodapop:
This PR has been on-hold while we fixed permissions issues on the SQL warehouse. We needed to `GRANT CREATE ON hive_metastore TO <service principal user id>`. I tested locally to confirm this works. It does! So just now I manually kicked-off a Github Actions workflow to run all our integration tests. If this passes, I will merge.

susodapop:
It's not presented in the "All checks have passed" badging at the bottom of _this PR_ because the change hasn't merged to `main` yet. But you can see the outcome of the action run here: https://github.com/databricks/dbt-databricks/actions/runs/4206303847/jobs/7549931337

susodapop:
Progress: all the functional tests are now passing but the integration tests fail. We presently invoke these tests manually with `pytest` in the Github action. Whereas during local development we use `tox`. 

We skipped `tox` on the github action initially because we couldn't figure out how to pass the environment secrets to `tox` in a way that didn't leak secret material in the GH action logs. By invoking `pytest` directly we avoided that issue. But since invoking `pytest` directly won't work for the integration tests we must now return to the problem of how to use `tox` in GH actions without leaking sensitive data.

andrefurlan-db (approved):
Looks good to me. We can integrate the rest after.

PR #280, opened 2023-02-24
Add regexp_instr

frankivo:
### Description

This PR adds an implementation for [regexp_instr](https://learn.microsoft.com/en-us/azure/databricks/sql/language-manual/functions/regexp_instr) as used in [dbt-expections](https://github.com/calogica/dbt-expectations#expect_column_values_to_match_regex).

### Checklist


frankivo:
I could use some help to add a test (and how to run these) and what to add in the CHANGELOG.

benc-db:
This feels like a more appropriate change to submit to dbt-expectations (this file: https://github.com/calogica/dbt-expectations/blob/main/macros/regex/regexp_instr.sql).  We can't maintain macros that exist primarily to support an external plugin.

frankivo:
I agree, but they don't support databricks. Thanks for the feedback anyway :)

PR #291, opened 2023-03-09
adding optimizeWrite and autoCompact by default

andrefurlan-db:
Resolves #

### Description

This PR automatically sets autoCompact and optimizeWrite for delta tables.

### Checklist


andrefurlan-db:
We ultimately decided that we do not want to set anything by default because:
1- Any default has some drawbacks that may complicate more than help
2- We are working on eliminating a lot of these knobs anyway, so in the future they wouldn't be needed.

PR #292, opened 2023-03-10
[feat] Support ZORDER as a model config

### LINKED ISSUE #122: Support ZORDER on models as model configuration
### Describe the feature
`ZORDER` is a useful way to get natural colocation for data. It can only be run as part of the [OPTIMIZE](https://docs.databricks.com/spark/latest/spark-sql/language-manual/delta-optimize.html) command. I would like to be able to set it as model configuration. In the implementation, we would run the OPTIMIZE command, which would use the model metadata to figure out the right ZORDER columns

### Who will this benefit?
Analytics engineers who want to keep model metadata in one location - the model itself.

andrefurlan-db:
Resolves #122

### Description
This PR augments the dbt model config with:
```python
config(
    materialized='incremental',
    zorder="column_A" | ["column_A", "column_B"]
)
```

Under the hood, after every model with zorder is build (initial or incremental), we'll run a `OPTIMIZE relation ZORDER BY ()` statement.

A var is available to skip OPTIMIZE if necessary: `databricks_skip_optimize`

One can set it in the `dbt-project.yaml` or directly in the comman line: `dbt run --models stg_payments --vars "{'databricks_skip_optimize': 'true'}"`

**Benefits:**
This simplifies the project as it requires one less post hook for maintenance operations.

#### Potential improvements
* Add predicates to OPTIMIZE statement. `OPTIMIZE table WHERE ...`
* Do not fail model if optimize fails: warn if optimize fails, but not error our

### Checklist

### DIFF
--- a/tests/integration/zorder/test_zorder.py
+++ b/tests/integration/zorder/test_zorder.py
@@ -1,7 +1,4 @@
 from tests.integration.base import DBTIntegrationTest, use_profile
-from dbt.tests.util import (
-    run_dbt_and_capture,
-)
 
 
 class TestZOrder(DBTIntegrationTest):
--- a/tests/integration/zorder/test_zorder.py
+++ b/tests/integration/zorder/test_zorder.py
@@ -1,0 +1,39 @@
+from tests.integration.base import DBTIntegrationTest, use_profile
+from dbt.tests.util import (
+    run_dbt_and_capture,
+)
+
+
+class TestZOrder(DBTIntegrationTest):
+    @property
+    def schema(self):
+        return "zorder"
+
+    @property
+    def models(self):
+        return "models"
+
+    @property
+    def project_config(self):
+        return {"config-version": 2}
+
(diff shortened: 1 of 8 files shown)


tests/integration/zorder/test_zorder.py
```diff
@@ -0,0 +1,35 @@
...
+    def schema(self):
+        return "zorder"
+
+    @property
+    def models(self):
+        return "models"
+
+    @property
+    def project_config(self):
+        return {"config-version": 2, "vars": {"databricks_skip_optimize": "True"}}
```
ueshin, line 15: Seems like we only test skipping optimization?

ueshin (approved):
Otherwise, LGTM.

tests/integration/zorder/test_zorder.py
andrefurlan-db: oh,. good catch. I added that temporarily and forgot to remove

PR #295, opened 2023-03-14
Integration tests: Update cloud env

susodapop:
### Description

DECO team now gives us the HTTP_PATH directly so we don't have to parse it from the JDBC connection URL.

### Checklist

- **NA** [x] I have run this code in development and it appears to resolve the stated issue
- **NA** [x] This PR includes tests, or tests are not required/relevant for this PR
- **NA** [x] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section.


andrefurlan-db (approved):
looks good. We will enable the tests once we have the users

PR #297, opened 2023-03-16
[feat] Support ZORDER as a model config (#292)

### LINKED ISSUE #122: Support ZORDER on models as model configuration

andrefurlan-db:
Resolves #122

### Description
This PR augments the dbt model config with:

```python
config(
    materialized='incremental',
    zorder="column_A" | ["column_A", "column_B"]
)
```
Under the hood, after every model with zorder is build (initial or incremental), we'll run a OPTIMIZE relation ZORDER BY () statement.

A var is available to skip OPTIMIZE if necessary: `databricks_skip_optimize`

One can set it in the dbt-project.yaml or directly in the comman line: `dbt run --models stg_payments --vars "{'databricks_skip_optimize': 'true'}"`

#### Benefits:
This simplifies the project as it requires one less post hook for maintenance operations.

#### Potential improvements
* Add predicates to OPTIMIZE statement. OPTIMIZE table WHERE ... 
* Do not fail model if optimize fails: warn if optimize fails, but not error out

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -59,6 +59,7 @@ class DatabricksConfig(AdapterConfig):
     options: Optional[Dict[str, str]] = None
     merge_update_columns: Optional[str] = None
     tblproperties: Optional[Dict[str, str]] = None
+    zorder: Optional[Union[List[str], str]] = None
 
 
 def check_not_found_error(errmsg: str) -> bool:
--- a/dbt/include/databricks/macros/adapters.sql
+++ b/dbt/include/databricks/macros/adapters.sql
@@ -111,6 +111,37 @@
   {% endif %}
 {% endmacro %}
 
+{% macro optimize(relation) %}
+  {{ return(adapter.dispatch('optimize', 'dbt')(relation)) }}
+{% endmacro %}
+
+{% macro databricks__optimize(relation) %}
+  {% if config.get('zorder', False) and config.get('file_format', 'delta') == 'delta' %}
+    {% if var('DATABRICKS_SKIP_OPTIMIZE', 'false')|lower != 'true' and var('databricks_skip_optimize', 'false')|lower != 'true' %}
+      {% call statement('run_optimize_stmt') %}
+        {{ get_optimize_sql(relation) }}
+      {% endcall %}
+    {% endif %}
+  {% endif %}
+{% endmacro %}
+
+{% macro get_optimize_sql(relation) %}
+  {% if config.get('zorder', False) and config.get('file_format', 'delta') == 'delta' %}
+     {%- set zorder = config.get('zorder', none) -%}
+    optimize {{ relation }}
+    {# TODO: predicates here? WHERE ...  #}
+    {% if zorder is sequence and zorder is not string %}
+      zorder by (
+        {%- for col in zorder -%}
+        {{ col }}{% if not loop.last %}, {% endif %}
+        {%- endfor -%}
+      )
+    {% else %}
+      zorder by ({{zorder}})
+    {% endif %}
+  {% endif %}
+{% endmacro %}
+
 {% macro alter_table_add_constraints(relation, constraints) %}
   {{ return(adapter.dispatch('alter_table_add_constraints', 'dbt')(relation, constraints)) }}
 {% endmacro %}
--- a/tests/unit/test_macros.py
+++ b/tests/unit/test_macros.py
@@ -17,17 +17,21 @@ class TestSparkMacros(unittest.TestCase):
         )
 
         self.config = {}
+        self.var = {}
         self.default_context = {
             "validation": mock.Mock(),
             "model": mock.Mock(),
             "exceptions": mock.Mock(),
             "config": mock.Mock(),
             "adapter": mock.Mock(),
+            "var": mock.Mock(),
             "return": lambda r: r,
         }
         self.default_context["config"].get = lambda key, default=None, **kwargs: self.config.get(
             key, default
         )
+
+        self.default_context["var"] = lambda key, default=None, **kwargs: self.var.get(key, default)
 
     def __get_template(self, template_filename):
         parent = self.parent_jinja_env.get_template(template_filename, globals=self.default_context)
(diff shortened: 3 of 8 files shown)


dbt/include/databricks/macros/adapters.sql
```diff
@@ -111,6 +111,37 @@
   {% endif %}
 {% endmacro %}
 
+{% macro optimize(relation) %}
+  {{ return(adapter.dispatch('optimize', 'dbt')(relation)) }}
+{% endmacro %}
```
susodapop, line 116: At first I wondered why these lines are here. It seems to bring the `dbt.optimize` macro into the current namespace so you can reference it on line 131?

For my own understanding: Jinja supports importing macros similar to Python (`{% import "../path/to/file" as <importable-name> %}`), but it looks like dbt doesn't use that syntax. Instead we "import" functions by making local macros that call `adapter.dispatch`. Does that sound correct?

dbt/include/databricks/macros/adapters.sql
```diff
@@ -111,6 +111,37 @@
...
+
+{% macro databricks__optimize(relation) %}
+  {% if config.get('zorder', False) and config.get('file_format', 'delta') == 'delta' %}
```
susodapop, line 119: Again I'm unfamiliar with how much of Jinja2 dbt uses, but customarily the `'delta'` string literal here would be a defined constant that's available to Jinja's context, so we can change it down the line. If that's not possible then this line is good.

dbt/include/databricks/macros/adapters.sql
```diff
@@ -111,6 +111,37 @@
...
+    {% if var('DATABRICKS_SKIP_OPTIMIZE', 'false')|lower != 'true' and var('databricks_skip_optimize', 'false')|lower != 'true' %}
```
susodapop, line 120: For readability can we check `== 'false'` rather than `!= 'true'` same as we do on line 119?

Also, why are there two configurations (uppercase and lowercase) for the same thing? Will this confuse users?

```suggestion
    {% if var('DATABRICKS_SKIP_OPTIMIZE', 'false')|lower == 'false' and var('databricks_skip_optimize', 'false')|lower == 'false' %}
```

dbt/include/databricks/macros/adapters.sql
```diff
@@ -111,6 +111,37 @@
...
+
+{% macro get_optimize_sql(relation) %}
+  {% if config.get('zorder', False) and config.get('file_format', 'delta') == 'delta' %}
+     {%- set zorder = config.get('zorder', none) -%}
+    optimize {{ relation }}
+    {# TODO: predicates here? WHERE ...  #}
+    {% if zorder is sequence and zorder is not string %}
+      zorder by (
+        {%- for col in zorder -%}
+        {{ col }}{% if not loop.last %}, {% endif %}
```
susodapop, line 136: As janky as this looks it's very common when writing Jinja FYI

dbt/include/databricks/macros/adapters.sql
```diff
@@ -111,6 +111,37 @@
...
+    {% endif %}
+  {% endif %}
+{% endmacro %}
+
+{% macro get_optimize_sql(relation) %}
+  {% if config.get('zorder', False) and config.get('file_format', 'delta') == 'delta' %}
+     {%- set zorder = config.get('zorder', none) -%}
+    optimize {{ relation }}
+    {# TODO: predicates here? WHERE ...  #}
+    {% if zorder is sequence and zorder is not string %}
```
susodapop, line 133: I would propose moving the check for `zorder is string` to line 139. `zorder is sequence` will not return True if `zorder` is a string. In the case where `zorder is not sequence` and is also not a string, Jinja will try to iterate over it and it will fail.

tests/unit/test_macros.py
```diff
@@ -17,18 +17,22 @@ def setUp(self):
...
             "config": mock.Mock(),
             "adapter": mock.Mock(),
+            "var": mock.Mock(),
             "return": lambda r: r,
         }
         self.default_context["config"].get = lambda key, default=None, **kwargs: self.config.get(
             key, default
         )
 
+        self.default_context["var"] = lambda key, default=None, **kwargs: self.var.get(key, default)
```
susodapop, line 34: Holy lambda, batman. not sure what this does here. But it's identical to line 30 apart from the `default_context` key so I suppose it's fine?

dbt/adapters/databricks/impl.py
```diff
@@ -59,6 +59,7 @@ class DatabricksConfig(AdapterConfig):
     options: Optional[Dict[str, str]] = None
     merge_update_columns: Optional[str] = None
     tblproperties: Optional[Dict[str, str]] = None
+    zorder: Optional[Union[List[str], str]] = None
```
susodapop, line 62: Is this configuration per-relation? Or per project?

susodapop (approved):
Overall looks good. I have some newbie questions. Nothing that blocks merge.

susodapop (changes requested):
You need to rebase off of `main` in order to pick up the full matrix of integration tests.

dbt/include/databricks/macros/adapters.sql
andrefurlan-db: this is some dbt magic to allow for inheritance and user overwrites.

dbt/include/databricks/macros/adapters.sql
andrefurlan-db: delta is specific to our adapter. Not something that can come from jinja. The context is created by dbt-core.

dbt/include/databricks/macros/adapters.sql
andrefurlan-db: not really. The default is to not set anything. So we need to check if the string is 'true', which means that the user actually wants to skip. If this was a boolean check it would be fine, but as a string I don't think it helps.

dbt/adapters/databricks/impl.py
andrefurlan-db: you can define dbt configs on a per project, per folder, or per model basis

dbt/include/databricks/macros/adapters.sql
andrefurlan-db: I don't understand. zorder should only be sequence or string. This same sort of jinja code is all over dbt and its macros and it works.

tests/unit/test_macros.py
andrefurlan-db: it is fine

susodapop (approved):
LGTM

Issue #232, opened 2022-12-05
Automatically start cluster for Python models


bilalaslamseattle:
### Describe the feature
The context API won't automatically start a cluster today. `create_notebook` uses a different method which does automatically start a cluster. This could be solved in the dbt-databricks adapter by checking if the cluster is active and starting it if it isn't.

boxysean:
Hi, I am experiencing this issue in my project. I understood your note to mean that, if I were to specify `create_notebook=True` in my `config()` block within my python model, then it would not start a cluster. However, I am now believing you mean that dbt python models are not able to auto-start a cluster because such execution creates a notebook, which requires a started cluster (and does not auto-start). Is that correct?

What do the solution look like? Which API call to the Databricks cluster -- do you envision a trigger and polling routine? Thanks!

PR #302, opened 2023-03-29
Copy python_submissions.py into dbt-databricks

susodapop:
Related to #232 which will be solved in three steps:

1. (This PR) Add `python_submissions.py` to dbt-databricks
2. Update `DBCommand` class to automatically start the cluster using the command API if it is not already started.
2. Remove `DbtDatabricksJobClusterPythonJobHelper` since this execution method isn't supported on Databricks anymore

### Description

This pull request copies the contents of dbt-spark's `python_submissions.py` directly into dbt-databricks. This is the path proposed in https://github.com/dbt-labs/dbt-spark/discussions/407 now that Python models have been released in dbt 1.4. It is a precursor to fixing #232, which will require modifying the `DBCommand` class to start an all-purpose cluster as needed.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/python_submissions.py
+++ b/dbt/adapters/databricks/python_submissions.py
@@ -10,7 +10,6 @@ import uuid
 
 import dbt.exceptions
 from dbt.adapters.base import PythonJobHelper
-from dbt.adapters.spark import SparkCredentials
 from dbt.adapters.spark import __version__
 
 DEFAULT_POLLING_INTERVAL = 10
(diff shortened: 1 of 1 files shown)


susodapop:
All integration tests pass.

dbt/adapters/databricks/python_submissions.py
```diff
@@ -1,14 +1,315 @@
...
 from dbt.adapters.databricks.connections import DatabricksCredentials
 
+import base64
+import time
+import requests
+import uuid
+
+import dbt.exceptions
+from dbt.adapters.base import PythonJobHelper
+from dbt.adapters.spark import SparkCredentials
```
andrefurlan-db, line 13: why do we need these dependencies?

andrefurlan-db: If there is no drawback, I'd rather decouple from dbt-spark where ever we can. 
Also, do we need to port tests? Or is this code tested on our side?

susodapop: Thanks @andrefurlan-db. Done! It was actually easier than I thought because we already made a `DatabricksCredentials` class -- I just needed to update the type hints and remove the unnecessary `SparkCredentials` import. All tests pass.

As for the tests: we already duplicated the dbt-spark tests on our side in the past so that work is done 🙌

PR #306, opened 2023-04-07
Auto start cluster for Python models

### LINKED ISSUE #232: Automatically start cluster for Python models

susodapop:
Resolves #232

### Description

This change follows #302, which imported some Databricks-specific code from dbt-spark directly into this repository. In this PR I implemented logic to check whether the cluster for Python models is running, start it if not, and wait for the cluster to start before proceeding.

Testing this change is necessarily manual, for now. The test uses a `dbt` profile that connects to a SQL warehouse and then tells the specific Python model to connect to an all purpose cluster which is in TERMINATED state at the beginning of the test. Our automated integration tests run on shared infrastructure (multiple PRs use the same environment) so we can't guarantee that the all-purpose cluster is TERMINATED automatically.

Therefore the integration test in this PR is skipped automatically. To run it, clone the repository, set `DBT_DATABRICKS_CLUSTER_HTTP_PATH` in `test.env` to a TERMINATED all-purpose cluster, and then run `python` integration test. It should pass and you will see that the cluster is started automatically. 

In a follow-up PR I can mock the cluster status API response and assert that the `start_cluster()` method was called. But a full integration test will require more work from the DECO team.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/python_submissions.py
+++ b/dbt/adapters/databricks/python_submissions.py
@@ -8,9 +8,12 @@ import time
 import requests
 import uuid
 
+from dbt.events import AdapterLogger
 import dbt.exceptions
 from dbt.adapters.base import PythonJobHelper
 from dbt.adapters.spark import __version__
+
+logger = AdapterLogger("Databricks")
 
 DEFAULT_POLLING_INTERVAL = 10
 SUBMISSION_LANGUAGE = "python"
@@ -241,6 +244,12 @@ class DBCommand:
         self.host = credentials.host
 
     def execute(self, context_id: str, command: str) -> str:
+
+        if self.get_cluster_status().get("status") in ["Terminated", "Terminating"]:
+            logger.debug(f"Cluster {self.cluster_id} is not running. Attempting to restart.")
+            self.start_cluster()
+            logger.debug(f"Cluster {self.cluster_id} is now running.")
+
         # https://docs.databricks.com/dev-tools/api/1.2/index.html#run-a-command
         response = requests.post(
             f"https://{self.host}/api/1.2/commands/execute",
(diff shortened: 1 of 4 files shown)


andrefurlan-db:
great!

tests/integration/python/test_python.py
```diff
@@ -0,0 +1,36 @@
...
+    @property
+    def project_config(self):
+        return {
+            "config-version": 2,
+            "vars": {"http_path": os.getenv("DBT_DATABRICKS_CLUSTER_HTTP_PATH")},
+        }
+
+    def python_exc(self):
+        self.run_dbt(["run"])
+        self.run_dbt(["run"])  # make sure it also run in incremental
```
andrefurlan-db, line 28: for the purpose of this test you can remove the second run

dbt/adapters/databricks/python_submissions.py
```diff
@@ -231,6 +241,59 @@ def destroy(self, context_id: str) -> str:
...
+            return time.time() - start_time
+
+        while get_elapsed() < MAX_CLUSTER_START_TIME:
+            status_response = self.get_cluster_status()
+            if status_response.get("state") == "Running":
+                return
+            else:
+                time.sleep(5)
+
+        raise dbt.exceptions.DbtRuntimeError(f"Cluster {self.cluster_id} restart timed out")
```
andrefurlan-db, line 295: maybe say after how long?

dbt/adapters/databricks/python_submissions.py
```diff
@@ -201,6 +204,13 @@ def __init__(
 
     def create(self) -> str:
         # https://docs.databricks.com/dev-tools/api/1.2/index.html#create-an-execution-context
```
andrefurlan-db, line 206: isnt' it better t use https://<host>/api/2.0/clusters/get ?

dbt/adapters/databricks/python_submissions.py
```diff
@@ -231,6 +241,59 @@ def destroy(self, context_id: str) -> str:
...
+
+        def get_elapsed() -> float:
+            return time.time() - start_time
+
+        while get_elapsed() < MAX_CLUSTER_START_TIME:
+            status_response = self.get_cluster_status()
+            if status_response.get("state") == "Running":
+                return
+            else:
+                time.sleep(5)
```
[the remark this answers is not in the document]
susodapop: This API doesn't. But the gateway probably does if it's overloaded. I think we can implement handling for 503 and 429 errors in a subsequent if this becomes an issue.

tests/integration/python/test_python.py
susodapop: Done in 4c767f4dd6be4f83b4974df6c7366fbf843d4572

dbt/adapters/databricks/python_submissions.py
susodapop: Fixed in 8d00df96915e42ffa5a0f7c2782f35ca332f8faf

dbt/adapters/databricks/python_submissions.py
susodapop: I followed the pre-existing convention in this file, which is to link the documentation for the API endpoint.

andrefurlan-db: that's now what I meant. I meant using the 2.0 API instead of the 1.2

dbt/adapters/databricks/python_submissions.py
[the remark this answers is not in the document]
andrefurlan-db: okay. Later on we can also do something a bit smarter an backoff a bit. But for now it seems okay

dbt/adapters/databricks/python_submissions.py
andrefurlan-db: ahh, sorry, my bad. This API is not something you added in this PR. This is the old behavior.

andrefurlan-db (approved):
looks good

susodapop:
🚀 🚀 🚀

mmenjivarST:
Hi, I've been testing this functionality today after upgrading dbt-databricks to 1.4.3 and it didn't work for me, the cluster was still starting when my python models failed with the error: 
```
Error starting terminated cluster.
14:51:11     b'{"error_code":"INVALID_STATE","message":"Cluster 0802-060324-ir9ydk9r is in unexpected state Pending.","details":[{"@type":"type.googleapis.com/google.rpc.ErrorInfo","reason":"CM_API_ERROR_SOURCE_CALLER_ERROR","domain":""}]}'
```

susodapop:
Thanks for the report @mmenjivarST. Do you see log messages `Cluster {self.cluster_id} is not running. Attempting to restart.` and `Cluster {self.cluster_id} is now running.`?

The INVALID_STATE error means that dbt-databricks attempted to start a cluster that was already started. But when dbt-databricks starts a cluster, it _blocks_ until the cluster goes from `Pending` to `Running` state before resuming the dbt workflow.

I wonder if we have a threading concern and need to adjust the logic so dbt-databricks not only checks for a TERMINATED cluster but also for a PENDING one. If it's TERMINATED then start it. If it's PENDING, then wait for it.

mmenjivarST:
Hi @susodapop, I don't see those logs, I think the cluster is not ready yet when the models fail. From my perspective looks like the blocking part is malfunctioning

PR #326, opened 2023-04-28
allow-table-extended-to-handle-more-than-2048-chars

josephberni:
Resolves [#326](https://github.com/databricks/dbt-databricks/pull/326)

### Description

This PR allows the adapter to handle strings of more than 2048 characters when running show tables extended in Databricks. This is currently preventing users from being able to run this command when using dbt if they have a large number of tables in any given schema. 

The way this is done is that it will pass the `*` in instances where the string length of the pipe separated tables is more than 2048 chars.

I have successfully tested both unit tests and `tox -e integration-databricks-sql-endpoint`.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## dbt-databricks 1.5.0 (Release TBD)
 
+### Fixes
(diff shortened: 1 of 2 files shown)


andrefurlan-db:
Thanks for the PR. I'll merge to staging-326 and run the integration tests.

PR #345, opened 2023-05-12
Added target schema and db to temp views

### LINKED ISSUE #344: `make_temp_relation` uses default database/schema instead of `target.database` and `target.schema`
Hi team, 
We use this method as part of our dbt package, and for Databricks users it fails as many of them run dbt with a user that doesn't have permission to write to the default database. 
This is also inconsistent with how other adapters implemented the method (using the target database and schema nad not just the identifier). 

[We solved it on our end](https://github.com/elementary-data/dbt-data-reliability/pull/348/files) but I believe you should consider changing this on the adapter as well. 

https://github.com/databricks/dbt-databricks/blob/4d6e1ad03c5ef2a1e7f2a949dadbf9f8f3b37ff9/dbt/include/databricks/macros/adapters.sql#L427

Maayan-s:
Resolves #344 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ### Features
 
-- When installing python libraries onto clusters, you can now specify an index_url (Thanks @casperdamen123) ([367](https://github.com/databricks/dbt-databricks/pull/367))
+- When installing python libraries onto clusters, you can now specify an index_url (Thanks @casperdamen123) ([#367](https://github.com/databricks/dbt-databricks/pull/367))
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


benc-db:
I've updated with a test and changelog.  If the integration tests pass, @susodapop please take a quick look tomorrow.

benc-db (approved):

benc-db:
> I've updated with a test and changelog. If the integration tests pass, @susodapop please take a quick look tomorrow.

Err, need to figure out how to run the integration tests for this.  Still working through the best way to integrate external PRs :/

rcypher-databricks (approved):

benc-db:
The integration tests fail, and unfortunately this is because Databricks is weird: https://docs.databricks.com/en/sql/language-manual/sql-ref-syntax-ddl-create-view.html

Basically temporary views cannot have 3 level names (or even 2 level names).

PR #357, opened 2023-06-07
Bump pysql to 2.6.1

susodapop:
### Description

Bumping `databricks-sql-connector` to version 2.6.1 will grant performance improvements due to adding support for HTTP 1.1 (keep alive) connections. Also will boost stability by adding a default socket timeout for all thrift rpcs which will benefit dbt cloud users.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,11 @@
 
-## dbt-databricks 1.5.2 TBD
+## dbt-databricks 1.5.3 TBD
+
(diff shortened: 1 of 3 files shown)


susodapop:
This change has actually already been deployed to Pypi as part of version 1.5.4, but we need to land it in `main` also.

benc-db (approved):

susodapop:
This has gone stale. Closing. We'll be bumping to 2.7.0 later today.

PR #369, opened 2023-06-20
Fix connection issue

### LINKED ISSUE #368: The start_cluster command in the python flow never returns successfully
### Describe the bug
The start_cluster command in the python flow never returns successfully

### Steps To Reproduce
Run a python model where the cluster has not started yet, watch as you see the following in the logs:

```
[0m10:23:55.871336 [debug] [Thread-1 (]: Databricks adapter: Cluster <cluster> is not running. Attempting to restart.
[0m10:23:55.871748 [debug] [Thread-1 (]: Databricks adapter: Sending restart command for cluster id <cluster>
[0m10:38:59.219720 [debug] [Thread-1 (]: Timing info for model.test.basic (execute): 10:23:55.693165 => 10:38:59.219158
[0m10:38:59.228408 [debug] [Thread-1 (]: Runtime Error in model basic (models/basic.py)
  Cluster <cluster> restart timed out after 900 seconds
```

This despite the cluster starting up in about 2-3 minutes.

### Expected behavior
The ability to recover from starting a terminated cluster when running python models

See above

### System information
```
Core:
  - installed: 1.5.1
  - latest:    1.5.1 - Up to date!

Plugins:
  - spark: 1.5.0 - Up to date!
```

Ventura 13.4

Python 3.11.3

### Additional context
Making this issue so that I can close it with a fix

benc-db:
Resolves #368 

### Description

Needed to change expected value from "Running" to "RUNNING".  Added unit test of same.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/python_submissions.py
+++ b/dbt/adapters/databricks/python_submissions.py
@@ -38,7 +38,9 @@ class BaseDatabricksHelper(PythonJobHelper):
 
     @property
     def cluster_id(self) -> str:
-        return self.parsed_model["config"].get("cluster_id", self.credentials.cluster_id)
+        return self.parsed_model["config"].get(
+            "cluster_id", self.credentials.cluster_id
+        )
 
     def get_timeout(self) -> int:
         timeout = self.parsed_model["config"].get("timeout", DEFAULT_TIMEOUT)
@@ -128,15 +130,20 @@ class BaseDatabricksHelper(PythonJobHelper):
                 "url": f"https://{self.credentials.host}/api/2.1/jobs/runs/get?run_id={run_id}",
                 "headers": self.auth_header,
             },
-            get_state_func=lambda response: response.json()["state"]["life_cycle_state"],
+            get_state_func=lambda response: response.json()["state"][
+                "life_cycle_state"
+            ],
             terminal_states=("TERMINATED", "SKIPPED", "INTERNAL_ERROR"),
             expected_end_state="TERMINATED",
-            get_state_msg_func=lambda response: response.json()["state"]["state_message"],
+            get_state_msg_func=lambda response: response.json()["state"][
+                "state_message"
+            ],
         )
 
         # get end state to return to user
         run_output = requests.get(
-            f"https://{self.credentials.host}" f"/api/2.1/jobs/runs/get-output?run_id={run_id}",
+            f"https://{self.credentials.host}"
+            f"/api/2.1/jobs/runs/get-output?run_id={run_id}",
             headers=self.auth_header,
         )
         json_run_output = run_output.json()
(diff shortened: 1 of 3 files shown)


benc-db:
Sorry about formatting going crazy.  Not sure what's different about my VS Code settings.

dbt/adapters/databricks/python_submissions.py
```diff
@@ -288,7 +287,7 @@ def get_elapsed() -> float:
 
         while get_elapsed() < MAX_CLUSTER_START_TIME:
             status_response = self.get_cluster_status()
-            if status_response.get("state") == "Running":
+            if status_response.get("state") == "RUNNING":
```
susodapop, line 290: Since this behaviour previously worked, what changed?. Did DBR change the capitalisation? If so, we need to check for `.get("state") in ["Running", "RUNNING"]` or similar. Thoughts?

benc-db: Has the behavior previously worked?  Do we have evidence of that?

benc-db: I'm fine to make the comparison more flexible :)

susodapop: Yep, the evidence is in the PR that implemented this behaviour a couple months ago: https://github.com/databricks/dbt-databricks/pull/306

Our integration test used to pass.

This wouldn't be the first time DBR change capitalisation and broke things for us, though. We had the same problem with `pysql` a few months ago 🤦

susodapop: > I'm fine to make the comparison more flexible :)

I think we should, since applying this fix will break things for anyone using older DBRs.

dbt/adapters/databricks/python_submissions.py
```diff
@@ -287,7 +300,7 @@ def get_elapsed() -> float:
 
         while get_elapsed() < MAX_CLUSTER_START_TIME:
             status_response = self.get_cluster_status()
-            if status_response.get("state") == "RUNNING":
+            if status_response.get("state").lower() == "running":
```
susodapop, line 290: hehe clever 😉

susodapop (dismissed):

dbt/adapters/databricks/python_submissions.py
benc-db: How much should we trust the api generally?  Like, do I need to provide a default in case state is not present?

susodapop: I wouldn't use a default because if the api changes without notice it's probably best to fail early.

benc-db:
Anything blocking signoff?

susodapop (approved):
LGTM

benc-db:
jenkins merge

andrefurlan-db (approved):

susodapop:
In the future let's prefer to squash and merge these changes onto `main` with a descriptive commit message 🙏

susodapop:
When I cherry-picked this onto `1.5.latest` I squashed it and wrote the following, as an example:

```
Fix: Python models auto start logic wasn't case insensitive (#369)

The code expected the status == "Running" when the status is "RUNNING"
```

PR #375, opened 2023-06-23
Refactoring macro tests to make them more modular

benc-db:
### Description

There was a ton of duplication in the existing macro tests, not to mention code style that needed improving.  I figured before adding more macro tests for my python model fixes, I should clean up the camp site.

I investigated moving to pytest fixtures for the reused data, but they work kind of awkwardly with unittest TestCases, and wasn't prepared to move off unittest for this PR.  In other words, these tests could still be improved more.

I have not updated the changelog.  I was unclear if refactoring merited an entry in the changelog.

### Checklist


susodapop (approved):
LGTM.

Agree that no changelog is needed here. Pytest fixtures _are_ useful but not necessary at present.

PR #377, opened 2023-06-27
Capture databricks-sql-connector logs in dbt.log

susodapop:
### Description

This pull request captures log messages from databricks-sql-connector and pipes them into dbt's AdapterLogger fixtures. The goal here is to increase our debugging speed. Til now, we've been flying blind whenever there are databricks-sql-connector issues because those log messages weren't captured anywhere when running `dbt`.

I just released databricks-sql-connector==2.7.0 which makes some convenience fixes to logging out of the connector. Specifically, the contents of thrift responses are now redacted from the main logger. This is to avoid writing secrets into log files.

This PR therefore has two parts:

1. Pin the newer databricks-sql-connector dependency
2. Capture the log messages and re-emit them

After this change, looking at `dbt.log` after a dbt run will include messages that look like this:

```
[0m10:20:10.029895 [debug] [ThreadPool]: databricks-sql-connector adapter: retry parameter: _retry_delay_min given_or_default 1.0
[0m10:20:10.030321 [debug] [ThreadPool]: databricks-sql-connector adapter: retry parameter: _retry_delay_max given_or_default 60.0
[0m10:20:10.030743 [debug] [ThreadPool]: databricks-sql-connector adapter: retry parameter: _retry_stop_after_attempts_count given_or_default 30
[0m10:20:10.031150 [debug] [ThreadPool]: databricks-sql-connector adapter: retry parameter: _retry_stop_after_attempts_duration given_or_default 900.0
[0m10:20:10.031555 [debug] [ThreadPool]: databricks-sql-connector adapter: retry parameter: _retry_delay_default given_or_default 5.0
[0m10:20:10.040874 [debug] [ThreadPool]: databricks-sql-connector adapter: Sending request: OpenSession(<REDACTED>)
[0m10:20:10.370359 [debug] [ThreadPool]: databricks-sql-connector adapter: Received response: TOpenSessionResp(<REDACTED>)
[0m10:20:10.372085 [info ] [ThreadPool]: databricks-sql-connector adapter: Successfully opened session 01ee14fe-1b0f-1233-a60e-36efb3903fee
[0m10:20:10.373849 [debug] [ThreadPool]: databricks-sql-connector adapter: Sending request: ExecuteStatement(<REDACTED>)
[0m10:20:11.099219 [debug] [ThreadPool]: databricks-sql-connector adapter: Received response: TExecuteStatementResp(<REDACTED>)
[0m10:20:11.102871 [debug] [ThreadPool]: SQL status: OK in 1.0700000524520874 seconds
[0m10:20:11.161322 [debug] [ThreadPool]: On list_schemas: Close
[0m10:20:11.162009 [info ] [ThreadPool]: databricks-sql-connector adapter: Closing session 01ee14fe-1b0f-1233-a60e-36efb3903fee
[0m10:20:11.162861 [debug] [ThreadPool]: databricks-sql-connector adapter: Sending request: CloseSession(<REDACTED>)
[0m10:20:11.254630 [debug] [ThreadPool]: databricks-sql-connector adapter: Received response: TCloseSessionResp(<REDACTED>)
```

### Checklist


susodapop:
Integration test check fails right now because of an unrelated model contract test.

susodapop:
The model contract test failed because dbt-core has updated their warning message from "Constraint validation failed" to "Contract enforcement failed" and we are checking for exact exception language:

https://github.com/dbt-labs/dbt-core/issues/7335

susodapop:
https://github.com/databricks/dbt-databricks/pull/378 fixed the failing integration test. I rebased these changes off main to capture that fix.

dbt/adapters/databricks/connections.py
```diff
@@ -58,6 +61,26 @@
...
+    def emit(self, record: logging.LogRecord) -> None:
+        # record.levelname will be debug, info, warning, error, or critical
+        # these map 1-to-1 with methods of the AdapterLogger
+        log_func = getattr(self.logger, record.levelname.lower())
+        log_func(record.msg)
+
+
+pysql_handler = DbtCoreHandler(level=0)
+
+pysql_logger = logging.getLogger("databricks.sql")
```
andrefurlan-db, line 79: don't we want this to be configured by an environment variable? Are you sure we always want to log debug level messages from sql connector?

susodapop: We can make it configurable. I believe the default for `dbt.log` is DEBUG which is why I kept it that way here.

dbt/adapters/databricks/connections.py
```diff
@@ -58,6 +61,27 @@
 
 logger = AdapterLogger("Databricks")
 
+
+class DbtCoreHandler(logging.Handler):
+    def __init__(self, level: Union[str, int], *args: Tuple, **kwargs: Mapping):
+        super().__init__(level=level)
+        self.logger = AdapterLogger("databricks-sql-connector")
```
benc-db, line 68: Two things here:
1. Why

*args: Tuple, **kwargs: Mapping

if you're not using them for anything?
2. Why not pass the logger in?  I'm generally not a fan of instantiating objects outside of a [composition root](https://blog.ploeh.dk/2011/07/28/CompositionRoot/)

dbt/adapters/databricks/connections.py
```diff
@@ -58,6 +61,27 @@
...
+        # these map 1-to-1 with methods of the AdapterLogger
+        log_func = getattr(self.logger, record.levelname.lower())
+        log_func(record.msg)
+
+
+pysql_logger_level = os.environ.get("DBT_DATABRICKS_CONNECTOR_LOG_LEVEL", "INFO").upper()
+pysql_handler = DbtCoreHandler(level=pysql_logger_level)
+
+pysql_logger = logging.getLogger("databricks.sql")
+pysql_logger.setLevel(logging.DEBUG)
```
benc-db, line 81: I'm confused why you default the pysql_handler to INFO (based on env variable), but then set the psyql_logger to DEBUG.

dbt/adapters/databricks/connections.py
```diff
@@ -58,6 +61,27 @@
 
 logger = AdapterLogger("Databricks")
 
+
+class DbtCoreHandler(logging.Handler):
+    def __init__(self, level: Union[str, int], *args: Tuple, **kwargs: Mapping):
+        super().__init__(level=level)
+        self.logger = AdapterLogger("databricks-sql-connector")
+
+    def emit(self, record: logging.LogRecord) -> None:
```
benc-db, line 70: No tests?

dbt/adapters/databricks/connections.py
benc-db: Does this only turn on the pysql_logger if the level is DEBUG, or does this overwrite what was done on line 78?

dbt/adapters/databricks/connections.py
susodapop: Works for me!

Great point about args and kwargs. Removed them. I revised the composition to inject the AdapterLogger rather than instantiating it within `__init__`.

dbt/adapters/databricks/connections.py
susodapop: Ugh, testing logs is a pain in the neck. Especially when they're configured by environment variables. I'll make a follow-up to this.

benc-db: See my last comment, I don't consider a blocker

dbt/adapters/databricks/connections.py
susodapop: No it doesn't overwrite anything. The logger and the handler each have a level setting. Sometimes a handler should send an email but only for messages at `CRITICAL` level. Whereas other handlers need to capture every possible log message (`DEBUG` level).

Changing the level of the logger itself means the events never even reach the handler if they don't meet or exceed its `level`.

The reason the code was different initially is that I figured we should just turn the pysql logger onto the loudest setting (`DEBUG`) and then use the environment variable to control which messages the DbtCoreHandler actually passes to its AdapterLogger.

It's not a major difference. So I changed both the logger and the handler to use the same level.

PR #379, opened 2023-06-27
Support more config options in python

### LINKED ISSUE #246: Python models don't respect config for properties, or incrementals
### Describe the bug
When running dbt with two python models, one table and one incremental referencing that table, it fails to apply several configurations. 

1. location_root and partition_by are ones I tested for a table 
2. Incremental models seem to be recognized as incremental, but the logic `dbt.is_incremental` fails to get set properly.

I think this is related to the macro `py_write_table`, which seems to hardcode the dataframe writer. 

This does not happen with sql models, and with a custom macro to add the location it does save the table to azure data lake. 

### Steps To Reproduce
You should just be able to set up a basic project and use these models, I am using Azure Storage, not sure if it happens with S3. 
The models should compile and run just fine, and with the custom macro I wrote, it will save. But you will see that none of the other config properties are accounted for. 

I am attaching the `run` output code.

First model: 

```
from pyspark.sql.types import *
from datetime import date


def model(dbt, session):
    dbt.config(
        materialized='table',
        partition_by="AirportCode",
        unique_key="TempRecordId"
    )
    file_format = dbt.config.get("file_format")
    location_root = dbt.config.get("location_root")
    partition_by = dbt.config.get("partition_by")

    schema = StructType([
        StructField('TempRecordId', IntegerType(), False),
        StructField('AirportCode', StringType(), False),
        StructField('DateRecorded', DateType(), False),
        StructField('TempHighF', IntegerType(), False),
        StructField('TempLowF', IntegerType(), False)
    ])

    data = [
        [1, 'BLI', date(2021, 4, 3), 52, 43],
        [2, 'BLI', date(2021, 4, 2), 50, 38],
        [3, 'BLI', date(2021, 4, 1), 52, 41],
        [4, 'PDX', date(2021, 4, 3), 64, 45],
        [5, 'PDX', date(2021, 4, 2), 61, 41],
        [6, 'PDX', date(2021, 4, 1), 66, 39],
        [7, 'SEA', date(2021, 4, 3), 57, 43],
        [8, 'SEA', date(2021, 4, 2), 54, 39],
        [9, 'SEA', date(2021, 4, 1), 56, 41]
    ]

    df = session.createDataFrame(data, schema)

    return df
```

Incremental model:
```
import pyspark.sql.functions as F


def model(dbt, session):
    dbt.config(
        materialized='incremental',
        partition_by="AirportCode",
        unique_key="TempRecordId",
        merge_update_columns=['AirportCode', 'DateRecorded', 'TempHighF', 'TempLowF']
    )
    file_format = dbt.config.get("file_format")
    location_root = dbt.config.get("location_root")
    partition_by = dbt.config.get("partition_by")
    unique_key = dbt.config.get("unique_key")
    merge_update_columns = dbt.config.get("merge_update_columns")

    df = dbt.ref('my_python_model')

    if dbt.is_incremental:
        df = df.where('DateRecorded >= "2021-04-03"')

    return df

```
[py_example.zip](https://github.com/databricks/dbt-databricks/files/10355264/py_example.zip)

schema.yml 

```
version: 2

models:
  - name: my_python_model
    config:
      file_format: "delta"
      location_root: "abfss://<container>@<storage_account>.dfs.core.windows.net/data-lake/dbt_test/my_python_dbt_model"
      packages:
        - "numpy==1.23.1"

  - name: my_second_python_model
    config:
      file_format: "delta"
      location_root: "abfss://<container>@<storage_account>.dfs.core.windows.net/data-lake/dbt_test/my_second_python_model"
      packages:
        - "numpy==1.23.1"
```

example macro that was able to work ( this was a quick fix ) 
```
{% macro py_write_table(compiled_code, target_relation) %}
{{ compiled_code }}
# --- Autogenerated dbt materialization code. --- #
dbt = dbtObj(spark.table)
df = model(dbt, spark)

partition_by = dbt.config.get('partition_by')
location_root = dbt.config.get('location_root')

# make sure pyspark exists in the namepace, for 7.3.x-scala2.12 it does not exist
import pyspark
# make sure pandas exists before using it
try:
  import pandas
  pandas_available = True
except ImportError:
  pandas_available = False

# make sure pyspark.pandas exists before using it
try:
  import pyspark.pandas
  pyspark_pandas_api_available = True
except ImportError:
  pyspark_pandas_api_available = False

# make sure databricks.koalas exists before using it
try:
  import databricks.koalas
  koalas_available = True
except ImportError:
  koalas_available = False

# preferentially convert pandas DataFrames to pandas-on-Spark or Koalas DataFrames first
# since they know how to convert pandas DataFrames better than `spark.createDataFrame(df)`
# and converting from pandas-on-Spark to Spark DataFrame has no overhead
if pyspark_pandas_api_available and pandas_available and isinstance(df, pandas.core.frame.DataFrame):
  df = pyspark.pandas.frame.DataFrame(df)
elif koalas_available and pandas_available and isinstance(df, pandas.core.frame.DataFrame):
  df = databricks.koalas.frame.DataFrame(df)

# convert to pyspark.sql.dataframe.DataFrame
if isinstance(df, pyspark.sql.dataframe.DataFrame):
  pass  # since it is already a Spark DataFrame
elif pyspark_pandas_api_available and isinstance(df, pyspark.pandas.frame.DataFrame):
  df = df.to_spark()
elif koalas_available and isinstance(df, databricks.koalas.frame.DataFrame):
  df = df.to_spark()
elif pandas_available and isinstance(df, pandas.core.frame.DataFrame):
  df = spark.createDataFrame(df)
else:
  msg = f"{type(df)} is not a supported type for dbt Python materialization"
  raise Exception(msg)

if location_root:
    ( df.write.mode("overwrite")
        .format("delta")
        .option("overwriteSchema", "true")
        .option("path", location_root)
        .saveAsTable(
            name="{{ target_relation }}"
        )
    )
else:
    df.write.mode("overwrite").format("delta").option("overwriteSchema", "true").saveAsTable("{{ target_relation }}")
{%- endmacro -%}

{%macro py_script_comment()%}
# how to execute python model in notebook
# dbt = dbtObj(spark.table)
# df = model(dbt, spark)
{%endmacro%}
```

### Expected behavior
I would expect the compiled files to account for the configuration properties accordingly. 

### System information
```
Core:
  - installed: 1.3.0
  - latest:    1.3.2 - Update available!

Plugins:
  - databricks: 1.3.2 - Up to date!
  - spark:      1.3.0 - Up to date!

```

```
WindowsBuildLabEx                                    : 19041.1.amd64fre.vb_release.191206-1406
WindowsCurrentVersion                              : 6.3
WindowsEditionId                                        : Enterprise
WindowsProductName                                : Windows 10 Enterprise
```

`Python 3.8.0`

benc-db:
Resolves #246

### Description

Adding support for the missing python model config options

### Checklist


tests/unit/macros/base.py
```diff
@@ -49,5 +49,8 @@ def dispatch(macro_name, macro_namespace=None, packages=None):
 
         self.default_context["adapter"].dispatch = dispatch
 
-        value = getattr(self.template.module, name)(*args)
+        return getattr(self.template.module, name)(*args)
+
+    def _run_macro(self, name, *args):
```
susodapop, line 54: I'm not super familiar with the dbt macro system. But this is simply a refactor, right? It doesn't change the behavior of the `_run_macro()` method -- only decomposes such that each function (_run_macro and _run_macro_raw) does just one thing. Right?

dbt/include/databricks/macros/python.sql
```diff
@@ -0,0 +1,77 @@
...
+        .mode("overwrite")
+        .option("overwriteSchema", "true")
+{{ py_get_writer_options()|indent(8, True) }}
+)
+
+writer.saveAsTable("{{ target_relation }}")
+{% endmacro %}
+
+{%- macro py_get_writer_options() -%}
+{%- set location_root = config.get('location_root', validator=validation.any[basestring]) -%}
```
susodapop, line 47: Can we set a default of `None` or `""` here? Or are we now requiring users to always include a location root in their Python models?

This is important, I think, because doing so will break backwards compatibility and we'll need to make a major version change.

dbt/include/databricks/macros/python.sql
```diff
@@ -0,0 +1,77 @@
...
+{%- set file_format = config.get('file_format', validator=validation.any[basestring])|default('delta', true) -%}
+{%- set partition_by = config.get('partition_by', validator=validation.any[list, basestring]) -%}
+{%- set clustered_by = config.get('clustered_by', validator=validation.any[list, basestring]) -%}
+{%- set buckets = config.get('buckets', validator=validation.any[int]) -%}
```
susodapop, line 51: General question: how often do we get new configs? Will this file need to be revised whenever new configs become available from dbt?

dbt/include/databricks/macros/python.sql
```diff
@@ -0,0 +1,77 @@
+{% macro databricks__py_write_table(compiled_code, target_relation) %}
```
susodapop, line 1: This file is called `python.sql` but it writes Python syntax. Can we change its extension to `.py`? Or is the `.sql` file-type required by dbt?

tests/integration/python/test_python.py
```diff
@@ -20,16 +20,27 @@ def models(self):
...
-            "vars": {"http_path": os.getenv("DBT_DATABRICKS_CLUSTER_HTTP_PATH")},
+            "vars": {
+                "http_path": os.getenv("DBT_DATABRICKS_CLUSTER_HTTP_PATH"),
+                "location_root": "",
+            },
         }
 
     def python_exc(self):
-        self.run_dbt(["run"])
+        self.run_dbt(["run", "-s", "basic"])
```
susodapop, line 30: Assuming we can handle a default null `location_root`, this line feels like the only necessary change to `test_python.py` in this pull request.

tests/unit/macros/test_adapters_macros.py
```diff
@@ -213,21 +212,6 @@ def setUp(self):
 
         self.relation = DatabricksRelation.from_dict(data)
 
-    def __run_macro2(self, template, name, relation, *args):
```
susodapop, line 216: Was this code being used anywhere? Or is this just cleaning up unused cruft?

 I see its the same as `_run_macro` but I don't see that any references to `__run_macro2` were removed in this PR.

susodapop (commented):
My only big concern is about a default for the `location_root` config. If this is now required then it's technically a breaking change.

tests/integration/python/models/models.yml
```diff
@@ -3,5 +3,19 @@ models:
...
+    config:
+      marterialized: table
+      tags: ["python"]
+      http_path: '{{ var("http_path") }}'
+      location_root: '{{ var("location_root") }}'
+    columns:
+      - name: date
+        tests:
+          - not_null
+
```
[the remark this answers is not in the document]
benc-db: I was following this guideline from the dbt style guide (which we have adapted to some extent in the Partner Connect project):

"Use a new line to separate list items that are dictionaries, where appropriate."

But, so long as its consistent with our other test projects, I guess its reasonable to drop the extra line break.

tests/integration/python/test_config.py
```diff
@@ -0,0 +1,34 @@
+from tests.integration.base import DBTIntegrationTest, use_profile
+import os
+import pytest
+
+
+@pytest.mark.skip(
+    reason="Run manually. Test must start with the Python compute\
+          resource in TERMINATED or TERMINATING state"
+)
```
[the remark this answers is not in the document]
benc-db: Oh, I thought that the python tests in general could collide with other tests.  Ok.  That means I actually do need to figure out a location_root solution for Github action :/.

tests/integration/python/test_python.py
```diff
@@ -20,16 +20,27 @@ def models(self):
     def project_config(self):
         return {
             "config-version": 2,
-            "vars": {"http_path": os.getenv("DBT_DATABRICKS_CLUSTER_HTTP_PATH")},
+            "vars": {
+                "http_path": os.getenv("DBT_DATABRICKS_CLUSTER_HTTP_PATH"),
+                "location_root": "",
```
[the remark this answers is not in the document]
benc-db: Yes, because dbt evaluates the whole project for Jinja compilation, so if this is missing it will fail, even though the model we're building doesn't need it.  Note: this is specific to this test, because of the var interpolation, not something that users would hit.  I could try to come up with a better solution for defaulting the location_root variable, but basically all of the vars need to be defined in order to compile.

tests/integration/python/test_python.py
```diff
@@ -20,16 +20,27 @@ def models(self):
...
     def test_python_databricks_sql_endpoint(self):
         self.python_exc()
 
     @use_profile("databricks_uc_sql_endpoint")
     def test_python_databricks_uc_sql_endpoint(self):
+        self.use_default_project(
+            {
+                "vars": {
+                    "http_path": os.getenv("DBT_DATABRICKS_UC_CLUSTER_HTTP_PATH"),
+                    "location_root": "",
```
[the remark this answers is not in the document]
benc-db: So, this was a change I needed to make in order to get the two tests to pass when I had a UC cluster and a non-UC cluster.  My understanding is that they were passing before because the cluster that was pointed at by DBT_DATABRICKS_CLUSTER_HTTP_PATH in the integration environment (and you and Andre's local test env) was UC enabled.

tests/unit/macros/base.py
benc-db: yes, decomposed because the space stripping is used in every test of test_adapters_macros, but for the test_python_macros tests, I actually care about spacing (because python is like that :/)

tests/unit/macros/test_adapters_macros.py
benc-db: I missed it in the refactor PR I sent out.  Unused cruft at this point.

tests/unit/test_adapter.py
```diff
@@ -638,7 +645,10 @@ def test_parse_relation_with_statistics(self):
             ("Location", "/mnt/vo"),
             ("Serde Library", "org.apache.hadoop.hive.serde2.lazy.LazySimpleSerDe"),
             ("InputFormat", "org.apache.hadoop.mapred.SequenceFileInputFormat"),
-            ("OutputFormat", "org.apache.hadoop.hive.ql.io.HiveSequenceFileOutputFormat"),
+            (
+                "OutputFormat",
+                "org.apache.hadoop.hive.ql.io.HiveSequenceFileOutputFormat",
+            ),
```
benc-db, line 651: I would love to figure out what's different in my VS Code settings from you guys...I don't mean to reformat random files in every PR :)

dbt/include/databricks/macros/python.sql
benc-db: These are options that we (Databricks) support for SQL models.  They also are valid for Python models, but the upstream code in dbt-spark was not flexible enough to support them.

benc-db: So, they'll update based on changes in Spark, whenever we want this library to support said changes.

dbt/include/databricks/macros/python.sql
benc-db: It already defaults to Jinja none.  No changes required of users.  That's part of what is ensured by running test_python.py, that things continue to build even if none of the new variables are provided.

dbt/include/databricks/macros/python.sql
benc-db: the .sql file is a dbt standard (even for python code), but I haven't tried it as a different file extension.  Personally, I would have named the files something like adapter.sql.template or adapter.sql.jinja.  I'll see what happens if I use a different extension.

benc-db: Did not work with me changing the extension...without understanding a lot more about how the macro engine works, I think we have to stick with .sql.

tests/integration/python/test_python.py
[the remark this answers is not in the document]
benc-db: Point is moot.  I'm dropping location_root from the integration test as too costly to maintain.  I've manually validated that it works E2E, and we can get coverage of this line through unit testing the macro.

tests/integration/python/test_config.py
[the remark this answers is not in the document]
benc-db: Decided against testing location root in the integration test.

susodapop:
I rebased off of `main` and updated the changelog

PR #382, opened 2023-06-30
Synchronise changes from 1.5.latest back to main

### LINKED ISSUE #381: Synchronise changes from 1.5.latest branch back to main
### Describe the bug
While expediting fixes for customers, we've implemented a few changes directly on the `1.5.latest` branch in git that we need to sync back to `main` before they get too numerous to reason about.

This is most notable because the CHANGELOG on `main` extends only to version 1.5.1 whereas the `1.5.latest` branch shows 1.5.4

PR incoming to fix this.

susodapop:
### Note to reviewers

This PR uses commits with good commit messages to explain each step of the process. It will be easiest to review commit-by-commit.

Resolves #381

### Description

The way I made this PR was to compare the tip of `1.5.latest` to the tip of `main` and only look at changed files that _haven't been touched by a PR since the last release (1.5.4)_.

This pull request uses several commits with relevant comments to explain what changes I made. The bulk of this is just updating the CHANGELOG on `main`.

There are three commits that actually affect code execution:

- [ ] 4866583c741f9ce149dbca62a300e37d18c9ea4a Set a default socket timeout of 180 seconds
- [ ] 1f9f269c258c90d7da4c86b9bc54b0402a0aa900 pin to minor version of dbt-spark
- [ ] 579a4a99ca8f4f03127d4c713865ba3289930fc6 pin to exact version of databricks-sdk

These three exact changes have been part of existing dbt-databricks releases and therefore should be safe to release now.

Everything else affects only the CHANGELOG or our development setup, which won't impact customers.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@
 - Include log events from databricks-sql-connector in dbt logging output.
 - Adapter now populates the `query_id` field in `run_results.json` with Query History API query ID.
 
+## dbt-databricks 1.5.4 (June 9, 2023)
+
+### Features
+- Added support for model contracts ([#336](https://github.com/databricks/dbt-databricks/pull/336))
+
+## dbt-databricks 1.5.3 (June 8, 2023)
+
+### Fixes
+- Pins dependencies to minor versions
+- Sets default socket timeout to 180s
+
+## dbt-databricks 1.5.2 (May 17, 2023)
+
+### Fixes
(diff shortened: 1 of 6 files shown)


rcypher-databricks (approved):

marijncv:
Hey @susodapop, out of curiosity, why was the databricks-sdk version hard pinned to 0.1.7 in this PR? 

For my usage this causes a dependency conflict when also trying to install `requests==2.31.0` because databricks-sdk 0.1.7 has requests pinned to `<2.29`.

Curious for your thoughts!

PR #389, opened 2023-07-17
Added tests for materialized view/streaming table

rcypher-databricks:
Added integration tests for materialized view and streaming table materializations.  Tests for using a view or table without changed data feed as a base are currently disabled as dbsql is not currently returning an error in these situations.
Signed-off-by: Raymond Cypher <raymond.cypher@databricks.com>


susodapop (approved):
My level of knowledge on streaming tables is pretty low but this test code looks fine.

PR #390, opened 2023-07-19
Support for materialized views and streaming tables

rcypher-databricks:
Two new materialization types were added (materialized_view and streaming_table) to access DBSQL support for Delta Live Tables .  These allow models to be materialized as streaming tables (ST) or materialized views (MV).
- On the first run of the model the MV/ST will be created. Subsequent runs will result in the MV/ST being refreshed, rather than recreated.  Running dbt with 'full-refresh' will cause the MV/ST to be dropped and recreated.
- DBSQL MV/ST creates are synchronous operations on the DBSQL warehouse. DBSQL MV/ST refreshes are asynchronous. When a DBSQL MV/ST refresh is initiated a background job (pipeline) is initiated on a serverless DLT cluster and the command returns immediately before the data load is completed.
	- Modified DatabricksSQLCursorWrapper to recognize the case of an MV/ST being refreshed.
	- When the refresh command returns use the unity catalog REST API to determine the id of the refresh pipeline
	- Poll the pipeline until the refresh completes/fails using the pipeline REST API
	- In the case of a refresh failure use the pipeline REST API to retrieve any events associated with the pipeline, search for the failure event, and generate an error message based on the event.

Materialized views:

To create a materialized view set the materialized property of the model to 'materialized_view'.  On the first run of the model the materialized view will be created. Subsequent runs will result in the materialized view being refreshed, rather than recreated.  Running dbt with 'full-refresh' will cause the view to be dropped and recreated.
```
{{ config(
    materialized = 'materialized_view',
    tblproperties = {
'delta.enableChangeDataFeed': 'true'}
) }}
SELECT
  i.time_stamp,
  u.risk,
  u.age,
  avg(i.calories_burnt) as avg_calories_burnt, 
  COUNT(*) AS num_records,
  AVG(i.num_steps) AS avg_num_steps,
  AVG(i.miles_walked) AS avg_miles_walked
FROM
  {{ ref('devices') }} i
  JOIN {{ ref('users') }} u
  ON i.user_id = u.userid
GROUP BY
ALL
```

Queries supported for incremental refresh 

- A materialized view may query only a single table or  INNER JOIN, UNION ALL (or combinations of INNER JOIN and UNION ALL) on multiple tables.
- A materialized view must have a GROUP BY.
- The aggregate functions that are supported in the SELECT clause in materialized views are:
	- SUM
	- COUNT
	- AVG
	- STDEV
- Any aggregate function that is not explicitly included in the above section are not supported
- Incremental refresh is not supported for MVs that include:
	- Window functions,
	- HAVING clauses,
	- Subqueries in SELECT or WHERE clauses
- CDF is required on the base table.  This can be done by setting the delta.enableChangeDataFeed table property 
```
{{ config(
    tblproperties = {
        'delta.enableChangeDataFeed': 'true'
    }
) }}
```

General limitations. 

The following queries will not be incrementally refreshed when defined in an MV:
- Non-deterministic queries. Functions used in a materialized view must be deterministic. For example, using CURRENT_TIMESTAMP is not permitted.
- Queries that use time travel. Materialized views cannot be created using the delta lake time travel feature.
- LEFT JOINs and OUTER JOINs are not supported 

Streaming tables:

- Streaming tables require streaming sources. Use the STREAM keyword to stream changes from read_files or other Delta tables.
```
{{ config(
    materialized = 'streaming_table',
) }}
SELECT id AS id, msg AS msg FROM STREAM {{ ref('base') }}
```

- The new read_files table valued function lets you read files in external locations. File format and data schema are automatically inferred if not provided.
```
{{ config(materialized='streaming_table') }}
SELECT CAST(userid AS INTEGER) as userid,
          gender,
          CAST(age AS INTEGER) as age,
          CAST(height AS INTEGER) as height,
          CAST(weight AS INTEGER) as weight,
          smoker,
          familyhistory,
          cholestlevs,
          bp,
          CAST(risk AS INTEGER) as risk
   FROM STREAM read_files("dbfs:/.../data-user/*");
```


andrefurlan-db (approved):
looks good to me. Make sure you release RC1 and let's test well before going to production

PR #397, opened 2023-07-27
dbt clone support

benc-db:
### Description

Adds support for dbt clone.  I grabbed the materialization from dbt-spark to remove the filter for delta as the specified format, as Databricks can pass all of the fixture tests with shallow clone without setting file_format.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## dbt-databricks 1.6.0 (Release TBD)
 
 ### Features
+
 - Added support for materialized_view and streaming_table materializations
 - Support new dbt `limit` command-line flag
+- Support dbt clone operation
 
 ## dbt-databricks 1.5.5 (July 7, 2023)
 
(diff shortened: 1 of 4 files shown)


dev-requirements.txt
```diff
@@ -22,7 +22,6 @@ pytz
 tox>=3.2.0
 types-requests
 
-dbt-spark==1.6.0rc1
```
benc-db, line 25: I can revert if you like.  This was just redundant with requirements.txt.  We can rewrite dev-requirements so that it installs requirements when its installed, btw.

rcypher-databricks (approved):

CHANGELOG.md
susodapop, line 1: what is going on with your auto format settings? Can you disable it so we don't keep trying to rewrite the changelog?

susodapop (approved):
LGTM

It's fine to reformat the changelog. It just doesn't seem like you intended to include those changes in this PR. Autoformat is a nightmare sometimes...

PR #398, opened 2023-07-28
Liquid Clustering config for table materialization

ammarchalifah:
Resolves https://github.com/databricks/dbt-databricks/issues/399

### Description

With this change, dbt user could supply `liquid_clustered_by` model config with column (or list of columns) to be used as clustering keys using **Liquid Clustering** feature of Delta. The change introduces a small change in DDL query, while it does nothing for `python`-based models (as I couldn't find the python API for **Liquid Clustering**).

### Checklist

### DIFF
--- a/tests/unit/macros/test_adapters_macros.py
+++ b/tests/unit/macros/test_adapters_macros.py
@@ -8,9 +8,7 @@ class TestAdaptersMacros(TestMacros):
         super().setUp()
         self.template = self._get_template("adapters.sql")
 
-    def _render_create_table_as(
-        self, relation="my_table", temporary=False, sql="select 1"
-    ):
+    def _render_create_table_as(self, relation="my_table", temporary=False, sql="select 1"):
         self.default_context["model"].alias = relation
 
         return self._run_macro("databricks__create_table_as", temporary, relation, sql)
@@ -20,9 +18,7 @@ class TestSparkMacros(TestAdaptersMacros):
     def test_macros_create_table_as(self):
         sql = self._render_create_table_as()
 
-        self.assertEqual(
-            sql, "create or replace table my_table using delta as select 1"
-        )
+        self.assertEqual(sql, "create or replace table my_table using delta as select 1")
 
     def test_macros_create_table_as_file_format(self):
         for format in ["parquet", "hudi"]:
@@ -115,8 +111,7 @@ class TestSparkMacros(TestAdaptersMacros):
 
         self.assertEqual(
             sql,
-            "create or replace table my_table "
-            "using delta cluster by (cluster_1) as select 1",
+            "create or replace table my_table " "using delta cluster by (cluster_1) as select 1",
         )
 
     def test_macros_create_table_as_liquid_clusters(self):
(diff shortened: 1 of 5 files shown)


ammarchalifah:
Tested on private Databricks cluster, works as intended. Resulting table is clustered correctly.

Screenshot 2023-07-28 at 19 00 55
Screenshot 2023-07-28 at 19 01 34

ammarchalifah:
Hi, could you help review this PR? @andrefurlan-db @susodapop @ueshin

tests/unit/macros/test_adapters_macros.py
```diff
@@ -154,6 +176,7 @@ def test_macros_create_table_as_all_delta(self):
             "create or replace table my_table "
             "using delta "
             "partitioned by (partition_1,partition_2) "
+            "cluster by (cluster_1,cluster_2)"
```
susodapop: From the [docs](https://learn.microsoft.com/en-us/azure/databricks/delta/clustering#enable-liquid-clustering) (emphasis mine):

> Clustering is not compatible with **partitioning** or ZORDER, and requires that the Azure Databricks client manages all layout and optimization operations for data in your table.

Therefore the output of this `self._render_create_as()`, while expected, would technically raise an exception. Should we enforce any kind of block or just allow the compute cluster to raise an exception if the dbt user configures both `partition_by` and `liquid_clustered_by`?

susodapop (commented):
Change looks good with one question 👍 Thanks for this contribution.

tests/unit/macros/test_adapters_macros.py
ammarchalifah: That's a good question. I'm for leaving the compute cluster to raise an exception, but would be open for suggestion

benc-db: My stance has been to let Databricks surface the issues, rather than intercepting in the adapter.  I'm not sure if we're doing this consistently everywhere.  My reasoning is that, for any rule we hard code in, the Databricks implementation could change, and it would be better for that functionality to be available to customers without us having to spin a new release.

benc-db:
@ammarchalifah can you file an issue for us to add this capability for python models as well?  The python model support is currently in flux, so no need for you to make the change, but I would appreciate if you capture the expected behavior (matching your sql tests) in an issue.

ammarchalifah:
> @ammarchalifah can you file an issue for us to add this capability for python models as well? The python model support is currently in flux, so no need for you to make the change, but I would appreciate if you capture the expected behavior (matching your sql tests) in an issue.

Yes, opened a new issue [here](https://github.com/databricks/dbt-databricks/issues/411).

susodapop (approved):
Thanks again! I've updated this to merge to a staging branch on the main repository. This enables the staging branch to run our e2e tests. Once those pass, we'll merge to `main`

ammarchalifah-bolt:
> Thanks again! I've updated this to merge to a staging branch on the main repository. This enables the staging branch to run our e2e tests. Once those pass, we'll merge to `main`

@susodapop Thank you! Really happy to contribute and can't wait to try out Liquid Clustering in our system!

rcypher-databricks (approved):

ammarchalifah:
All tests passed. Should we merge the PR? @susodapop

PR #400, opened 2023-08-01
Patching external location support in python models

benc-db:
### Description

Fixes issues where incremental + external_location would complain about location overlap on incremental runs.  It does this by writing the temp tables to their own folder.  Note: since Databricks does not delete files in external locations when relations are dropped, owners will need to manage these folders themselves.

This support is experimental.

### Checklist

I have tested manually since we are still working on test infrastructure to run integration tests against external locations in github.  My manual tests involved two types of tests:
1.) verify that we can run the same dbt command 3 times.  Why 3?  To make sure that we can create the incremental staging tables repeatedly without complaint about overlapping location.
2.) verify that when using python models, incremental, and external locations, we merge successive runs as expected.

I have not included the tests here as they are both hacky and incompatible with our current test environment, but have committed them locally for inclusion when our test environment is ready for external location testing.

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -145,7 +145,7 @@ class DatabricksAdapter(SparkAdapter):
                 identifier=name,
                 type=self.Relation.get_relation_type(kind),
             )
-            for database, schema, name, kind in results.select(
+            for database, schema, name, kind in results.select(  # type: ignore[attr-defined]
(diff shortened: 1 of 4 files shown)


susodapop (approved):
lgtm

benc-db:
Not sure why mypy is failing for things that I did not touch :/

susodapop:
Yeah strange. The mypy check is passing on `main` at least. I'm investigating.

PR #405, opened 2023-08-02
[PECO-921] (Reimplemented) Fix: allow DESCRIBE TABLE EXTENDED to handle more than 2048 characters

### LINKED ISSUE #325: dbt docs generate (>1.3) does not work with glue catalog with large number of tables in schema
### Describe the bug
A clear and concise description of what the bug is. What command did you run? What happened?

Currently when attempting to run dbt docs generate you get a failure message if you have a large number of tables in any given schema. This is because the package will attempt to run the following code show table extended in <schema_name> like '<table_1>|<table_2>|...'.

The call to glue has the following constraint 'Length Constraints: Minimum length of 0. Maximum length of 2048.' which results in the query failing and ultimately an inability to create dbt docs when using glue on any version above 1.3

### Steps To Reproduce

Use glue as your metastore
Have a large number of tables in a given schema
Attempt to run dbt docs generate

### Expected behavior

dbt docs generate to work regardless of how many tables you have in a given schema.

### System information
```
failed to satisfy constraint: Member must have length less than or equal to 2048 (Service: AWSGlue; Status Code: 400; Error Code: ValidationException; Request ID: <REQUEST_ID>; Proxy: null))

```

databricks-cli==0.17.5
dbt-core==1.4.6
dbt-databricks==1.4.2
dbt-spark[PyHive]==1.4.1
elementary-data==0.7.7
graphviz==0.20.1
pre-commit==3.1.1
PyYAML==6.0
rich==13.3.2
shyaml==0.6.2
sqlfluff-templater-dbt==2.0.6
sqlfluff==2.0.6

3.8.10

susodapop:
Resolves #325

### Description

Follow-on work after we reverted the original fix in #404.

The behaviour is now gated behind an environment variable called `DBT_DESCRIBE_TABLE_2048_CHAR_BYPASS`. When that environment variable is set to `true` then the fix is applied. Else we do not cap the joined table names length to 2048 characters. This 2048 limit only applies to Glue Catalog users and should not be applied to all dbt runs.

### Checklist

### DIFF
--- a/tests/unit/test_adapter.py
+++ b/tests/unit/test_adapter.py
@@ -970,6 +970,47 @@ class TestDatabricksAdapter(unittest.TestCase):
                 get_identifier_list_string(list(table_names)[:5]), "|".join(list(table_names)[:5])
             )
 
+    def test_describe_table_extended_should_not_limit(self):
+        """GIVEN a list of table_names whos total character length exceeds 2048 characters
+        WHEN the environment variable DBT_DESCRIBE_TABLE_2048_CHAR_BYPASS is not set
+        THEN the identifier list is not truncated
+        """
+
+        table_names: set(str) = set([f"customers_{i}" for i in range(200)])
+
+        # By default, don't limit the number of characters
+        self.assertEqual(get_identifier_list_string(table_names), "|".join(table_names))
+
+    def test_describe_table_extended_should_limit(self):
+        """GIVEN a list of table_names whos total character length exceeds 2048 characters
+        WHEN the environment variable DBT_DESCRIBE_TABLE_2048_CHAR_BYPASS is "true"
+        THEN the identifier list is replaced with "*"
+        """
+
+        table_names: set(str) = set([f"customers_{i}" for i in range(200)])
+
+        # If environment variable is set, then limit the number of characters
+        with mock.patch.dict("os.environ", **{"DBT_DESCRIBE_TABLE_2048_CHAR_BYPASS": "true"}):
+
+            # Long list of table names is capped
+            self.assertEqual(get_identifier_list_string(table_names), "*")
+
+    def test_describe_table_extended_may_limit(self):
+        """GIVEN a list of table_names whos total character length does not 2048 characters
+        WHEN the environment variable DBT_DESCRIBE_TABLE_2048_CHAR_BYPASS is "true"
+        THEN the identifier list is not truncated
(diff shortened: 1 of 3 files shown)


tests/unit/test_adapter.py
```diff
@@ -947,6 +948,28 @@ def test_parse_columns_from_information_with_table_type_and_parquet_provider(sel
...
+        WHEN the environment variable DBT_DESCRIBE_TABLE_2048_CHAR_BYPASS is "true"
+        THEN the identifier list is replaced with "*"
+        """
+
+        table_names: set(str) = set([f"customers_{i}" for i in range(200)])
+
+        # By default, don't limit the number of characters
+        self.assertEqual(get_identifier_list_string(table_names), "|".join(table_names))
+
+        # If environment variable is set, then limit the number of characters
```
benc-db, line 962: Separate the different test conditions.  General rule of thumb is one assert per test (unless your setup is too expensive).

susodapop: Fixed 2fb7247f05f22c198120852647e7ee66258213ff

tests/unit/test_adapter.py
```diff
@@ -947,6 +948,69 @@ def test_parse_columns_from_information_with_table_type_and_parquet_provider(sel
...
+        table_names: set(str) = set([f"customers_{i}" for i in range(200)])
+
+        # If environment variable is set, then limit the number of characters
+        with mock.patch.dict("os.environ", **{"DBT_DESCRIBE_TABLE_2048_CHAR_BYPASS": "true"}):
+
+            # Long list of table names is capped
+            self.assertEqual(get_identifier_list_string(table_names), "*")
+
+    def test_describe_table_extended_may_limit(self):
+        """GIVEN a list of table_names whos total character length does not 2048 characters
```
benc-db, line 999: When we tackle improving our tests, we should discuss how we want to encode this information.  I tend to favor encoding in the test name in python, but mainly because I haven't come across a test framework in python that did a great job of it.

benc-db (approved):

colin-rogers-dbt:
@benc-db / @susodapop would it be possible to create a 1.5 backport patch?

benc-db:
> @benc-db / @susodapop would it be possible to create a 1.5 backport patch?

What is blocking you/your customer from using 1.6.x?

PR #437, opened 2023-09-05
Filter databricks-sql-connector logs to warn or higher

### LINKED ISSUE #432: Disable log output of databricks-sql-connector in cli
### Describe the feature

Since version 1.5.5 of this adapter, the log output of `databricks-sql-connector` is printed to the console:

image
image

I discovered this now since I upgraded from 1.4.6 to 1.6.1. I have multiple non-technical people which I like to introduce to dbt but this output now looks cumbersome and hard to understand for non-technical people. It might have its reason, but I see it more as disturbing when I not actually search for a bug ...

Can this be moved into the dbt log file and removed from the console output? I think this is against what dbt stands for: a easy to use and understand piece of software. This log just disturbes this which makes it hard for me to train non-technical people for it.

Maybe add an opt-in or opt-out option for this which can be set via an environment variable?

### Additional context
dbt-core 1.6.1
dbt-databricks 1.6.1

### Who will this benefit?
Everybody using this provider and read the cli logs: It gets easier to understand

No. I don't know how.

benc-db:
Resolves #432

### Description

databricks-sql-connector is chatty, which ends up making dbt logs harder to read.  Setting level to WARN by default, so that issues still make it through, but we don't see extra details about every connection initiated and closed.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -78,7 +78,9 @@ class DbtCoreHandler(logging.Handler):
 dbt_adapter_logger = AdapterLogger("databricks-sql-connector")
 
 pysql_logger = logging.getLogger("databricks.sql")
-pysql_logger_level = os.environ.get("DBT_DATABRICKS_CONNECTOR_LOG_LEVEL", "INFO").upper()
+pysql_logger_level = os.environ.get(
+    "DBT_DATABRICKS_CONNECTOR_LOG_LEVEL", "WARN"
+).upper()
 pysql_logger.setLevel(pysql_logger_level)
 
 pysql_handler = DbtCoreHandler(dbt_logger=dbt_adapter_logger, level=pysql_logger_level)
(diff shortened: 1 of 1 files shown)


rcypher-databricks (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Improved legibility of python stack traces ([#434](https://github.com/databricks/dbt-databricks/pull/434)).
+- Streamlined sql connector output in dbt.log ([#437](https://github.com/databricks/dbt-databricks/pull/437))
```
susodapop, line 6: Can we please please please fix your auto formatter settings so we don't get so many changed lines in your pull requests 🙏 ?

susodapop (approved):

CHANGELOG.md
benc-db: Better?

PR #455, opened 2023-09-21
Expand integration tests so that they can be approved for public

benc-db:
### Description

My understanding of Github Actions and configuration is such that this change should allow me to choose to run the integration tests on PRs from public forks, but not do so by default.

### Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,17 +1,11 @@
 name: Integration Tests
 on:
   push:
-    branches:
-      - "main"
-      - "*.latest"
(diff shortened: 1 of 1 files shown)


susodapop:
I believe this will not work since actions run from public forks will never have access to our repository secrets. Will follow-up with you privately.

benc-db:
@susodapop adding workflow dispatch instead.

benc-db:
Closing after talking to @susodapop about internal security requirements :/.  We will continue to push a new branch here to run the integration tests.

PR #461, opened 2023-09-25
Python: if the cluster is already starting up, just wait for it

benc-db:
### Description

This should fix the issue where multiple threads try to start the same cluster, and we get an exception saying the cluster is Pending.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,13 +3,14 @@
 ### Features
 
 - When installing python libraries onto clusters, you can now specify an index_url (Thanks @casperdamen123) ([367](https://github.com/databricks/dbt-databricks/pull/367))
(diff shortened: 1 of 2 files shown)


susodapop (dismissed):
LGTM. This is just wrapping existing code so we can call it from a different place, yes?

rcypher-databricks (dismissed):

benc-db:
> LGTM. This is just wrapping existing code so we can call it from a different place, yes?

It also changes the behavior if the cluster is already starting.  If the cluster was already starting, we would get those pending exceptions.  Now if the state isn't terminated or terminating, it waits until the state is running.  This is important when running with multiple threads because one will start the cluster, then the next will see that its not terminated, so attempt to submit the job.

PR #481, opened 2023-10-13
Implementing metadata freshness checks

benc-db:
### Description

Addresses the 'metadata freshness changes' part of https://github.com/dbt-labs/dbt-core/discussions/8307
Doing this in a way that can work for hive and unity was interesting.  I'm a little concerned about the accuracy/perf around the hive implementation, but its the best I can do, and I don't have a great way to insert a check for hive metastore earlier in the call chain.  

@dataders any concern that if someone has a view as a source (as is the case in my bigger transform jobs), then this mechanism doesn't work?  I didn't see if there was any logic that only called down to this path if the relation is a table.  Both using information schema in this way and describe history are table-only features.  I mean we could look at information_schema.views, but last altered doesn't give the same sort of information for views as it does for tables.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -20,7 +20,7 @@ from typing import (
 from agate import Row, Table, Text
 
 from dbt.adapters.base import AdapterConfig, PythonJobHelper
-from dbt.adapters.base.impl import catch_as_completed, FreshnessResponse
+from dbt.adapters.base.impl import catch_as_completed
 from dbt.adapters.base.meta import available
 from dbt.adapters.base.relation import BaseRelation, InformationSchema
 from dbt.adapters.capability import CapabilityDict, CapabilitySupport, Support, Capability
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -20,9 +20,10 @@ from typing import (
 from agate import Row, Table, Text
 
 from dbt.adapters.base import AdapterConfig, PythonJobHelper
-from dbt.adapters.base.impl import catch_as_completed
+from dbt.adapters.base.impl import catch_as_completed, FreshnessResponse
 from dbt.adapters.base.meta import available
 from dbt.adapters.base.relation import BaseRelation, InformationSchema
+from dbt.adapters.capability import CapabilityDict, CapabilitySupport, Support, Capability
 from dbt.adapters.spark.impl import (
     SparkAdapter,
     GET_COLUMNS_IN_RELATION_RAW_MACRO_NAME,
@@ -106,6 +107,10 @@ class DatabricksAdapter(SparkAdapter):
     connections: DatabricksConnectionManager
 
     AdapterSpecificConfigs = DatabricksConfig
+
+    _capabilities = CapabilityDict(
+        {Capability.TableLastModifiedMetadata: CapabilitySupport(support=Support.Full)}
+    )
 
     @available.parse(lambda *a, **k: 0)
     def compare_dbr_version(self, major: int, minor: int) -> int:
--- a/dbt/adapters/databricks/relation.py
+++ b/dbt/adapters/databricks/relation.py
@@ -3,7 +3,7 @@ from typing import Any, Dict, Optional, Type
 from dbt.contracts.relation import (
     ComponentName,
 )
-from dbt.adapters.base.relation import BaseRelation, Policy
+from dbt.adapters.base.relation import BaseRelation, Policy, InformationSchema
 from dbt.adapters.spark.impl import KEY_TABLE_OWNER, KEY_TABLE_STATISTICS
 from dbt.dataclass_schema import StrEnum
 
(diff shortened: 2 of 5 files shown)


benc-db:
BTW, there may be additional tests required as the guide calls out two more functional tests, but since those tests had a low of snowflake specific comments in them (and so far are only in the snowflake adapter), I'm waiting until they're part of core before implementing here.  I wanted to get this code up for people to look at while I'm away.

dbt/adapters/databricks/relation.py
```diff
@@ -115,3 +125,13 @@ def matches(
     @classproperty
     def get_relation_type(cls) -> Type[DatabricksRelationType]:
         return DatabricksRelationType
+
+    def information_schema(self, view_name=None) -> InformationSchema:
```
benc-db, line 129: This is mostly copied from core, but I had to override so that it would use an information schema that used "`" for quoting, since InformationSchema inherits from BaseRelation.

tests/functional/adapter/test_source_freshness.py
```diff
@@ -0,0 +1,59 @@
...
+            project.adapter.drop_schema(relation)
+            project.adapter.create_schema(relation)
+
+        yield relation.schema
+
+        with project.adapter.connection_named("__test"):
+            project.adapter.drop_schema(relation)
+
+    def test_get_relation_last_modified(self, project, set_env_vars, custom_schema):
+        project.run_sql(
```
benc-db, line 44: This test is currently hand copied from dbt-snowflake.  Nothing I do in the test is specific to databricks, so hopefully it gets pulled into core.

benc-db: Essentially the test just says, were we able to run the metadata commands and not warn/error.  We definitely need another test that asserts that something useful happens.  Will add after I'm back.

benc-db: Updated the test to ensure the freshness tests passed, as opposed to not having a warning or error.  @dataders might be worth considering this form for core; to me at least its more obvious what the test is doing (as opposed to the probe function).  Also, this file did not match the format I found in the documentation for sources.json, btw.

dataders:
@benc-db you da real MVP for posting this. I'll definitely flag this for the adapters eng team to look at once we're back from Coalesce

dbt/adapters/databricks/impl.py
```diff
@@ -107,6 +108,10 @@ class DatabricksAdapter(SparkAdapter):
 
     AdapterSpecificConfigs = DatabricksConfig
 
+    _capabilities = CapabilityDict(
+        {Capability.TableLastModifiedMetadata: CapabilitySupport(support=Support.Full)}
+    )
+
     @available.parse(lambda *a, **k: 0)
```
susodapop, line 115: Is the first time our dialect has needed to use `_capabilities`? Or had we just inherited the `_capabilities` from dbt-spark? Not a blocker for this PR but I wonder if more capabilities should be reflected explicitly here.

dbt/adapters/databricks/relation.py
```diff
@@ -37,6 +37,16 @@ class DatabricksRelationType(StrEnum):
...
 
 
+@dataclass(frozen=True, eq=False, repr=False)
+class DatabricksInformationSchema(InformationSchema):
+    quote_policy: Policy = field(default_factory=lambda: DatabricksQuotePolicy())
+    include_policy: Policy = field(default_factory=lambda: DatabricksIncludePolicy())
+    quote_character: str = "`"
+
+    def is_hive_metastore(self):
+        return self.database is None or self.database == "hive_metastore"
```
susodapop, line 47: Just confirming: does dbt's `database` == Databricks' `schema`?

dbt/include/databricks/macros/metadata.sql
```diff
@@ -0,0 +1,32 @@
+{% macro databricks__get_relation_last_modified(information_schema, relations) -%}
+
+  {%- call statement('last_modified', fetch_result=True) -%}
+    {% if information_schema.is_hive_metastore %}
+        {%- for relation in relations -%}
+            select '{{ relation.schema }}' as schema,
+                    '{{ relation.identifier }}' as identifier,
+                    max(timestamp) as last_modified,
```
susodapop, line 8: just confirming: the output of a Databricks SQL query that includes a `max()` aggregation is never more than a single row, yes? That's why there is no `GROUP BY` present here?

dbt/include/databricks/macros/metadata.sql
```diff
@@ -0,0 +1,32 @@
...
+        {%- endfor -%}
+    {% else %}
+        select table_schema as schema,
+               table_name as identifier,
+               last_altered as last_modified,
+               {{ current_timestamp() }} as snapshotted_at
+        from {{ information_schema }}.tables
+        where (
+          {%- for relation in relations -%}
+            (upper(table_schema) = upper('{{ relation.schema }}') and
```
susodapop, line 23: Per the [databricks docs](https://docs.databricks.com/en/sql/language-manual/sql-ref-identifiers.html#identifiers), identifiers are case-insensitive. Did something break without these `upper()` calls?

tests/functional/adapter/test_source_freshness.py
```diff
@@ -0,0 +1,52 @@
+import os
+import pytest
+
+from dbt.tests.util import get_artifact, run_dbt
+
+freshness_via_metadata_schema_yml = """version: 2
```
susodapop, line 6: Nit, can we move `version: 2` to the following line?
```suggestion
freshness_via_metadata_schema_yml = """
version: 2
```

susodapop (dismissed):
LGTM with a few comments. Agree we need a test of the true output of the freshness check. But the smoke test is a good start. Nice work

dbt/adapters/databricks/impl.py
benc-db: This is new in 1.7

dbt/adapters/databricks/relation.py
benc-db: database == catalog.

dbt/include/databricks/macros/metadata.sql
benc-db: with the absence of group by, it will be a single row.

PR #483, opened 2023-10-19
Allow more arguments to be added on streaming table create + refresh

kdazzle:
Resolves #

### Description

Allow Streaming Tables to be created and refreshed with Options, Tblproperties, etc.

I'm sure this isn't a complete MR, but it would be great to see this brought into the fold, as the streaming table functionality is incomplete without it.

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/materializations/streaming_table/create_or_refresh_streaming_table_as.sql
+++ b/dbt/include/databricks/macros/materializations/streaming_table/create_or_refresh_streaming_table_as.sql
@@ -4,6 +4,11 @@
 
 {% macro databricks__get_create_streaming_table_as_sql(relation, sql) -%}
   create streaming table {{ relation }}
+  {{ file_format_clause() }}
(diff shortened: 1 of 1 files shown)


benc-db:
@kdazzle thank you for the PR!  However, I will probably not be merging at this time.  The reason I'm not planning on merging is that we actually have a much more substantial change coming for ST planned in the 1.7.x timeframe that will incorporate these features into dbt's new framework that handles changes to the table config appropriately.  I'll keep the PR open for now, as I may consider merging in a couple of weeks depending on how far out I see that work landing in the adapter.

kdazzle:
Thanks for taking a look at this @benc-db! I'm excited to see what's in store for the next release. 

BTW, unless I'm confusing all of the different things I've tried, the current 1.7.0rc1 doesn't work when doing a full refresh on a large set of data (hundreds of TBs), but it seems to work with these changes. For my query, at least.

benc-db:
Working on a full implementation with new 1.7. config management now.

PR #486, opened 2023-10-24
1.7.0 Catalog fetch improvement

benc-db:
### Description

[Catalog fetch improvement](https://github.com/dbt-labs/dbt-core/discussions/8307)

Core idea is that we only get data for relations managed by dbt.  For UC, this uses information schema.  For HMS, this continues to use the existing approach, but filters to selected nodes.  Will be updating this PR to refactor and add a couple of unit tests, but the key tests are basically the existing functional tests around documentation.

### Checklist


benc-db:
@mikealfare please take a look when you have time.

dbt/include/databricks/macros/catalog.sql
```diff
@@ -18,3 +18,86 @@
...
+        columns as (
+            {{ databricks__get_catalog_columns_sql(information_schema) }}
+            {{ databricks__get_catalog_relations_where_clause_sql(relations) }}
+        )
+        {{ databricks__get_catalog_results_sql() }}
+    {%- endset -%}
+
+  {{ return(run_query(query)) }}
+    {{ return(run_query(query)) }}
+
```
rcypher-databricks, line 38: extra call to return?

benc-db: Good catch.  certainly a copy-paste issue that didn't cause tests to fail.

tests/integration/debug/test_debug.py
```diff
@@ -13,23 +13,23 @@ def schema(self):
     def models(self):
         return "models"
 
-    def run_and_test(self, contains_catalog: bool):
+    def run_and_test(self):
```
benc-db, line 16: These tests changed because I now explicitly set the catalog to hive_metastore if it is set to None (after checking the cluster setting).  We were defaulting to spark_catalog, but I couldn't find where this was set, and it was causing me pain.

benc-db: So question: is there any situation where changing this default behavior is breaking?

benc-db:
@susodapop @rcypher-databricks ready for review...I had planned on doing more refactoring, in part to get functions that are only for hive into their own module, but I think I'll wait for the day when I can take advantage of what Jade is working on.

rcypher-databricks (approved):
looks good to me

dbt/adapters/databricks/impl.py
```diff
@@ -425,60 +431,54 @@ def parse_columns_from_information(  # type: ignore[override]
     def get_catalog(
         self, manifest: Manifest, selected_nodes: Optional[Set] = None
```
susodapop, line 432: Noting for my own sake that in the previous implementation of this method, `selected_nodes` was optional but was never enforced.

susodapop (approved):
LGTM

dbt/adapters/databricks/impl.py
mikealfare: I believe `selected_nodes` was added recently for the `get_catalog` performance improvements. We had a miss where we called it with the additional argument before checking if we could and @benc-db had to update this signature as a result. So the kwarg did nothing besides allowing `get_catalog` to be called with an additional argument.

PR #488, opened 2023-10-26
Allow models to execute on different warehouses

rcypher-databricks:
Resolves PECO-1182

### Description

Allows models in a dbt project to run on different SQL warehouses.
The available warehouses are specified as named values in the dbt profile.
The configuration for a model can specify by name which warehouse it should run on.

### Checklist

### DIFF
--- a/tests/functional/adapter/warehouse_per_model/test_warehouse_per_model.py
+++ b/tests/functional/adapter/warehouse_per_model/test_warehouse_per_model.py
@@ -1,0 +1,100 @@
+import pytest
+from dbt.tests import util
+from tests.functional.adapter.warehouse_per_model import fixtures
+
+
+class BaseWarehousePerModel:
+    args_formatter = ""
+
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        return {
+            "source.csv": fixtures.source,
+        }
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        d = dict()
+        d["target4.sql"] = fixtures.target3
+        return {
+            "target.sql": fixtures.target,
+            "target2.sql": fixtures.target2,
+            "target3.sql": fixtures.target3,
+            "schema.yml": fixtures.model_schema,
+            "special": d,
+        }
+
+
+class BaseSpecifyingCompute(BaseWarehousePerModel):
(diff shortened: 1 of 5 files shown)


benc-db:
Out of curiosity, how did you validate?  Do you have tests that are uncommitted, or did you look at the logs of some existing tests?

rcypher-databricks:
I verified by modifying one of the integration tests and stepping through the code as well as adding some temporary print statements. I validated the open session parameters for a model that specifies a different warehouse than the one in the profile.

benc-db:
Stylistically looks good, and to the extent I can follow, the logic looks right.  Assuming we go this way, please add unit tests for the new logic, and at least 1 functional test that uses the compute definitions from the existing profiles.

benc-db:
> Stylistically looks good, and to the extent I can follow, the logic looks right. Assuming we go this way, please add unit tests for the new logic, and at least 1 functional test that uses the compute definitions from the existing profiles.

Per discussion with dbt, we're going to try to finish migrating to their functional tests, so no new integration tests.

rcypher-databricks:
Definitely going to add tests if this is the implementation we decide to use.  That's why its a draft PR at this point

tests/functional/adapter/warehouse_per_model/test_warehouse_per_model.py
```diff
@@ -0,0 +1,100 @@
...
+                "+databricks_compute": "alternate_warehouse",
+            }
+        }
+
+    @pytest.fixture(scope="class")
+    def test_config(self):
+        return {"model_names": ["target3"]}
+
+
+class TestSpecifyingForProjectModelsInFolder(BaseSpecifyingCompute):
```
benc-db, line 70: Do we know if we can specify a compute to use with models of a particular tag?  This came up in a customer call where they would want to tag certain models as heavy_compute for example.

benc-db: After reading the dbt docs on tags, I don't think that would work, which is probably fine.  I think having the named compute approach gets us 95% of the way to what it would be if they could target compute to tags.

benc-db (approved):

PR #492, opened 2023-10-27
Beginning to reorg the functional tests

benc-db:
### Description
To make it easier to determine which functional tests we're missing, I'm adopting the organizational strategy of the adapter folder in core.  From migrating these tests, I found that I had added a duplicate test in the last PR because of the difficulty of tracking which tests we have vs which we're missing.

Couple of other notes
* Added alias tests we were missing
* test_basic.py has been exploded into a lot of really simple files, primarily to keep alignment with core, but also because sometimes we have a situation like with doc generation where we need a large override, and I don't like python files that feel like a grab bag of unrelated things.
* I've started moving Databricks to the front of the test class names.  I prefer to have the stuff we want to ignore (TestDatabricks) in one place, rather than sandwiching them around the part of the test name that has meaning.

### Checklist


benc-db:
For discussion:  Do we need Databricks in the name at all?  In cases where dbt doesn't have a base for us to inherit its necessary, but I think when they do that, it's more an accident due to time crunch than an intentional choice.

rcypher-databricks (approved):

susodapop (approved):
LGTM. I checked these out locally and explored in the VSCode test explorer and they are **much** easier to walk through 👍 

_Should we include "Databricks" in the test cases?_
I think no. If we need to override a test case in dbt-core or dbt-spark we can add "Override" to the test case to make it more clear what's happening and why.

PR #493, opened 2023-10-27
Improve dbt seed times by removing type lookup and casting during INSERT

### LINKED ISSUE #476: dbt seed never completing
### Describe the bug
Running `dbt seed` with a number of files >50k rows does not finish despite all jobs having completed on the cluster and `connect_timeout` value being set in the profile. 

### Steps To Reproduce
run `dbt seed` with a number of large files. 

### Expected behavior
Succeed or fail

### System information
tested on
```
dbt=1.5.7
databricks=1.5.6
```

ubuntu 22:04

py38

nrichards17:
Resolves #476 

### Description

Several users have noticed slow run times for loading dbt seeds with >1k records when using the `dbt-databricks` adapter, with run times becoming prohibitively slow for seeds with >10k records.  This PR speeds up that run time substantially by removing unnecessary type lookup and casting during the `INSERT` statement.

DBT seeds essentially are built in two steps:
1. The table is created with the appropriate column types (explicit or inferred) with `CREATE TABLE AS ...`
2. The values are loaded into that table with `INSERT OVERWRITE INTO table.schema VALUES ...`

For some reason in the second step, there is both a type lookup and subsequent `CAST(x) AS type` for **every single value (rows x columns)** in the seed.  This is effectively redundant and unnecessary, since the type information was already used when defining the column types during table creation.

Removing these steps significantly speeds up the seed run times.  For example, I was able to load a seed with 47k records and 9 columns in about 1 minute with this change, whereas previously the seed hadn't even finished after 10+ minutes of loading.

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/materializations/seeds/helpers.sql
+++ b/dbt/include/databricks/macros/materializations/seeds/helpers.sql
@@ -20,9 +20,7 @@
           insert {% if loop.index0 == 0 -%} overwrite {% else -%} into {% endif -%} {{ this.render() }} values
           {% for row in chunk -%}
               ({%- for col_name in agate_table.column_names -%}
-                  {%- set inferred_type = adapter.convert_type(agate_table, loop.index0) -%}
-                  {%- set type = column_override.get(col_name, inferred_type) -%}
-                    cast({{ get_binding_char() }} as {{type}})
+                  {{ get_binding_char() }}
                   {%- if not loop.last%},{%- endif %}
               {%- endfor -%})
               {%- if not loop.last%},{%- endif %}


benc-db:
@nrichards17 I'm on vacation next week, but I'll run this change against our test suite when I get back, and assuming it passes, will try to sneak it in before I release 1.7.0.

benc-db:
Thanks for the PR, appreciate it.

nrichards17:
happy to help, thank you @benc-db !

benc-db:
Closing in favor of 498 which can run against our infra.  Thanks @nrichards17!

PR #499, opened 2023-11-08
1.7.0 Compat

benc-db:
### Description

This PR includes
a.) change to get_catalog to account for changes that dbt made between 1.7.0rc1 and 1.7.0
b.) a fix for a PR I merged earlier that removed casts from seeding.  Turns out that with hive and parquet, the cast is necessary for numeric types; the tests passed initially because the old version of the copy_into tests didn't use seeds, and nowhere else in our project do we test parquet and seeds.
c.) a modification to our external location integration variable because something has changed (I think in the dbr) so that its extra sensitive to running in the top level of a container

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -428,12 +428,15 @@ class DatabricksAdapter(SparkAdapter):
             columns.append(column)
         return columns
 
-    def get_catalog(
-        self, manifest: Manifest, selected_nodes: Optional[Set] = None
+    def get_catalog(self, manifest: Manifest) -> Tuple[Table, List[Exception]]:
+        catalog_relations = self._get_catalog_relations(manifest)
+        return self.get_catalog_by_relations(manifest, catalog_relations)
+
+    def get_catalog_by_relations(
+        self, manifest: Manifest, relations: Set[BaseRelation] = None
     ) -> Tuple[Table, List[Exception]]:
         with executor(self.config) as tpe:
-            catalog_relations = self._get_catalog_relations(manifest, selected_nodes)
-            relations_by_catalog = self._get_catalog_relations_by_info_schema(catalog_relations)
+            relations_by_catalog = self._get_catalog_relations_by_info_schema(relations)
 
             futures: List[Future[Table]] = []
 
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -234,8 +234,7 @@ class DatabricksAdapter(SparkAdapter):
         kwargs = {"relation": relation}
 
         new_rows: List[Tuple[Optional[str], str, str, str]]
-        if relation.database is not None:
-            assert relation.schema is not None
+        if all([relation.database, relation.schema]):
             tables = self.connections.list_tables(
                 database=relation.database, schema=relation.schema
             )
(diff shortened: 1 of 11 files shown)


rcypher-databricks (dismissed):

susodapop (approved):
LGTM

dbt/adapters/databricks/impl.py
```diff
@@ -429,10 +429,20 @@ def parse_columns_from_information(  # type: ignore[override]
         return columns
 
     def get_catalog(
-        self, manifest: Manifest, selected_nodes: Optional[Set] = None
+        self, manifest: Manifest, selected_nodes: Optional[Set[Any]] = None
```
mikealfare, line 432: I believe we added `selected_nodes` as an argument and then removed it again, both in `1.7.0rc1`. However we forgot to remove it from `dbt-spark`. I'm verifying this now.

dbt/adapters/databricks/impl.py
```diff
@@ -429,10 +429,20 @@ def parse_columns_from_information(  # type: ignore[override]
...
+        if selected_nodes:
+            relations: Set[BaseRelation] = {
+                self.Relation.create_from(self.config, n) for n in selected_nodes
+            }
+        else:
+            relations = set(self._get_catalog_relations(manifest))
+        return self.get_catalog_by_relations(manifest, relations)
+
+    def get_catalog_by_relations(
+        self, manifest: Manifest, catalog_relations: Set[BaseRelation]
```
mikealfare, line 443: Minor nit, we call this `relations` in `dbt-core`, not `catalog_relations`. That technically changes the signature.

dbt/adapters/databricks/impl.py
mikealfare, line 1: If the integer type in Databricks matches that of `dbt-spark` (`'bigint'`), then this is fine. Otherwise you'll need to override `convert_integer_type`.

mikealfare: Per our discussion, we should try to update `get_relation` to use the cache instead of hitting the database each time.

mikealfare (commented):
I made some comments of a few things to discuss. I'll look through the full project, specifically for the Adapter and Relation classes, to see if I find anything else.

dbt/adapters/databricks/impl.py
benc-db: Will correct this, thanks.  Python is loosy goosy.

dbt/adapters/databricks/impl.py
benc-db: Is there a test for the convert_integer behavior?  Or would my existing seed tests fail if this wasn't working, since they have ints in them?  All my current tests are passing without overriding convert_integer_type.

mikealfare: https://github.com/dbt-labs/dbt-core/blob/bce7d56eb2c1536a516af811b56756083486c99c/tests/adapter/dbt/tests/adapter/simple_seed/test_seed.py#L370-L390

dbt/adapters/databricks/impl.py
mikealfare: We accidentally left this in for `dbt-spark`, so it actually makes sense to leave it here for now. Please disregard.

rcypher-databricks (approved):

PR #501, opened 2023-11-09
Update `get_relation_without_caching` to use show with wildcard search in place of describe extended

mikealfare:
Resolves #

### Description

I'm attempting to replace the `describe extended` scenario with a wildcard search in `show table extended`. The current behavior is to inspect each table one at a time to determine whether it is a streaming table or a traditional table. However, I think this can be replaced:

Instead of:
```sql
describe extended my_schema.my_table_1
describe extended my_schema.my_table_2
describe extended my_schema.my_table_3
...
```

Do this:
```sql
show table extended in my_schema like '*'
```

This PR is meant to be for discovery. I have not tested this via unit test, integration tests, etc. But I have confirmed that the wildcard search returns all tables in the schema. It's also possible it only returns the first X number of objects; I haven't created thousands of tables to test that.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -234,8 +234,7 @@ class DatabricksAdapter(SparkAdapter):
         kwargs = {"relation": relation}
 
         new_rows: List[Tuple[Optional[str], str, str, str]]
-        if relation.database is not None:
-            assert relation.schema is not None
+        if all([relation.database, relation.schema]):
             tables = self.connections.list_tables(
                 database=relation.database, schema=relation.schema
             )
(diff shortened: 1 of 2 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -234,8 +234,7 @@ def get_relations_without_caching(self, relation: DatabricksRelation) -> Table:
         kwargs = {"relation": relation}
 
         new_rows: List[Tuple[Optional[str], str, str, str]]
-        if relation.database is not None:
-            assert relation.schema is not None
```
mikealfare, line 238: `assert` does not do anything in production. `pytest` flips a debug switch when running to make the assert active, but you would never want that in production. So this actually doesn't do anything, which can be misleading as it suggests we're testing when we're not.

dbt/adapters/databricks/impl.py
```diff
@@ -256,47 +255,67 @@ def get_relations_without_caching(self, relation: DatabricksRelation) -> Table:
 
         # if there are any table types to be resolved
         if any(not row[3] for row in new_rows):
-            # Get view names and create a dictionay of view name to materialization
```
mikealfare, line 259: typo

dbt/adapters/databricks/impl.py
```diff
@@ -256,47 +255,67 @@ def get_relations_without_caching(self, relation: DatabricksRelation) -> Table:
...
+            # Get view names and tables and create a dictionary of name to materialization
+            relation_all_tables = self.Relation.create(
```
mikealfare, line 259: Create a relation named in such a way as to pass the wildcard into the `like` clause.

dbt/adapters/databricks/impl.py
```diff
@@ -256,47 +255,67 @@ def get_relations_without_caching(self, relation: DatabricksRelation) -> Table:
...
                 views = self.execute_macro(SHOW_VIEWS_MACRO_NAME, kwargs=kwargs)
+                tables = self.execute_macro(
+                    SHOW_TABLE_EXTENDED_MACRO_NAME, kwargs={"relation": relation_all_tables}
+                )
 
             view_names: Dict[str, bool] = {
                 view["viewName"]: view.get("isMaterialized", False) for view in views
             }
 
+            def parse_type(information: str) -> str:
```
mikealfare, line 274: The information in this format is a `\n`-delimited list of attributes that are of the form `<attribute_name>: <attribute_value>`.

dbt/adapters/databricks/impl.py
```diff
@@ -256,47 +255,67 @@ def get_relations_without_caching(self, relation: DatabricksRelation) -> Table:
...
+                    )
 
             # create a new collection of rows with the correct table types
             new_rows = [
                 (
                     row[0],
                     row[1],
                     row[2],
-                    str(row[3] if row[3] else typeFromNames(row[0], row[1], row[2])),
+                    str(row[3] if row[3] else typeFromNames(row[0], row[2])),
```
mikealfare, line 318: `schema` is no longer needed in `typeFromNames`, hence remove it here.

requirements.txt
```diff
@@ -1,4 +1,4 @@
 databricks-sql-connector>=2.9.3, <3.0.0
 dbt-spark==1.7.1
-databricks-sdk==0.9.0
+databricks-sdk>=0.9.0
```
mikealfare, line 3: This file did not align with `setup.py`.

dbt/adapters/databricks/impl.py
```diff
@@ -256,47 +255,63 @@ def get_relations_without_caching(self, relation: DatabricksRelation) -> Table:
...
+                for table in tables
+            }
+
             # a function to resolve an unknown table type
-            def typeFromNames(
-                database: Optional[str], schema: str, name: str
-            ) -> DatabricksRelationType:
-                if name in view_names:
+            def typeFromNames(database: Optional[str], name: str) -> DatabricksRelationType:
+                if is_hive_metastore(database):
```
benc-db, line 287: this isn't a safe reordering.  We can only default to table if we know its hive metastore, and not a view.

dbt/adapters/databricks/impl.py
benc-db: My understanding is that the point of this is to detect if the assertion is violated when debugging/testing.  It doesn't actually do nothing, it just doesn't execute in the deployed code.

benc-db: basically, it is there to verify what we believe to be an invariant.  If it is actually invariant, then we don't need the if clause below.  If clause is definitely safer though.

dbt/adapters/databricks/impl.py
mikealfare: Ah, ok.

benc-db:
closing in favor of moving into the 1.7.0 PR.  Thanks @mikealfare

PR #511, opened 2023-11-17
Updates to warehouse-per-model logging and tests

rcypher-databricks:
### Description

- Updated the debug logging for warehouse-per-model to include thread id and relation name. 
- Updated the warehouse-per-model functional tests to include a seed and a snapshot that specify an alternate compute.  Changed the test conditions to include checking log output for messages that indicate the alternate compute was used.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -1122,26 +1122,29 @@ def _get_compute_name(node: Optional[ResultNode]) -> Optional[str]:
 
 def _get_http_path(node: Optional[ResultNode], creds: DatabricksCredentials) -> Optional[str]:
     thread_id = (os.getpid(), get_ident())
-    # Get the http path of the compute resource specified in the node's config.
-    # If none is specified return the default path from creds.
-    compute_name = _get_compute_name(node)
-    if not node or not compute_name:
-        if node:
-            logger.debug(
-                f"On thread {thread_id}: {node.relation_name} using default compute resource."
-            )
-        else:
-            logger.debug(f"Thread {thread_id}: using default compute resource.")
 
+    # If there is no node we return the http_path for the default compute.
+    if not node:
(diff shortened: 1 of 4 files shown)


dbt/adapters/databricks/connections.py
```diff
@@ -1120,21 +1121,31 @@ def _get_compute_name(node: Optional[ResultNode]) -> Optional[str]:
 
 
 def _get_http_path(node: Optional[ResultNode], creds: DatabricksCredentials) -> Optional[str]:
+    thread_id = (os.getpid(), get_ident())
     # Get the http path of the compute resource specified in the node's config.
     # If none is specified return the default path from creds.
     compute_name = _get_compute_name(node)
     if not node or not compute_name:
```
benc-db, line 1128: This structure has become a little weird.  I believe we can drop the `if not node`.

benc-db: maybe we can't due to the creds.http_path.  Still, worth thinking about restructuring because its so odd to check if something is not, and then immediately check if it is.

benc-db: Given this weirdness, let's add either a unit test or an end to end test that validates we do the expected thing in the default compute case.

benc-db (changes requested):
Please add a test validating that we use the default case as expected when in an env where we have some models with specified compute and some without.

dbt/adapters/databricks/connections.py
rcypher-databricks: I agree that `if not node or not compute_name:` is an odd structure.  However I originally did it that way because using just `if not compute_name:` produced linting errors.  I'll re-organize things.

rcypher-databricks: I think the existing unit tests cover all the possibilities for what can be passed in to _get_http_path

benc-db (approved):

PR #513, opened 2023-11-21
Persist docs on incremental only if changed

### LINKED ISSUE #514: Persist Docs fails when a column is defined that does not exist
### Describe the bug
When a column is defined in the `yml` file for a model but does not exist in the model itself, and persist docs is set to `true`, the model will error. Note this does not fail for other adapters (postgres, redshift, snowflake, and biquery all pass), and I am not using contracts to enforce structure.

### Steps To Reproduce
1. Create a new dbt project
2. Add the following to the `dbt_project.yml`:
  ```yaml
  models:
    +persist_docs:
      relation: true
      columns: true
  ```
4. Create a new model file (`models/test_table.sql`) with the following content:
  ```sql
  {{
    config(
      materialized = 'table',
      )
  }}
  
  select 1 as test_col
  ```
6. Create a new `yaml` file `models/tables.yml` with the following content:
  ```yaml
  version: 2
  models:
    - name: test_table
      description: 'A table to test persist docs on'
      columns:
        - name: test_col
          description: 'A test column that does exist'
        - name: test_col2
          description: 'A test column that does NOT exist'
  ```
7. Execute `dbt run`
8. See error

### Expected behavior
The model would succeed and the non-existent column would just not attempt to be commented. 

```log
15:03:15  Running with dbt=1.6.9
15:03:17  Registered adapter: databricks=1.6.7
15:03:17  Unable to do partial parsing because saved manifest not found. Starting full parse.
15:03:18  Found 1 model, 0 sources, 0 exposures, 0 metrics, 476 macros, 0 groups, 0 semantic models
15:03:18  
15:03:20  Concurrency: 1 threads (target='databricks')
15:03:20  
15:03:20  1 of 1 START sql table model dbt_ryan.test_table ............................... [RUN]
15:03:24  1 of 1 ERROR creating sql table model dbt_ryan.test_table ...................... [ERROR in 4.20s]
15:03:25  
15:03:25  Finished running 1 table model in 0 hours 0 minutes and 6.83 seconds (6.83s).
15:03:25  
15:03:25  Completed with 1 error and 0 warnings:
15:03:25  
15:03:25    Runtime Error in model test_table (models/test_table.sql)
  Missing field test_col2 in table hive_metastore.dbt_ryan.test_table with schema:
  root
   |-- test_col: integer (nullable = true)
  .; line 4 pos 8
15:03:25  
15:03:25  Done. PASS=0 WARN=0 ERROR=1 SKIP=0 TOTAL=1
```

### System information
```
Core:
  - installed: 1.6.9
  - latest:    1.7.2 - Update available!

Plugins:
  - databricks: 1.6.7 - Update available!
  - bigquery:   1.6.9 - Update available!
  - snowflake:  1.6.5 - Update available!
  - redshift:   1.6.4 - Update available!
  - postgres:   1.6.9 - Update available!
  - spark:      1.6.1 - Update available!
```

**The operating system you're using:** MacOSX

**The output of `python --version`:** 3.19.3

### Additional context
We encounter this issue in our packages as we may have columns that are optionally included in models based on variables, we would like to be able to document these as they may be complex, but this issue means they would have to be undocumented entirely.

Linked issue: https://github.com/snowplow/dbt-snowplow-web/issues/234

This may actually be an issue with [dbt-spark](https://github.com/dbt-labs/dbt-spark) and is inherited, but I can replicate it specifically with the databricks adapter. My best guess is that this macro, although I haven't forked it to try a fix:
https://github.com/dbt-labs/dbt-spark/blob/5bbf67379c6ee0f8e13819ff736ef9fc90c3f27c/dbt/include/spark/macros/adapters.sql#L232-L238

Should actually contain a check before trying to add the command, such as in the snowflake adatper:
https://github.com/dbt-labs/dbt-snowflake/blob/f67098d585b90303e5a946e363ba860ce9efe78c/dbt/include/snowflake/macros/adapters.sql#L1-L16

Or alternatively _this_ macro, I am not sure which one is actually used - maybe both need updating.
https://github.com/dbt-labs/dbt-spark/blob/5bbf67379c6ee0f8e13819ff736ef9fc90c3f27c/dbt/include/spark/macros/adapters.sql#L362-L381

again compared to snowflake
https://github.com/dbt-labs/dbt-snowflake/blob/f67098d585b90303e5a946e363ba860ce9efe78c/dbt/include/snowflake/macros/adapters.sql#L183-L194

benc-db:
Doesn't fully resolve persist_docs perf issues, but should help for the case of incremental tables.  Resolves #514 for incremental and regular tables.

### Description

For incremental tables, only run alter column if the description has changed.  This incurs an additional describe extended (because we don't have a cache for this information), but since alter column purportedly takes 6 seconds per command, this should be a net benefit to all but the thinnest of incremental tables.

### Checklist


andrefurlan-db (dismissed):

benc-db:
Expanded to be general.  Since this macro will always be called after we have created the table, we can always compare with the existing table's columns to see what we need to alter.  This allows us to avoid the call on columns that are missing descriptions on a fresh table create, and only alter columns that have changed on later calls to incremental.

andrefurlan-db (approved):

PR #517, opened 2023-11-27
Long sessions

rcypher-databricks:
Resolves #

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -24,6 +24,7 @@ from typing import (
     Tuple,
     cast,
     Union,
+    Hashable,
 )
 
 from agate import Table
@@ -35,6 +36,7 @@ from dbt.adapters.spark.connections import SparkConnectionManager
 from dbt.clients import agate_helper
 from dbt.contracts.connection import (
     AdapterResponse,
+    AdapterRequiredConfig,
     Connection,
     ConnectionState,
     DEFAULT_QUERY_COMMENT,
@@ -44,7 +46,10 @@ from dbt.contracts.connection import (
 from dbt.events.types import (
     NewConnection,
     ConnectionReused,
+    ConnectionLeftOpenInCleanup,
+    ConnectionClosedInCleanup,
 )
+
 from dbt.contracts.graph.manifest import Manifest
 from dbt.contracts.graph.nodes import ResultNode
 from dbt.events import AdapterLogger
@@ -718,13 +723,22 @@ class DatabricksAdapterResponse(AdapterResponse):
 
 @dataclass(init=False)
 class DatabricksDBTConnection(Connection):
-    last_used_time: Optional[str] = None
+    last_used_time: Optional[float] = None
     acquire_release_count: int = 0
+    compute_name: str = ""
+    http_path: str = ""
 
 
 class DatabricksConnectionManager(SparkConnectionManager):
     TYPE: str = "databricks"
     credentials_provider: CredentialsProvider = None
+
+    def __init__(self, profile: AdapterRequiredConfig) -> None:
+        super().__init__(profile)
+        if USE_LONG_SESSIONS:
+            self.threads_compute_connections: Dict[
+                Hashable, Dict[Hashable, DatabricksDBTConnection]
+            ] = {}
 
     def compare_dbr_version(self, major: int, minor: int) -> int:
         version = (major, minor)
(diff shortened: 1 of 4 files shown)


dbt/adapters/databricks/connections.py
```diff
@@ -105,6 +110,8 @@ def emit(self, record: logging.LogRecord) -> None:
 CLIENT_ID = "dbt-databricks"
 SCOPES = ["all-apis", "offline_access"]
 
+USE_LONG_SESSIONS = os.getenv("DBT_DATABRICKS_LONG_SESSIONS", "FALSE").upper() == "TRUE"
```
benc-db, line 113: Any reason to default to false?  I think if we're concerned about breaking behavior, we should release as a beta with defaulted to true.  If this change improves performance, as we expect, it would be annoying to make users opt in, unless we know if breaks some use cases.

dbt/adapters/databricks/connections.py
```diff
@@ -797,6 +865,168 @@ def set_connection_name(
 
         return conn
 
+    # override
+    def release(self) -> None:
+        if not USE_LONG_SESSIONS:
+            return super().release()
+
+        with self.lock:
```
benc-db, line 873: why lock here but no lock on acquire?

dbt/adapters/databricks/connections.py
```diff
@@ -797,6 +865,168 @@ def set_connection_name(
...
+        self,
+        conn: DatabricksDBTConnection,
+        new_name: str,
+        node: Optional[ResultNode] = None,
+    ) -> DatabricksDBTConnection:
+        """Update a connection that is being re-used with a new name, handle, etc."""
+        _long_sessions_only("_update_compute_connection")
+
+        compute_name = _get_compute_name(node=node) or ""
+        if conn.name == new_name and conn.state == "open" and conn.compute_name == compute_name:
```
benc-db, line 935: apologies if this is enforced somewhere I'm not seeing in the PR, but are we enforcing this to lower?  We've hit issues in the past where we compare to a particular case and the DBR decides to change what case they send.

dbt/adapters/databricks/connections.py
```diff
@@ -1147,8 +1384,17 @@ def _get_http_path(node: Optional[ResultNode], creds: DatabricksCredentials) ->
...
-    )
+    if not USE_LONG_SESSIONS:
+        logger.debug(
+            f"On thread {thread_id}: {node.relation_name} using compute resource '{compute_name}'."
+        )
 
     return http_path
+
+
+def _long_sessions_only(name: str) -> None:
```
benc-db, line 1395: I feel like it's more pythonic to do something like:

```
assert USE_LONG_SESSIONS, "This path should only be reachable with USE_LONG_SESSIONS"
```

My reasoning is that, I believe your intention is 2-fold:

a.) communicate to readers that these code paths should not be reached if USE_LONG_SESSIONS is False
b.) during development, fail execution if you've violated this pre-condition that you intended to enforce.

By the time it gets to the user, we should not have any execution paths that trigger this exception.  If you are worried that we still might, then I believe the best response is to write more tests to validate that we can't.

benc-db: Long story short, favor asserts for conditions that you intend to enforce, exceptions for things outside of your control.

benc-db (changes requested):
Minor comments, overall looks good.

dbt/adapters/databricks/connections.py
rcypher-databricks: I copied this from BaseConnectionManager, but you make a good point.  I will make it more robust.

dbt/adapters/databricks/connections.py
```diff
@@ -728,30 +735,37 @@ class DatabricksDBTConnection(Connection):
...
     def _release(self) -> None:
         """Indicate that this connection is not in use."""
         logger.debug(f"DatabricksDBTConnection._release: {self._get_conn_info_str()}")
-        self.last_used_time = time.time()
         # Need to check for > 0 because in some situations the dbt code will make an extra
         # release call on a connection.
         if self.acquire_release_count > 0:
             self.acquire_release_count -= 1
 
+        if self.acquire_release_count == 0:
```
benc-db, line 754: If, as stated above, sometimes dbt makes extra release calls, is it safe to have this based on release_count == 0?  Is this counter ever above 1?  If so, doesn't that mean we could hit 0 when we should actually be at 1?

dbt/adapters/databricks/connections.py
```diff
@@ -1255,6 +1282,81 @@ def exponential_backoff(attempt: int) -> int:
             retry_timeout=(timeout if timeout is not None else exponential_backoff),
         )
 
+    @classmethod
+    def _open2(cls, connection: Connection) -> Connection:
```
benc-db, line 1286: We could put these into separate namespaces (if we continue functional) or separate class (if we want to be OOP) so that they could both be called open, and then pass the appropriate function in when we create the connection object.

benc-db (commented):
Please review comments.  Overall looks good, but would like clarification on the counter, and consider refactor the open functionality to be injectable.

benc-db (approved):

PR #524, opened 2023-12-04
Refactor macro tests so that we can move macros

benc-db:
### Description

This PR was originally going to be reorganizing the macros to be more aligned with dbt-core.  When I tried to move a macro, I realized how insane our macro tests were, as they all broke.

In this PR:

* Move create_table_as to be aligned with dbt-core organization
* Refactor macro tests to use pytest fixtures
* Refactor the base of the macro tests so that an inheriting test can specify which templates it needs to load into the context in order for the render to work properly.

Now, in a world of fine-grained unit tests, this would not be as gnarly, because part of the issue here is needing to load in the dependencies of the macro.  For example, if your macro references file_format_clause(), you'll need the template context to know how to resolve that.  In this case, it comes from dbt-spark, in the adapters.sql file.  An alternative approach is to mock out everything, but it may be quite hard to judge if the generated sql is actually something that will work.  So instead, I give pytest fixtures that can be overridden in order to give whatever template resolution you need to get the macros to work.  In my next commit, I'll add documentation to base.

### Checklist


tests/unit/macros/base.py
```diff
@@ -1,56 +1,213 @@
...
+        Anything you put in this dict will be returned by config in the rendered template
+        """
+        local_config: Dict[str, Any] = {}
+        context["config"].get = lambda key, default=None, **kwargs: local_config.get(key, default)
+        return local_config
+
+    @pytest.fixture(autouse=True)
+    def var(self, context) -> dict:
+        """
+        Anything you put in this dict will be returned by config in the rendered template
```
rcypher-databricks, line 29: Same comment as in config() above

rcypher-databricks (dismissed):
Lot's of really good work here. Thanks for doing this.

tests/unit/macros/base.py
benc-db: Good catch

PR #527, opened 2023-12-07
Fix minor test bug for WPM

benc-db:
### Description

Fixes minor test log bugs for WPM and long sessions.

### Checklist

### DIFF
--- a/tests/functional/adapter/warehouse_per_model/test_warehouse_per_model.py
+++ b/tests/functional/adapter/warehouse_per_model/test_warehouse_per_model.py
@@ -1,3 +1,5 @@
+import os
+import mock
 import pytest
 from dbt.tests import util
 from tests.functional.adapter.warehouse_per_model import fixtures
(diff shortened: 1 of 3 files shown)


rcypher-databricks (dismissed):

tests/functional/adapter/warehouse_per_model/test_warehouse_per_model.py
```diff
@@ -24,35 +26,45 @@ def models(self):
...
+            res = util.run_dbt(
+                ["run", "--select", model_name, "--profiles-dir", profile_dir],
+                expect_pass=False,
+            )
             msg = res.results[0].message
             assert "Compute resource alternate_warehouse does not exist" in msg
             assert model_name in msg
 
 
+@pytest.mark.skip("May fail non-deterministically due to issue in dbt test framework.")
```
rcypher-databricks, line 53: Shouldn't we be disable TestWarehousePerModel and leaving these ones to run.  I though we should be disabling the tests that use the profiles_config_update fixture.

benc-db: I feel like it's more valuable to validate the affirmative case, and even with the bug we identified, these additions to the profile shouldn't impact any other tests (since their schema's won't reference the compute block)

rcypher-databricks (approved):

PR #529, opened 2023-12-08
Fixing list_schemas so that we stop trying to create schemas that exist

### LINKED ISSUE #464: Check if a schema exists before running "create schema if not exists"
### Describe the bug
When adding a "catalog" parameter to a dbt profile, the command create schema if not exists `catalog`.`schema` will always be executed even if the schema already exists. If the user does not have the right to create on catalog, he will have an error. When not specifying the catalog, the "show databases" and then "show schemas" is executed which avoid running  the "create schema if not exists". 

### Steps To Reproduce
1- Create manually a Schema (e.g. schema_test) in a catalog (e.g. catalog_test)
2- Remove right to Create on Catalog for a user
3 - Create a dbt profile with 
      schema: schema_test
      catalog: catalog_test
4- dbt run
5- Check the error

### Expected behavior
If a catalog parameter is in the profile, check if the schemas exists before running the "create schema if not exists".

benc-db:
Resolves #464

### Description

The inclusion of the backticks was causing list_schemas to return only 'global_temp'.  This meant that the precaution in dbt-core to not try to create schemas that already exist wasn't seeing the correct list of schemas.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -1179,12 +1179,19 @@ class DatabricksConnectionManager(SparkConnectionManager):
                     cursor.close()
 
     def list_schemas(self, database: str, schema: Optional[str] = None) -> Table:
+        database = database.strip("`")
+        if schema:
+            schema = schema.strip("`")
         return self._execute_cursor(
             f"GetSchemas(database={database}, schema={schema})",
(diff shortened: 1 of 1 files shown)


rcypher-databricks (approved):
Really nice catch here.

PR #536, opened 2023-12-18
Materialized View Config Level Up

benc-db:
### Description

Completes support of on_config_changes for MVs, expanding the config that we support.  Trying to decide how to clean up tblproperties, because it turns out they're not alterable for MVs, but are for other materializations.  Also, might try to move to_sql stuff out to macros if I come up with a comparably clean way to call and test there, as Mike convinced me that we should delegate to jinja as our presentation layer.

### Checklist


benc-db:
@mikealfare

benc-db:
Redone so as to minimize SQL rendering outside of Jinja, add Jinja tests, streamlined tblproperties (will add the complexity when its actually needed, since MVs can't alter them).

rcypher-databricks (approved):

PR #547, opened 2024-01-10
Python model specific session handling to prevent using invalid sessions

rcypher-databricks:
Resolves:
[ES-985588: dbt-databricks data models timeout](https://databricks.atlassian.net/browse/ES-985588)

### Description
A long running python model can leave a connection/session idle for long enough that it times out and is closed.
Subsequent use of the connection results in an 'Invalid Sessionhandle' error.
#### Cause
For non-python data models the model is executed using sql statements.  So even if the sql statement takes a long time to run, the connection is in active use and doesn't timeout on the server.  For a python model a connection is acquired, but the model is run using rpc calls. Thus, if the python model takes a long time to run, the session can timeout on the back end.
The code to cleanup idle connections was not catching this because when a connection is released, after running a model, the last used time is set.  So even though running the python model didn't use the connection as far as the cleanup code was concerned it had been in active use.
There are two cases where the invalid connection would be used.
- Depending on the python model additional sql needs to be run before the model is considered complete. So there would be an attempt to use the connection within the bounds of the connection being acquired/released to run the python model.
- After the python model is finished the connection is re-used by another model. 
#### Fix
Updated the DatabricksDBTConnection class to track the language of the model being executed.  If the language is python we do not update the last used time when the connection is released.  This is to reflect the fact that the python model isn't necessarily using the connection.
This change allows the idle cleanup code to correctly cleanup the idle connection.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Fix for issue where long-running python models led to invalid session errors ([544](https://github.com/databricks/dbt-databricks/pull/544))
+- Added python model specific connection handling to prevent using invalid sessions ([547](https://github.com/databricks/dbt-databricks/pull/547)) 
 
 ## dbt-databricks 1.7.3 (Dec 12, 2023)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ### Fixes
 
-- Fix for issue where long-running python models led to invalid session errors ([544](https://github.com/databricks/dbt-databricks/pull/544))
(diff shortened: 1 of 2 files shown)


benc-db (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Fix for issue where long-running python models led to invalid session errors ([544](https://github.com/databricks/dbt-databricks/pull/544))
```
benc-db, line 5: You can delete my previous line, since your commit is a much more complete solution.

benc-db (approved):

PR #552, opened 2024-01-16
Fixed issue with streaming tables always being dropped

rcypher-databricks:
Streaming tables were always being dropped and recreated instead of being refreshed. This was due to streaming tables being incorrectly identified as regular tables. 
The _parse_type() function was returning a type string of "Type:STREAMING_TABLE".  The calling code was expecting just the string "STREAMING_TABLE" and so wasn't correctly identifying streaming tables.

Updated the _parse_type() function to return the expected type string, ie stripped away the "Type:" part. 

After fixing the table type parsing testing showed that this exposed an issue with dropping streaming tables.
Implemented macro databricks__drop_relation so that we use the correct sql to drop a streaming table 
Signed-off-by: Raymond Cypher <raymond.cypher@databricks.com>


benc-db (approved):

benc-db (approved):

benc-db:
@rcypher-databricks merge if you're ready.

PR #555, opened 2024-01-17
[PECO-1387] Fix: mv/st network request authentication is being overridden by local .netrc file

susodapop:
### Description

This pull request fixes an issue in the materialized view / streaming table code path where a local `.netrc` file in a user's home directory could override the authentication headers used for polling the MV/ST status. The fix is the same as the one in #338 and uses the same `BearerAuth` class. 

The fix is to use a `requests.Session` with its `.auth` property set to an instance of our custom `BearerAuth` class and then replace every call to `requests.get(...)` with a call to `self.session.get(...)`. For simplicity, I also set the `User-Agent` header at the `Session` level so that we don't need to pass a `headers` dict around between methods.

To test this, I ran the `TestMaterializedView` integration test with a bad `.netrc` file in my home directory. _Before_ this change, the test failed with this error:

```
Error getting info for materialized view/streaming table
```
This indicates an authentication error. 

_After_ the change in this pull request, the materialized view / streaming table test still fails (because the test is broken - which I confirmed with @rcypher-databricks ) but it doesn't fail because of an authentication issue. This tells me that the fix is working as it should.

As for a direct test of this behaviour, we don't have a clean way to add a bad `.netrc` to our cloud testing environment so a manual test like I describe above is the only choice :/

### Checklist


dbt/adapters/databricks/connections.py
```diff
@@ -578,10 +580,13 @@ def pollRefreshPipeline(
         stopped_states = ("COMPLETED", "FAILED", "CANCELED")
         host: str = self._creds.host or ""
         headers = self._cursor.connection.thrift_backend._auth_provider._header_factory()
-        headers["User-Agent"] = self._user_agent
 
-        pipeline_id = _get_table_view_pipeline_id(host, headers, model_name)
-        pipeline = _get_pipeline_state(host, headers, pipeline_id)
+        session = Session()
+        session.auth = BearerAuth(headers)
+        session.headers = {"User-Agent": self._user_agent}
```
susodapop, line 586: I set the headers on the `Session` now so we don't need to pass `headers` to `_get_pipeline_state` and `_get_update_error_message` etc.

dbt/adapters/databricks/connections.py
```diff
@@ -606,7 +611,7 @@ def pollRefreshPipeline(
             # should we do exponential backoff?
             time.sleep(polling_interval)
 
-            pipeline = _get_pipeline_state(host, headers, pipeline_id)
+            pipeline = _get_pipeline_state(session, host, pipeline_id)
```
susodapop, line 614: Here you can see I replace `headers` with `session`.

benc-db (approved):

PR #566, opened 2024-01-24
Add SQL Column Mismatch Validation Into Incremental Append Strategy

case-k-git:
Resolves #
https://github.com/databricks/dbt-databricks/issues/567

### Description
I have added validation to the append process of the incremental_strategy. This change will help when we add new columns into incremental append table.

If a new column is added to an existing dbt SQL and forgot to update dbx existing table columns, The query runs and succeeds, but the newly added columns in SQL are not added when using the incremental append strategy. Data will only be added to the old columns that exist in the existing table.

Since no error occurs, new column data will not be added despite the addition of columns to the SQL model, and subsequent dbt with dependencies will fail due to the lack of columns.

In this pull request , I have added validation to the append process of the incremental_strategy. If there is a shortage of columns, the process will now fail as follows.

To ensure idempotence in the append process, incremental append strategy created a temporary table with the new columns added, then narrowed down to the existing columns before adding data. It will be like this.The issue of this specification is data will be append without new columns and any error or warning message.

```
insert into table `catalog_demo`.`test17060086231712186480_Incremental_strategies`.`append_delta`
      select `id`, `msg` from `append_delta__dbt_tmp`
```
With this new revision, if there is a shortage of SQL columns, it will now be detectable along with the following error.In this example I add `msg2` column as a new one. We will be able to know new column is missing into dbx table because of the error. After checked this validation results, user can update exiting table column by executing ALTER TABLE manually.

```
insert into table `catalog_demo`.`test17060620400077328677_Incremental_strategies`.`append_delta`
      select `id`, `msg`, `msg2` from `append_delta__dbt_tmp`
```
```
  Table schema:
  root
  -- id: long (nullable = true)
  -- msg: string (nullable = true)
 
 
  Data schema:
  root
  -- id: long (nullable = true)
  -- msg: string (nullable = true)
  -- msg2: string (nullable = true)
 
           
  Table ACLs are enabled in this cluster, so automatic schema migration is not allowed. Please use the ALTER TABLE command for changing the schema.In this example I add msg2 column as a new one.
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Fixed Hive performance regression by streamlining materialization type acquisition ([557](https://github.com/databricks/dbt-databricks/pull/557))
 - Fix: Python models authentication could be overridden by a `.netrc` file in the user's home directory ([338](https://github.com/databricks/dbt-databricks/pull/338))
 - Fix: MV/ST REST api authentication could be overriden by a `.netrc` file in the user's home directory ([555](https://github.com/databricks/dbt-databricks/pull/555))
+- Add SQL Column Mismatch Validation Into Incremental Append Strategy([566](https://github.com/databricks/dbt-databricks/pull/566))
 
 ### Under the Hood
 
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -50,7 +50,7 @@
 
 {% macro get_insert_into_sql(source_relation, target_relation) %}
 
-    {%- set dest_columns = adapter.get_columns_in_relation(target_relation) -%}
+    {%- set dest_columns = adapter.get_columns_in_relation(source_relation) -%}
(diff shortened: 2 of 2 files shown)


benc-db:
Looking at the incremental tests, I'm wondering if the issue you're hitting is that you need to use a different setting for incremental schema changes.  Have you tried on_schema_change='fail'?

case-k-git:
Ah I see I did not know that. Thank you for your review let me try tomorrow.
https://docs.getdbt.com/docs/build/incremental-models

case-k-git:
@case-k-git  I confirmed thank you . let me close this PR

case-k-git:
By the way do you if there are some option which can override default value from dbt project? if you know such a option appreciate if you let me know.

I created feature request to dbt core 
https://github.com/dbt-labs/dbt-core/issues/9449

benc-db:
Does it not work to specify it like other model config in the project file?  https://docs.getdbt.com/reference/model-configs

case-k-git:
Thank you for reply. I see it seems to be we can pass default option to apply each models. let me try this. Thank you!

case-k-git:
It worked Thank you for reply and solution !!

PR #580, opened 2024-02-07
Fixed the behavior of the incremental schema change ignore option to properly handle the scenario when columns are dropped

case-k-git:
Resolves #
https://github.com/databricks/dbt-databricks/issues/581

### Description
When processing incrementally, adding new columns is ignored by the ignore setting. However, when a SQL model is modified to remove columns, it fails despite the ignore setting. This is because it attempts to query a column that does not exist in the created temp table. According to the dbt documentation, the job should be designed not to fail when ignored, so it has been corrected.

For example, in this use case, even if we remove `column_2` from the SQL model, the query still attempts to include `column_2` because it exists in the current table schema. However, since `column_2 ` does not exist in the temporary table, the query fails.

The intended SQL insert statement looks like this:

```
insert into table `catalog_demo`.`test17060620400077328677_Incremental_strategies`.`append_delta`
      select  column_1, column_2 from `append_delta__dbt_tmp`

```

**Dbt documentation**
>Similarly, if you remove a column from your incremental model, and execute a dbt run, this column will not be removed from your target table.

So this should not happen
https://docs.getdbt.com/docs/build/incremental-models#default-behavior

### Checklist


benc-db:
Thanks for submitting this.  Will review and run tests today.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -51,8 +51,17 @@
...
 
     {%- set dest_columns = adapter.get_columns_in_relation(target_relation) -%}
-    {%- set dest_cols_csv = dest_columns | map(attribute='quoted') | join(', ') -%}
-    insert into table {{ target_relation }}
+    {%- set source_columns = adapter.get_columns_in_relation(source_relation) -%}
+    {%- set common_columns = [] -%}
+    {%- for dest_col in dest_columns -%}
+        {%- for source_col in source_columns -%}
+            {%- if dest_col.name == source_col.name -%}
+                {%- if common_columns.append(dest_col) -%}{%- endif -%}
```
benc-db, line 59: This seems like it will not work in the case where columns are added in the new code.  Also, I'm not sure this is any more efficient, but stylistically it might be better to do a single loop and check for presence in the second list:

```
{%- set source_columns = adapter.get_columns_in_relation(source_relation) | map(attribute="name") -%}
{%- for dest_col in dest_columns -%}
   {% if dest_col.name in source_columns %}
...
```

case-k-git: Thank you for your review year I will refactor it!

>This seems like it will not work in the case where columns are added in the new code

I think default case is ignore and existing code is also using destination columns. So I thought Adding new columns might not be problem. How do you think?

benc-db: I may not be familiar enough with the surrounding code, let me look a little deeper.

benc-db:
Running the integration tests now...the interaction here is complicated enough that I want to see if our tests yell, because there's a lot of back and forth between dbt-core and dbt-databricks to come up with the actual merge statement :P

dbt/include/databricks/macros/materializations/incremental/strategies.sql
benc-db: Ok, I understand now how this works, I think this is fine.

benc-db: Let me try explaining it to verify:

This statement is saying that in the target relation (which is the existing table), we are going to insert into it all the new incremental data that matches from the tmp source table; `insert into` will provide default values for any columns not named.  The issue you are solving is that previously we attempted to select columns from source that were only guaranteed to be in target.

What I'm confused by is why no one has raised this issue on dbt-spark :)

case-k-git: @benc-db  Thank you for testing and investigation. 
I see. Yea I agree and wander why no one has raised this issue till this time. 

I refactor this as you suggested. Could you please check this?
https://github.com/databricks/dbt-databricks/pull/580/commits/b33764ca2569a9d0f87d04a803486b888dab648b

benc-db:
Rerunning tests now.  Will merge after they pass.  I updated the changelog to reflect some weirdness that happened in deploys this week.

benc-db (approved):

case-k-git:
@benc-db  Thank you !

case-k-git:
I requested same change to dbt-spark

dbt/include/databricks/macros/materializations/incremental/strategies.sql
nimeshpatni: Hi @case-k-git , This single loop iteration is breaking functionality and require an immediate fix.
This will only work when source and destination have columns in same order, as a result we forcefully required to perform `--full-refresh`. For an instance: 
image

DBT's internal Temp view (source columns: [A, B, C, ... N]) and Destination table (target columns arranged in random fashion due to some process: [A, B, C, D, F, E, .. N])  have actually 14 columns in common, but due to this single iteration change, it compares sequentially and returns only 4 & as a result, inserts 10 blank columns at the time of final execution. Looking forward for the immediate resolutions and test cases as well. Thank you.
CC: @benc-db

case-k-git: @nimeshpatni  Thank you! I see let me check!

case-k-git: I confirm the error by changing column order . Thank you
https://github.com/databricks/dbt-databricks/issues/594#issuecomment-1978019472

PR #588, opened 2024-02-16
Work around a type annotation mistake in dbt-core.

### LINKED ISSUE #587: Access of Missing node_relation Attribute Causes dbt-core Hang
### Describe the bug

When a dbt project using the dbt-databricks adapter contains a saved query, and a user exports that saved query, the operation hangs indefinitely.

### Steps To Reproduce

1. Create a dbt project with a saved query.
2. Set the environment variable DBT_INCLUDE_SAVED_QUERY=TRUE
3. Run `dbt build`

You should see that all operations other than the saved query start and complete normally. The saved query operation starts but runs indefinitely, causing dbt-core to hang.

### Expected behavior

The saved query should be exported, and dbt should terminate normally after all build operations are complete. 

### System information

This problem affects (at least) dbt 1.7 and the current main branch.

Repro was done on macos using python 3.10.

### Futher Context

The hang is due to the fact that the dbt-databricks adapter throws an exception. This interacts with a very unfortunate behavior of dbt's (to be fixed in upcoming releases) whereby unhandled exceptions cause the worker thread they occur on to run indefinitely.

I will provide a PR shortly, showing where this exception originates, and providing a patch which avoids the exception.

peterallenwebb:
Resolves #587 

### Description

The type annotation `ResultNode` from dbt-core should guarantee that an object has a `relation_name` property. In practice, the object passed by dbt-core to the functions modified by this PR do not always have that property. In particular, the `node` object might be a `SavedQuery`, which does not have `relation_name`.

This PR adds explicit checks to work around the issue when `relation_name` is missing.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixes
 - Fixed the behavior of the incremental schema change ignore option to properly handle the scenario when columns are dropped (thanks @case-k-git!) ([580](https://github.com/databricks/dbt-databricks/pull/580))
+- Fixed export of saved queries ([588](https://github.com/databricks/dbt-databricks/pull/588))
 
 ## dbt-databricks 1.7.7 (Feb 6, 2024)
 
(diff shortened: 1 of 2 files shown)


benc-db:
Sorry I'm a little late in getting to this, looks reasonable, so will merge after tests pass.

benc-db:
Reopening here for minor changes and to run on our test infra: https://github.com/databricks/dbt-databricks/pull/589

Thanks for the submission!

PR #597, opened 2024-02-27
Shard long passwords on Windows

### LINKED ISSUE #563: OAuth token caching fails on Windows 11 laptop
### Describe the bug
OAuth token caching fails on Window 11 laptop

### Steps To Reproduce

1. Setup dbt profile for U2M OAuth to Databricks
2. Execute ”dbt run” from your VS Code terminal (or any other terminal)
3. In the web popup, authenticate yourself to Databricks.
4. You will be authed
5. You will get error message for ”Cannot save auth token” and dbt run will continue as expected
6. On the next dbt run, step 3 4 and 5 will repeat itself because it cannot retrieve the token from cache. Adding a new tab in the browser each time.

Unfortunately, this makes the development experience so poor that I switched back to hardcoding the token (which I was hoping to avoid because of security reasons).

### Expected behavior
After step (4) above, the OAuth token will be cached for subsequent dbt runs without the need to re-auth within the token lifetime.

```
07:44:32  Found 24 models, 133 tests, 4 seeds, 1 operation, 12 sources, 0 exposures, 0 metrics, 1301 macros, 0 groups, 0 semantic models
07:44:32
07:44:37  Databricks adapter: could not save token
07:44:45  Concurrency: 4 threads (target='dev')
``` 

### System information
```
Core:
  - installed: 1.7.2
  - latest:    1.7.5 - Update available!

Plugins:
  - databricks: 1.7.3 - Up to date!
  - spark:      1.7.1 - Up to date!
```

Windows 11

Python 3.12.0

### Additional context
To debug it, I changed my local code slightly to log the exception as info instead of debug ([line 411](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/connections.py#L411)) and than right before the 'Databricks adapter: could not save token' message I get this one: Databricks adapter: (1783, 'CredWrite', 'The stub received bad data').

I think it is related to this issue in keyring: https://github.com/jaraco/keyring/pull/544 . There is says that windows has a limit of storing 1.280 (utf-16) characters and that storing the entire jwt token can easily go over that limit. In fact, the token I got returned is 2.887 characters long. Here is a similar issue in poetry that is tracked back to keyring with the same reason: https://github.com/python-poetry/poetry/issues/6597 .

I checked and on my laptop keyring 24.3.0 is installed which is the latest.

thijs-nijhuis-shell:
Resolves #563 

### Description

When using the oAuth U2M flow on a windows laptop, the credentials are not stored in the Windows Credential Store because it exceed 1280 characters. Therefore, each call to Databricks results in a new authentication request (and opens a new tab in your browser). In this PR, all interaction with keyring (get, set and delete) is redirect to a custom function that will check the length and platform before storing the credentials. If on Windows and credentials are longer then 1280 characters, it will shard the password in chunk of 1280 and store those along with some meta data. When getting or deleting the credentials, it will check for this metadata and, if so fetch all shards and return the concatenated result or delete them all in case of a delete.

I have tested my code on a windows laptop as well as on a mac. On the mac, I first got the credentials with the current version in 1.7.8 so that it is stored, then replaced connections.py with the new version and ran a 'dbt debug' again to see if the stored credentials was fetched an used. It was.
I have added 2 Unit Tests to test both storing a long and a short password. All Unit Test succeeded.
I tried running the integration test but couldn't get them to run. The tox command seems to not like the bash commands on a windows laptop and running the python command directly causes all integration to fail on not finding the 'test' profile. 

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -445,7 +445,7 @@ class DatabricksCredentials(Credentials):
             keyring.set_password(service_name, username, json.dumps(shard_info))
             # then store all shards with the shard number as postfix
             for i, s in enumerate(password_shards):
-                keyring.set_password(service_name, f"{username}|{i}", s)
+                keyring.set_password(service_name, f"{username}__{i}", s)
 
     def get_sharded_password(self, service_name: str, username: str) -> Optional[str]:
         password = keyring.get_password(service_name, username)
@@ -459,7 +459,7 @@ class DatabricksCredentials(Credentials):
 
                 password = ""
                 for i in range(shard_count):
-                    password += str(keyring.get_password(service_name, f"{username}|{i}"))
+                    password += str(keyring.get_password(service_name, f"{username}__{i}"))
         except ValueError:
             pass
 
@@ -474,7 +474,7 @@ class DatabricksCredentials(Credentials):
             if password_as_dict.get("sharded_password"):
                 shard_count = int(password_as_dict.get("shard_count"))
                 for i in range(shard_count):
-                    keyring.delete_password(service_name, f"{username}|{i}")
+                    keyring.delete_password(service_name, f"{username}__{i}")
         except ValueError:
             pass
 
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -117,6 +117,7 @@ DBT_DATABRICKS_HTTP_SESSION_HEADERS = "DBT_DATABRICKS_HTTP_SESSION_HEADERS"
 REDIRECT_URL = "http://localhost:8020"
 CLIENT_ID = "dbt-databricks"
 SCOPES = ["all-apis", "offline_access"]
+MAX_NT_PASSWORD_SIZE = 1280
 
 # toggle for session managements that minimizes the number of sessions opened/closed
 USE_LONG_SESSIONS = os.getenv("DBT_DATABRICKS_LONG_SESSIONS", "True").upper() == "TRUE"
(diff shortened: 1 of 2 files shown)


benc-db:
It may take me a few days to get to reviewing this, but thank you for your contribution.  I'll run the integration tests today.

benc-db:
Please take a look at the unit test failure :)

benc-db:
I think the key thing is that, if the test doesn't pass on Ubuntu, is it not passing because we're not setting the test up correctly, is the new behavior not compatible with ubuntu, or is the change regressing something.  So long as we're not in the third category, we can skip the test (@pytest.mark.skip), though ideally we would detect the OS and only skip for Ubuntu.  If we're in the second category, and the test is just pointing this out, we'll also want to document the limitation.

benc-db:
> After looking at the failure again, I think if you can [patch](https://docs.python.org/3/library/unittest.mock.html) keyring to just read/write the secret to files in the test, it should work regardless of OS. Ultimately the test just needs to validate your slice and rebuild logic works, regardless of keyring implementation.

And if you can't get that to work, I can update the tests myself, but might be a couple of weeks until I can get to it.

thijs-nijhuis-shell:
Thanks @benc-db, I will try to implement the patch option than. Thanks for the feedback.

thijs-nijhuis-shell:
Hi @benc-db, I think the issue is solved now. I created the Mock Keyring Backend class that will write to files in a temp directory (copied some a bit of the logic from the integration test). Also expanded the UT a bit by also testing if the delete cleans-up the password properly. 
The UT now succeeds on the GitHub action (when I trigger it from my repo) as well as on my laptop.

benc-db:
Looks good, I'm going to add a changelog entry and merge.

benc-db:
Closing in favor of https://github.com/databricks/dbt-databricks/pull/599 (moved to this fork and added Changelog).  Thanks @thijs-nijhuis-shell!

PR #604, opened 2024-03-04
Ensure passing on_schema_change tests

### LINKED ISSUE #594: Not all columns populated on column addition
### Describe the bug
Rerunning dbt run with changed columns leads to dropped columns.

### Steps To Reproduce
Change schema of table (add new column) and rerun table only populates a limited set of columns.
old table1:
a
b
d

Adding new column:
c

D doesn't get populated.

### Expected behavior
All columns populated in new table.

### System information
```
Core:
  - installed: 1.7.8
  - latest:    1.7.8 - Up to date!

Plugins:
  - databricks: 1.7.8 - Up to date!
  - spark:      1.7.1 - Up to date!
```

Python 3.11.4

### Additional context

benc-db:
### Description

I replaced our old integration tests for incremental on_schema_change with functional tests from dbt-core, to try to suss out where the regression identified in #594 was coming from.  Found that we were failing some of these tests, regardless of changes in 1.7.8, so I have endeavored to update our incremental strategies sql to pass these tests, with one exception: sync_all_columns has never worked with dropping columns due to Delta not supporting column deletion.  Actually, this capability is in preview, so at some point I'll address that as well, but for now I found that existing tests were failing for column mismatches.  This PR fixes those failing tests, and now I believe fixes #594 

Edit:
Resolves #594 

### Checklist


dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -51,24 +51,32 @@
...
+        {%- do common_columns.append(dest_col) -%}
+      {%- else -%}
+        {%- do common_columns.append('NULL') -%}
       {%- endif -%}
     {%- endfor -%}
-    {%- set dest_cols_csv = common_columns | map(attribute='quoted') | join(', ') -%}
+    {%- set dest_cols_csv = dest_columns | join(', ') -%}
+    {%- set source_cols_csv = common_columns | join(', ') -%}
     insert into table {{ target_relation }} ({{ dest_cols_csv }})
-    select {{dest_cols_csv}} from {{ source_relation }}
```
andrefurlan-db, line 62: was this just wrong before?

benc-db: I merged a community PR a couple of weeks ago that tried to fix an issue where, when the on_schema_change strategy was 'ignore', it would fail for column mismatch between the target and the source (in contradiction to the dbt docs).  The change introduced a break for 'append' and 'sync_all_columns' when the incremental strategy is set to 'append'.  So until now, we could not pass the functional tests for 'append' and 'ignore' at the same time, but we didn't see it because our integration tests were modified to a weaker standard than the dbt-core tests.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -51,24 +51,32 @@
...
-    {%- set source_columns = adapter.get_columns_in_relation(source_relation) | map(attribute="name") -%}
-    {%- for dest_col in adapter.get_columns_in_relation(target_relation) -%}
-      {% if dest_col.name in source_columns %}
-          {%- if common_columns.append(dest_col) -%}{%- endif -%}
+    {%- set source_columns = adapter.get_columns_in_relation(source_relation) | map(attribute="quoted") | list -%}
+    {%- set dest_columns = adapter.get_columns_in_relation(target_relation) | map(attribute="quoted") | list -%}
+    {{ log("source_columns: " ~ (source_columns | join(', '))) }}
+    {{ log("dest_columns: " ~ (dest_columns | join(', '))) }}
+    {%- for dest_col in dest_columns -%}
+      {%- if dest_col in source_columns %}
```
andrefurlan-db, line 59: so if source has a column not in destination, what do we do?

benc-db: source column will be in destination unless the setting is ignore (because the schema is altered to include them before this called).  If setting is ignore, we are intentionally not taking source columns that are not in destination.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -100,7 +108,24 @@
...
+        *
+        {%- else %}
+        {%- for column in (source_columns | map(attribute='quoted')) %}
+            {{ column }} = DBT_INTERNAL_SOURCE.{{ column }}
+            {%- if not loop.last %}, {%- endif %}
+        {%- endfor %}
+        {% endif %}
+
+      when not matched then insert 
+      {%- if on_schema_change == 'ignore' %}
```
andrefurlan-db, line 121: is the default now scheme_change?

andrefurlan-db (dismissed):
LGTM

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -49,26 +49,34 @@
...
     {%- set common_columns = [] -%}
-    {%- set source_columns = adapter.get_columns_in_relation(source_relation) | map(attribute="name") -%}
-    {%- for dest_col in adapter.get_columns_in_relation(target_relation) -%}
-      {% if dest_col.name in source_columns %}
-          {%- if common_columns.append(dest_col) -%}{%- endif -%}
+    {%- for dest_col in dest_columns -%}
+      {%- if dest_col in source_columns -%}
+        {%- do common_columns.append(dest_col) -%}
+      {%- else -%}
+        {%- do common_columns.append('NULL') -%}
```
nimeshpatni, line 63: Hi, This might cause an explicit NULL insertion into the destination table and skip default values if set in schema ?

benc-db: Good call out.  I will add a test for this, and if necessary, spin a new release with the fix.

benc-db: The docs I came across said that the number of columns were assumed to match the target table, since they are not matched by name, but by column order.  I've switched to using DEFAULT in my new PR, which I think covers your concern.  And you're right, this is a feature we're looking into adding this year (full implementation of the column spec across all materializations).

nimeshpatni: One more thing, with this feature request, we'll now expect/restrict users to always have aliases in their models.!

```sql
-- A valid code dbt-databricks supports till now, since there is no column check in source/destination:
create or replace temporary view test_view__dbt as
  select cast('002' as int), 'Test name';

insert into `test_db`.`test_table` -- (Schema: id, name)
  select * from test_view__dbt;
```

If this is known, below code might work for most of the cases:

```jinja
{% macro get_insert_into_sql(source_relation, target_relation) %}
    {%- set source_columns = adapter.get_columns_in_relation(source_relation) | map(attribute="quoted") -%}
    {%- set dest_columns = adapter.get_columns_in_relation(target_relation) | map(attribute="quoted") -%}    
    {%- set common_columns = set(source_columns).intersection(dest_columns) -%}
    
    {%- set common_columns_csv = common_columns | join(', ') -%}

    insert into table {{ target_relation }} ({{ common_columns_csv }})
      select {{common_columns_csv}} from {{ source_relation }}
{% endmacro %}
```

Benefits:
1. Number of columns & the order in source and target - will be same
2. No unexpected overrides - default feature will work
3. The size of source and destination columns won't matter for iteration performance
4. Will return only the unique ones - Though its never a possibility of having duplicate columns in source/target, but no harm too
5. Might be a faster - No conversion of generator or set to list, less space and control statements

Cons:
1. Sequence in common_columns output won't be in order - this won't impact the final query.

If looks good, please give it a try. Thanks

benc-db: hmm, interesting, let me look into this.

PR #607, opened 2024-03-06
Prefer DEFAULT to NULL when inserting into, actually validate append

benc-db:
### Description

A github user pointed out that technically using NULL could lead to different behavior than *, where DEFAULT would be honored for missing columns, so I have adjusted `insert into` to use DEFAULT instead.  Also, noticed that at some point I lost my config that made the append tests actually test append, so added that back.  The second half of the 'append columns' test for the append strategy validates that DEFAULT is used and does not break.

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -60,7 +60,7 @@
       {%- if dest_col in source_columns -%}
         {%- do common_columns.append(dest_col) -%}
       {%- else -%}
-        {%- do common_columns.append('NULL') -%}
+        {%- do common_columns.append('DEFAULT') -%}
       {%- endif -%}
     {%- endfor -%}
     {%- set dest_cols_csv = dest_columns | join(', ') -%}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## dbt-databricks 1.7.10 (TBD)
+
+### Fixes
+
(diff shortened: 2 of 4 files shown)


dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -60,7 +60,7 @@
       {%- if dest_col in source_columns -%}
         {%- do common_columns.append(dest_col) -%}
       {%- else -%}
-        {%- do common_columns.append('NULL') -%}
+        {%- do common_columns.append('DEFAULT') -%}
```
nimeshpatni, line 63: Hi,
This seems to be a breaking change, "NULL" (String value) or "DEFAULT" (String value) won't solve the purpose. For all the unmatched columns, we will end up having multiple "DEFAULT" (String) values as columns that can raise conflict/ambiguous column error.

Default value works when we don't pass the column in the argument of insert statement. Could you please try [this simple resolution](https://github.com/databricks/dbt-databricks/pull/604#discussion_r1515123621):

` {%- set common_columns_csv = set(source_columns).intersection(dest_columns) | sort | join(', ') -%}`

benc-db: good call

benc-db: They're not inserted as strings, but as literals.

Example from test run:
```
insert into table `peco`.`test17110372962014884605_test_incremental_on_schema_change`.`incremental_append_new_columns_remove_one` (`id`, `field1`, `field2`, `field3`, `field4`)
select `id`, `field1`, DEFAULT, `field3`, `field4` from `incremental_append_new_columns_remove_one__dbt_tmp`
```

PR #615, opened 2024-03-19
[Bug] Persist docs performance issue

### LINKED ISSUE #608: decrease in job performance for dbt docs generate
### Describe the bug
When upgrading environments in dbt from v1.6 to v1.7, there were 4 times more log lines in the output of a job running the "dbt docs generate" step. At first glance, it seems v1.7.7 is scanning more metadata. Search for "describe table" term in both log files. v1.6.9 log doesn't have any matches, v1.7.7 has 3.1K matches. 

This has led to a performance difference. For example a customer reported that on 1.6.9 the job run was taking 32 seconds, after upgrading it increased to 11+ minutes. 

What dbt Support is seeing:
I looked into this with our engineers, and wanted to submit this case based on our findings. 

Our best guess is that this is happening due to a change in behavior in the dbt-databricks adapter. The noticeable difference in the logs is the huge increase in the number of "describe table <relation>..." statements being made against Databricks. We tracked this down to how dbt-databricks persists docs, in particular, how it gets the column comments from Databricks. It appears to do this one table at a time in 1.7. The _get_columns_for_catalog() function defined in the Databricks adapter seems to be getting called a lot more in 1.7. We are not sure why.

We are not sure if this could be something dbt-databricks implemented between 1.6 and 1.7, or if it could be something implemented because of a change we made. We would appreciate if someone could take a look with your better understanding of the adapter.

### Steps To Reproduce
Create a job that has "Generate docs on run" checked. 
Run the job on 1.6 and note the time the dbt docs generate step takes
Run the job on 1.7 and note the time that dbt docs generate steps takes

I tested with a very simple setup and only a handful of models. I could see each test that 1.7 ran a few seconds slower than 1.6. This is not significant when talking about a simple model setup. However, for a customer with a larger data set and a multitude of models the timing difference is considerable. 

### Expected behavior
After upgrading from 1.6 to 1.7 the expected behavior would be for dbt docs generate step to not increase significantly.

debug logs attached, please keep in mind this is the smaller dataset I was testing with as mentioned under steps to reproduce. The difference is very minuscule (6s vs 9s). I do have the full customer logging for debugging, but cannot attach it to here. Please feel free to reach out to me if you need more information. 

### System information
```
1.7.8
```
databricks version: 1.7.8

*Please note, customer was using 1.7.7 when they first saw this. 

I was using macOS Sonoma when testing

### Additional context
Please feel free to reach out with any questions. Thank you!

[1.6-debug.log](https://github.com/databricks/dbt-databricks/files/14517589/1.6-debug.log)
[1.7-debug.log](https://github.com/databricks/dbt-databricks/files/14517590/1.7-debug.log)

mikealfare:
Resolves #608

### Description

It appears that incremental catalog generation actually takes longer in databricks than generating the catalog all at once. This appears to be partially due to the column comment queries that are required while building the catalog. Running one query to build this all at once is faster once there is more than roughly 5 models. Given how small this break even number, and how infrequently it would be beneficial, it's probably a lot easier to simply turn off incremental catalog generation.

### Checklist


benc-db:
Closed in favor of https://github.com/databricks/dbt-databricks/pull/616

PR #631, opened 2024-04-02
Supporting tags

### LINKED ISSUE #606: Feature request: Sync dbt model tags with Databricks table/view tags
### Description
Dbt allow you to [tag](https://docs.getdbt.com/reference/resource-configs/tags) your models, and you can also [tag](https://learn.microsoft.com/en-us/azure/databricks/data-governance/unity-catalog/tags) your tables and views in Databricks. There should be a way to apply dbt tags to the tables and views in Databricks.

### Alternative
I have considered implementing a custom solution using:
1. Python to read the tags from the dbt models
2. [Databricks-sql-cli](https://docs.databricks.com/en/dev-tools/databricks-sql-cli.html) to read the tags from databricks
3. Then [databricks-sql-cli](https://docs.databricks.com/en/dev-tools/databricks-sql-cli.html) again to update the tags in databricks such that they are aligned with the dbt tags

### Benefits
This is useful for filtering the tables in databricks.

I would like to contribute in creating a seamless syncing process.

benc-db:
Resolves #606

### Description

Adds support for setting tags (for discovery, governance) on tables/views.  Unfortunately not yet supported on MV/STs, but I built the support in the same way I do on_config_change for them, so it should be trivial to add when they become supported.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Support `on_config_change` for materialized views, expand the supported config options ([536](https://github.com/databricks/dbt-databricks/pull/536)))
 - Support `on_config_change` for streaming tables, expand the supported config options ([569](https://github.com/databricks/dbt-databricks/pull/569)))
 - Support insert overwrite on SQL Warehouses ([623](https://github.com/databricks/dbt-databricks/pull/623))
+- Support Databricks tags for tables/views/incrementals ([631](https://github.com/databricks/dbt-databricks/pull/631))
 
 ### Under the Hood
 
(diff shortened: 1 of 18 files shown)


benc-db:
@jackyhu-db Yunbo suggested I add you as a reviewer while Ray is out.  Apologies that dbt code is a mix of python and jinja templating that can be pretty hard to parse.  Basic idea is that we want to support setting tags on tables/views/incremental tables.  Today we can only set tags with alter statements.  For table and views, since they are replaced on every run, we simply SET the tags the user set in databricks_tags (since dbt has its own idea of tags).  For incremental, if we have an existing table, we UNSET any tags that are present on the table but not in the config, and SET all tags that are present in the config.

jackyhu-db (approved):

PR #665, opened 2024-05-14
Remove dependence on unittest from Adapter tests

benc-db:
### Description

Remove dependence of the adapter tests on unittest (and instead depend on pytest).  Also updates some lagging references to the shared adapter libraries, that were still pointing at dbt-core.

### Checklist


benc-db:
@jackyhu-db when I went to add tests per your request on the other PR, I saw that this test suite had not been updated to our latest standards, and figured I would do that refactoring separate from adding new tests.

PR #668, opened 2024-05-16
Handle empty comments on MV/ST better

benc-db:
### Description

Prior to this change, when a user had not put a description on his model, we actually set the table comment as ''.  This change ensures empty strings are not treated as legitimate comments.

### Checklist

### DIFF
--- a/tests/functional/adapter/materialized_view_tests/fixtures.py
+++ b/tests/functional/adapter/materialized_view_tests/fixtures.py
@@ -30,7 +30,6 @@ def query_relation_type(project, relation: BaseRelation) -> Optional[str]:
 materialized_view = """
 {{ config(
     materialized='materialized_view',
-    description='this is a materialized view',
     partition_by='id',
     schedule = {
         'cron': '0 0 * * * ? *',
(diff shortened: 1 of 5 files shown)


tests/functional/adapter/materialized_view_tests/fixtures.py
```diff
@@ -30,7 +30,6 @@ def query_relation_type(project, relation: BaseRelation) -> Optional[str]:
 materialized_view = """
 {{ config(
     materialized='materialized_view',
-    description='this is a materialized view',
```
benc-db, line 33: Removed as this is not a valid way to specify a comment, and thus was not even addressed in the tests.

tests/functional/adapter/streaming_tables/test_st_changes.py
```diff
@@ -33,6 +32,7 @@ def check_start_state(project, streaming_table):
         _check_tblproperties(results.config["tblproperties"], {"key": "value"})
         assert results.config["refresh"].cron == "0 0 * * * ? *"
         assert results.config["refresh"].time_zone_value == "Etc/UTC"
+        assert results.config["comment"].comment == None
```
benc-db, line 35: The test modification is to ensure that when we don't specify a comment, as in this test, we actually read that as None, rather than ''

PR #686, opened 2024-05-28
Liquid clustering responsive to config change on incremental

benc-db:
### Description

When a user changes the columns that they specify for liquid clustering on an incremental table, prior to this PR, the change does not get propagated.  Since there is no set/unset behavior (so no need to compute a difference), and calling `describe extended` to check for existing cluster by columns is likely as or more expensive that calling alter table, I decided to always run an alter statement for liquid clustering on subsequent runs of incremental models.  In my testing it is a fast (sub-second) no-op when a change is not required.

### Checklist


dbt/adapters/databricks/relation_configs/tblproperties.py
```diff
@@ -28,6 +28,7 @@ class TblPropertiesConfig(DatabricksComponentConfig):
         "pipeline_internal.catalogType",
         "pipelines.metastore.tableName",
         "pipeline_internal.enzymeMode",
+        "clusteringColumns",
```
benc-db, line 31: cluster by columns are stored in tblproperties.  We don't want users attempting to manipulate them outside of specifying liquid_clustered_by

dbt/include/databricks/macros/adapters/python.sql
```diff
@@ -72,7 +72,7 @@ writer.saveAsTable("{{ target_relation }}")
     {%- endif %}
 .partitionBy({{ partition_by }})
 {%- endif -%}
-{%- if liquid_clustered_by -%}
+{%- if liquid_clustered_by and not is_incremental() -%}
```
benc-db, line 75: if is_incremental, it is pointless to liquid cluster by, as this is a temp table that will be merged into the table that we actual want to liquid cluster by, and optimize will not be called before then.

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -78,6 +78,7 @@
       Also, why does not either drop_relation or adapter.drop_relation work here?!
       --#}
     {%- endif -%}
+    {% do apply_liquid_clustered_cols(target_relation) %}
```
benc-db, line 81: only call this logic on an is_incremental run, otherwise liquid clustering is already specified in the create statement.

dbt/include/databricks/macros/relations/liquid_clustering.sql
```diff
@@ -0,0 +1,37 @@
...
+    {%- endif -%}      
+    {%- call statement('set_cluster_by_columns') -%}
+        ALTER {{ target_relation.type }} {{ target_relation }} CLUSTER BY (
+        {%- for item in cols -%}
+            {{ item }}
+            {%- if not loop.last -%},{%- endif -%}
+        {%- endfor -%}
+        )
+    {%- endcall -%}
+  {%- elif not target_relation.is_hive_metastore() and file_format == "delta" and not partition_by -%}
```
benc-db, line 32: If it is possible that this table was previously liquid clustered, and now the user is not specifying cluster by columns, unset the cluster by columns.  Since liquid clustering only works on delta, UC, and without partitions, we only want to run this statement if none of those things are true, otherwise we'll get an error.

dbt/include/databricks/macros/relations/table/create.sql
```diff
@@ -58,22 +58,3 @@
     )
   {%- endif %}
 {%- endmacro -%}
-
-{% macro liquid_clustered_cols(label, required=false) -%}
```
benc-db, line 62: simplified this stuff as we the 'label' is well know and 'required' was unused.

benc-db:
@jackyhu-db can you take a look?  I provided review comments explaining the changes.

dbt/include/databricks/macros/relations/liquid_clustering.sql
```diff
@@ -0,0 +1,37 @@
+{% macro liquid_clustered_cols() -%}
+  {%- set cols = config.get('liquid_clustered_by', validator=validation.any[list, basestring]) -%}
+  {%- if cols is not none %}
+    {%- if cols is string -%}
+      {%- set cols = [cols] -%}
+    {%- endif -%}
+    CLUSTER BY (
+    {%- for item in cols -%}
```
jackyhu-db, line 8: can you use cols | join(', ')

benc-db: this is the value of bringing in a fresh set of eyes...I've just been copy pasting this pattern, but I suspect your suggestion will work.  Trying now.

jackyhu-db (approved):

dbt/include/databricks/macros/relations/liquid_clustering.sql
jackyhu-db: I saw default of `partition_by` is `None`. shall it be `partition_by is None`?

benc-db: Tried it; unfortunately this is a place where jinja differs from python just a bit, and complains that None is not a test that it knows of.

dbt/include/databricks/macros/relations/liquid_clustering.sql
benc-db: Yep, totally works.  Will do another PR where I replace this pattern elsewhere, as join is much cleaner.

jackyhu-db (approved):

PR #688, opened 2024-05-29
Reverting the decision to remove dynamic setting

### LINKED ISSUE #687: Regression in insert_overwrite incremental strategy
### Describe the bug
After upgrading to `dbt-databricks` 1.8.0, incremental models using the `insert_overwrite` strategy replace the entire table rather than adding/overwriting specific partitions.

### Steps To Reproduce
We upgraded yesterday to `dbt-core` and `dbt-databricks` 1.8.0. We had several incremental models using the `insert_overwrite` strategy, with config like

```
{{ config(
  partition_by=['date_partition'],
  materialized='incremental',
  incremental_strategy = 'insert_overwrite'
) }}
```

Before the upgrade these incremental models were running normally; each incremental run would add a new partition and preserve all historical partitions.

On the next runs after the upgrade, we observed the historical data was lost. For instance, if an incremental job produced `date_partition=2024-05-28`, only that partition would be present in the table. We could recreate the historical data by running the model with `--full-refresh`, but the next incremental run would again remove all the historical data.

I believe this is due to removing `set spark.sql.sources.partitionOverwriteMode = DYNAMIC` in #623. When we added this setting to our Databricks cluster, the previous behavior was restored.

I know that property isn't supported in SQL warehouses, but this is a significant regression in the behavior of this adapter.

### Expected behavior
When an incremental model runs with the `insert_overwrite` strategy, only specific partitions should be added/overwritten, not the entire table. This should be the default behavior and not require custom configuration to support.

### System information
```
Core:
  - installed: 1.8.0
  - latest:    1.8.1 - Update available!

Plugins:
  - databricks: 1.8.0 - Up to date!
```

Clients are using OS X, the DBT models are running on a Databricks cluster with DBR 14.3

```
Python 3.11.7
```

benc-db:
Resolves #687

### Description

We had removed the spark setting for insert overwrite so that the strategy could be compatible with SQL Warehouses, but upon further reflection, it's unclear if the strategy is even useful for incremental tables without the setting, as it will just replace the entire table.  The consequence of having removed the spark setting was regressing users that were successfully using the strategy with clusters.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## dbt-databricks 1.8.2 (TBD)
+
+### Fixes
+
+- Undo the removal of spark.sql.sources.partitionOverwriteMode = DYNAMIC ([688](https://github.com/databricks/dbt-databricks/pull/688))
+
 ## dbt-databricks 1.8.1 (May 29, 2024)
 
 ### Features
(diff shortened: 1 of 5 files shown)


ajsquared:
🙏 Appreciate the quick turnaround on this!

jackyhu-db (dismissed):

Issue #684, opened 2024-05-28
Python model doesn't have cancel method


gaoshihang:
### Describe the bug
We found that when a python model is canceled from terminal, it doesn't have a cancel method to cancel the job_cluster on Databricks. 

We think this could cause resource/cost waste and data error in some case.

### Steps To Reproduce
1.include a SQL model and a Python model in project.
2.execute them together.
3.use Control + C to exit the DBT execution.
4.you can see the SQL model on SQL warehouse has been canceled successfully, but Python Model on Job Cluster is still running. 
5.and the DBT execution will stuck until the Python model is finished.

### Expected behavior
All model should be canceled no matter SQL model or Python model.

### Screenshots and log output
<img width="1714" alt="image" src="https://github.com/databricks/dbt-databricks/assets/20013931/b96f103f-e477-46cc-85eb-3368855f74fb">

### System information
<img width="602" alt="image" src="https://github.com/databricks/dbt-databricks/assets/20013931/d60aca91-6a9b-47aa-a66c-ea3147207451">

**The operating system you're using:**

**The output of `python --version`:**

benc-db:
Thanks for the report, I had a suspicion this might be the case.  Guess we need to figure out where dbt catches the cancel and put a call to the job api there.

gaoshihang:
Thank you very much! please let me know too!

gaoshihang:
Hi @benc-db ,
I think dbt is execute the cancel action in https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/runnable.py#L354
Can we add a list to record all the python model jobs in DatabricksAdaptor: https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/impl.py#L145
![image](https://github.com/databricks/dbt-databricks/assets/20013931/7b9e70ed-1cdf-4200-80aa-247bc747f39d)


And use Databricks api to kill them in  https://github.com/dbt-labs/dbt-core/blob/main/core/dbt/task/runnable.py#L354

benc-db:
I think if we override cancel_open() to call super, and then call the jobs api, that should do it.  Right now we inherit cancel_open from the SQLConnectionManager.

gaoshihang:
yeah, I got your mean. now in DatabricksConnectionManager, it inherit SparkConnectionManager, then inherit SQLConnectionManager, so it calls the cancel_open() in it.
[...]

PR #690, opened 2024-05-29
cancel python model job when dbt exit

gaoshihang:
Resolves https://github.com/databricks/dbt-databricks/issues/684

### Checklist

### DIFF
--- a/dbt/adapters/databricks/auth.py
+++ b/dbt/adapters/databricks/auth.py
@@ -15,9 +15,11 @@ from requests.auth import AuthBase
 
 class token_auth(CredentialsProvider):
     _token: str
+    _host: str
 
-    def __init__(self, token: str) -> None:
+    def __init__(self, token: str, host: str) -> None:
         self._token = token
+        self._host = host
 
     def auth_type(self) -> str:
         return "token"
@@ -29,7 +31,7 @@ class token_auth(CredentialsProvider):
     def from_dict(raw: Optional[dict]) -> Optional[CredentialsProvider]:
         if not raw:
             return None
-        return token_auth(raw["token"])
+        return token_auth(raw["token"], raw["host"])
 
     def __call__(self, _: Optional[Config] = None) -> HeaderFactory:
         static_credentials = {"Authorization": f"Bearer {self._token}"}
--- a/dbt/adapters/databricks/auth.py
+++ b/dbt/adapters/databricks/auth.py
@@ -25,7 +25,7 @@ class token_auth(CredentialsProvider):
         return "token"
 
     def as_dict(self) -> dict:
-        return {"token": self._token}
+        return {"token": self._token, "host": self._host}
 
     @staticmethod
     def from_dict(raw: Optional[dict]) -> Optional[CredentialsProvider]:
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -1,4 +1,5 @@
 import decimal
+import json
 import os
 import re
 import sys
@@ -76,6 +77,10 @@ from dbt_common.exceptions import DbtInternalError
 from dbt_common.exceptions import DbtRuntimeError
 from dbt_common.utils import cast_to_str
 from requests import Session
+
+from urllib3.util.retry import Retry
+from requests.adapters import HTTPAdapter
+from dbt.adapters.databricks.__version__ import version
 
 if TYPE_CHECKING:
     from agate import Table
@@ -474,6 +479,84 @@ class DatabricksDBTConnection(Connection):
 class DatabricksConnectionManager(SparkConnectionManager):
     TYPE: str = "databricks"
     credentials_provider: Optional[TCredentialProvider] = None
+
+    def cancel_open(self) -> List[str]:
+        run_id_dir = f"/Shared/dbt_python_model/{self.profile.credentials.database}/{self.profile.credentials.schema}/run_ids/"
+        run_ids = self._list_files(run_id_dir)
+        print(run_ids)
+        for run_id in run_ids:
+            self._cancel_run_id(run_id_dir, run_id)
+
+        return super().cancel_open()
+
+    def _cancel_run_id(self, run_id_dir: str, run_id: str) -> None:
+        retry_strategy = Retry(total=4, backoff_factor=0.5)
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        session = Session()
+        session.mount("https://", adapter)
+        extra_headers = {
+            "User-Agent": f"dbt-databricks/{version}",
+            "Authorization": f"Bearer {self.credentials_provider.as_dict()['token']}"
+        }
+
+        response = session.post(
+            f"https://{}/api/2.0/jobs/runs/cancel",
+            headers=extra_headers,
+            json={
+                "run_id": run_id
+            },
+        )
+        if response.status_code != 200:
+            raise DbtRuntimeError(f"Cancel run id failed.\n {response.content!r}")
+
+        self._delete_file(f'{run_id_dir}{run_id}')
+
+    def _delete_file(self, path: str) -> None:
+        retry_strategy = Retry(total=4, backoff_factor=0.5)
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        session = Session()
+        session.mount("https://", adapter)
+        extra_headers = {
+            "User-Agent": f"dbt-databricks/{version}",
+            "Authorization": f"Bearer {self.credentials_provider.as_dict()['token']}"
+        }
+        response = session.post(
+            f"https://{host}/api/2.0/workspace/delete",
+            headers=extra_headers,
+            json={
+                "path": path
+            },
+        )
+        if response.status_code != 200:
+            raise DbtRuntimeError(f"Error deleting file.\n {response.content!r}")
+
+
+    def _list_files(self, path: str) -> List[str]:
+        retry_strategy = Retry(total=4, backoff_factor=0.5)
+        adapter = HTTPAdapter(max_retries=retry_strategy)
+        session = Session()
+        session.mount("https://", adapter)
+        extra_headers = {
+            "User-Agent": f"dbt-databricks/{version}",
+            "Authorization": f"Bearer {self.credentials_provider.as_dict()['token']}"
+        }
+
+        response = session.get(
+            f"https://{host}/api/2.0/workspace/list",
+            headers=extra_headers,
+            json={
+                "path": path
+            },
+        )
+        if response.status_code != 200:
+            raise DbtRuntimeError(f"Error list workspace.\n {response.content!r}")
+
+        return_list = list()
+        for item in response.json()["objects"]:
+            path = str(item["path"])
+            run_id = path.split('/')[-1]
+            return_list.append(run_id)
+        return return_list
 
     def compare_dbr_version(self, major: int, minor: int) -> int:
         version = (major, minor)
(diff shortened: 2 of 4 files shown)


gaoshihang:
Hi @benc-db Could you please help review on this PR? the code is not ready, but I want you look it first, see if the method is fine. it related to this issue: https://github.com/databricks/dbt-databricks/issues/684

I think we can utilize Databricks workspace, we create a job_run_ids dir, each python model create one file named run_id in the job_run_ids dir.

When dbt canceled, we read from this job_run_ids dir, and cancel all run_id in it, then we delete the file.

gaoshihang:
And I have tested it, the job can be canceled, and fail-fast works ok.

dbt/adapters/databricks/connections.py
```diff
@@ -475,6 +480,84 @@ class DatabricksConnectionManager(SparkConnectionManager):
...
     credentials_provider: Optional[TCredentialProvider] = None
 
+    def cancel_open(self) -> List[str]:
+        run_id_dir = f"/Shared/dbt_python_model/{self.profile.credentials.database}/{self.profile.credentials.schema}/run_ids/"
+        run_ids = self._list_files(run_id_dir)
+        print(run_ids)
+        for run_id in run_ids:
+            self._cancel_run_id(run_id_dir, run_id)
+
+        return super().cancel_open()
```
benc-db, line 490: Below you raise an exception on a non-200, but that will interrupt cancelling the other operations.  Better to log a warning on non-200 I think.

dbt/adapters/databricks/connections.py
```diff
@@ -475,6 +480,84 @@ class DatabricksConnectionManager(SparkConnectionManager):
...
+
+    def _cancel_run_id(self, run_id_dir: str, run_id: str) -> None:
```
benc-db, line 492: since neither of these methods rely on anything in self, I think I would prefer them as static functions in python_submissions.py, so they are closer to the code that they are cleaning up.

benc-db:
> Hi @benc-db Could you please help review on this PR? the code is not ready, but I want you look it first, see if the method is fine. it related to this issue: #684
> 
> I think we can utilize Databricks workspace, we create a job_run_ids dir, each python model create one file named run_id in the job_run_ids dir.
> 
> When dbt canceled, we read from this job_run_ids dir, and cancel all run_id in it, then we delete the file.

If we can locate the target folder, I think I would prefer writing there, so that we don't rely on an API operation to store and retrieve.  Also, I think we need to lock around writing the file to ensure clean operation when kicking off multiple python models concurrently.

benc-db:
@mikealfare we're trying to figure out how to cancel python jobs as part of cleanup, similar to what is done for SQL queries when the user ctrl-Cs.  Is there a better way to communicate run_ids from the python job helper to the connection manager?  We were wondering if maybe there was some global state that would help the python job helper figure out the target directory?

benc-db:
@jtcohen6 as well

gaoshihang:
> > Hi @benc-db Could you please help review on this PR? the code is not ready, but I want you look it first, see if the method is fine. it related to this issue: #684
> > I think we can utilize Databricks workspace, we create a job_run_ids dir, each python model create one file named run_id in the job_run_ids dir.
> > When dbt canceled, we read from this job_run_ids dir, and cancel all run_id in it, then we delete the file.
> 
> If we can locate the target folder, I think I would prefer writing there, so that we don't rely on an API operation to store and retrieve. Also, I think we need to lock around writing the file to ensure clean operation when kicking off multiple python models concurrently.

Hi @benc-db Many thanks for you help. I didn't found a way to get target path in python_submission.py, I'll try to find today.

benc-db:
> > > Hi @benc-db Could you please help review on this PR? the code is not ready, but I want you look it first, see if the method is fine. it related to this issue: #684
> > > I think we can utilize Databricks workspace, we create a job_run_ids dir, each python model create one file named run_id in the job_run_ids dir.
> > > When dbt canceled, we read from this job_run_ids dir, and cancel all run_id in it, then we delete the file.
> > 
> > 
> > If we can locate the target folder, I think I would prefer writing there, so that we don't rely on an API operation to store and retrieve. Also, I think we need to lock around writing the file to ensure clean operation when kicking off multiple python models concurrently.
> 
> Hi @benc-db Many thanks for you help. I didn't found a way to get target path in python_submission.py, I'll try to find today.

I'm reaching out to dbt Labs folks to see if there is a better way.  In particular, @mikealfare has worked on the dbt-spark adapter, so if we figure it out, it might be good for that library too.

gaoshihang:
Hi @benc-db, I try to print self.parsed_model in python_submission.py, and I found there is a 'compiled_path' in it, do you think we can utilize this to find TARGET path?

For example, here I set TARGET path as 'target_2'
image
image

gaoshihang:
Hi @benc-db I'm thinking can we use a class static variable to share run_ids?
image

benc-db:
> Hi @benc-db I'm thinking can we use a class static variable to share run_ids? image

Good point.  I'm generally so anti-global state that I didn't even think of it :P.  We still need to protect it from concurrency issues, but global state is going to be our best bet until we get support from dbt-core, and better to store in memory than cloud files.

gaoshihang:
> > Hi @benc-db I'm thinking can we use a class static variable to share run_ids? image
> 
> Good point. I'm generally so anti-global state that I didn't even think of it :P. We still need to protect it from concurrency issues, but global state is going to be our best bet until we get support from dbt-core, and better to store in memory than cloud files.

Yes...I don't want to use it too...but seems like if we don't want to do things in dbt-core, it's the only way we can share the state in two class..Let's me write some code in this way!

gaoshihang:
Hi @benc-db, I revise the code, using a global variable to store all the run id, then cancel them in ConnectManager. please help to review this, thank you very much!

dbt/adapters/databricks/connections.py
```diff
@@ -475,6 +476,17 @@ class DatabricksConnectionManager(SparkConnectionManager):
     TYPE: str = "databricks"
     credentials_provider: Optional[TCredentialProvider] = None
 
+    def cancel_open(self) -> List[str]:
+        from dbt.adapters.databricks.python_submissions import BaseDatabricksHelper
```
benc-db, line 480: import at the top please.  We only import in place like this if the thing we're importing is too heavy to do at start up.

dbt/adapters/databricks/connections.py
```diff
@@ -475,6 +476,17 @@ class DatabricksConnectionManager(SparkConnectionManager):
...
+
+        for run_id in BaseDatabricksHelper.run_ids:
+            logger.info(f"cancel run id {run_id}")
```
benc-db, line 483: I think this can be debug, and we should mention that it's a python model job.

dbt/adapters/databricks/auth.py
```diff
@@ -15,9 +15,11 @@
 
 class token_auth(CredentialsProvider):
     _token: str
+    _host: str
```
benc-db, line 18: Why store this on the token?  It's already on the DatabricksCredentials.

benc-db (changes requested):
Mostly good, just some minor comments to clean up.

mikealfare:
> @mikealfare we're trying to figure out how to cancel python jobs as part of cleanup, similar to what is done for SQL queries when the user ctrl-Cs. Is there a better way to communicate run_ids from the python job helper to the connection manager? We were wondering if maybe there was some global state that would help the python job helper figure out the target directory?

It looks like we might be trying to do something similar [here](https://github.com/dbt-labs/dbt-bigquery/pull/1251) for `dbt-bigquery`. I haven't read through either PR in detail to know if this solves your problem, but I figured I'd link it here in the event it's helpful.

dbt/adapters/databricks/auth.py
gaoshihang: Seems like I can't get DatabricksCredentials in DatabricksConnectionManager.

I can just use self.credentials_provider in DatabricksConnectionManager, but there are no host in credentials_provider, so I put a host in the token_auth class.

Could you give me a direction how to get DatabricksCredentials in DatabricksConnectionManager?

dbt/adapters/databricks/connections.py
gaoshihang: done!

dbt/adapters/databricks/connections.py
gaoshihang: done!

dbt/adapters/databricks/auth.py
benc-db: The BaseDatabricksHelper has a copy of DatabricksCredentials.

benc-db: hmm, but that's an instance...let me think.

benc-db: I'm going to pull down a copy of this PR and see if i can figure it out.

gaoshihang: yes..think I can't get instance in DatabricksConnectionManager...

Thank you very much!

benc-db: Oh, did you already fix this?

dbt/adapters/databricks/connections.py
```diff
@@ -475,6 +477,14 @@ class DatabricksConnectionManager(SparkConnectionManager):
     TYPE: str = "databricks"
     credentials_provider: Optional[TCredentialProvider] = None
 
+    def cancel_open(self) -> List[str]:
+        for run_id in BaseDatabricksHelper.run_ids:
+            logger.debug(f"Cancel python model job: {run_id}")
+            BaseDatabricksHelper.cancel_run_id(run_id, self.credentials_provider.as_dict()['token'], self.credentials_provider.as_dict()['host'])
```
benc-db, line 483: ah, this is where you need to retrieve, and here you don't have an instance...we can use singleton pattern maybe?

dbt/adapters/databricks/auth.py
gaoshihang: No, I can't found a way.. so I still use the way put host in token_auth. so that the host can be retrieve from DatabricksConnectionManager.credentials_provider.

dbt/adapters/databricks/connections.py
benc-db: give me an hour to take a crack at refactoring this; I have an idea :)

gaoshihang: Singleton maybe its a way, Let me try some code.

gaoshihang: yeah! thank you very much!

benc-db:
This is annoying, but is actually teaching me a lot about our python model support lol.  It's taking me longer than I expected because I'm trying to get it to work with all credentials and all execution formats (i.e. commands vs notebooks)

gaoshihang:
> This is annoying, but is actually teaching me a lot about our python model support lol. It's taking me longer than I expected because I'm trying to get it to work with all credentials and all execution formats (i.e. commands vs notebooks)

No Rush! thanks for your support. please let me know when you done, and I can modify the code in this PR~

gaoshihang:
Hi @benc-db I'm reaching out to see if there are any things I can help!

benc-db:
> Hi @benc-db I'm reaching out to see if there are any things I can help!

I'll have a new PR up shortly...got stalled because weather knocked out my internet yesterday.  After I put up my PR, if you could download and validate that it works for your scenario, that would be great.

benc-db:
@gaoshihang closing in favor of #693.  Please take a look and verify it works for your use case.

gaoshihang:
Hi @benc-db , thank you very much! will do that later, and let you know!

PR #693, opened 2024-06-04
Cancel python models on Ctrl-C

### LINKED ISSUE #684: Python model doesn't have cancel method

benc-db:
Resolves #684 

### Description

When you cancel a running dbt job, running SQL queries are cancelled.  This PR extends #690 to propagate the cancel behavior to python models.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Undo the removal of spark.sql.sources.partitionOverwriteMode = DYNAMIC ([688](https://github.com/databricks/dbt-databricks/pull/688))
+- Cancel python jobs when dbt operation is canceled (thanks @gaoshihang for kicking this off!) ([693](https://github.com/databricks/dbt-databricks/pull/693))
 
 ## dbt-databricks 1.8.1 (May 29, 2024)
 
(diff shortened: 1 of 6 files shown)


gaoshihang:
Hi @benc-db I have tested our scenario, everything runs smoothly, the python model can be canceled! Thank you very much!
image
image

It's very pleasure to work with you.

gaoshihang:
Hi @benc-db Thanks for work on this. I still have a question, if we want to use this feature, which version should we use?

benc-db:
> Hi @benc-db Thanks for work on this. I still have a question, if we want to use this feature, which version should we use?
Hasn't been released yet, will be part of 1.8.2.

gaoshihang:
@benc-db Thanks!

PR #699, opened 2024-06-10
Debug logging for expired SPA and store refresh token

### LINKED ISSUE #640: Ongoing OAuth Bug Tracking
I'm looking to get a census of users ability to connect via OAuth

If you use or have tried to use OAuth and see this bug, please reply with the following:

Cloud: <AWS|Azure|GCP>
Mechanism: <M2M|U2M with default app|U2M with custom app>
Success/Failure:
dbt version:
Last date you tried to connect with OAuth:

thijs-nijhuis:
Resolves #640 

### Description
When using an Azure App Registration with a 'Single-page Application' authentication platform as the oAuth endpoint, the token will always expire after 24h. It is not possible to fetch a new token after this and a new consent (and external browser login) is required. This is 'by design' and should therefore not yield warning in the terminal hinting on a possible problem. Instead, writing to the debugger should suffice.
The second issue this PR tackles is that, after a token expires and a new refresh token is requested from the endpoint, that new token is never stored. This causes the authentication logic to request a new refresh token for every dbt command after initial token has expired (valid for 1h) where it is only needed after it is expired.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/credentials.py
+++ b/dbt/adapters/databricks/credentials.py
@@ -298,7 +298,7 @@ class DatabricksCredentials(Credentials):
                                 )
                             return provider
                     except Exception as e:
-                        # SPA token are suppose to expire after 24h, no need to warn
+                        # SPA token are supposed to expire after 24h, no need to warn
                         if SPA_CLIENT_FIXED_TIME_LIMIT_ERROR in str(e):
                             logger.debug(CredentialLoadError(e))
                         else:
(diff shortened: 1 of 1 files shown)


benc-db (approved):

benc-db:
Re-pushing to run integration tests.

Can you run
```
tox -e linter
```

to see what you need to fix linting issue?

thijs-nijhuis:
@benc-db , sorry about that. Forgot about the linter. I ran it and it fixed a couple of things. I also changed the code slightly to check if the token has changed after it is checked to be valid. Only when changed, it will store the new token using keyring

benc-db:
Going to open a new PR of this to update the changelog and then merge.  Thanks for the contribution!

benc-db:
Closed in favor of 700.

PR #706, opened 2024-06-21
Add support for Serverless jobs / refactor api usage

### LINKED ISSUE #653: Support for new serverless compute
### Describe the feature
Databricks just released the new Serverless Compute at May 1st, I think it will help us a lot on the pipeline, to avoid slow spin-up time of All-purpose/job cluster. So, can we add support for this type of Compute?

### Additional context
https://docs.databricks.com/en/workflows/jobs/run-serverless-jobs.html

### Who will this benefit?
I think all kinds of job can get benefit from this, because we don't need to control the cluster by ourselves.

Yes, but don't know how to do.

benc-db:
Resolves #653

### Description

This PR does 2 things:
1. Adds feature of using serverless compute for python model jobs
2. Refactors API usage to a.) prepare for eventually replacing with WorkspaceClient and b.) make it easier to unit test all of these API interactions.

### Checklist


dbt/adapters/databricks/connections.py
```diff
@@ -477,16 +478,15 @@ class DatabricksConnectionManager(SparkConnectionManager):
     credentials_provider: Optional[TCredentialProvider] = None
     _user_agent = f"dbt-databricks/{__version__}"
 
+    def __init__(self, profile: AdapterRequiredConfig, mp_context: SpawnContext) -> None:
```
benc-db, line 481: Can remove.  I was initially going to use the initializer to make the client, and reuse the client for making pipeline calls, but the pipeline API was complicated enough that I want to hold off until we can replace with an SDK client, especially because the 'open' method is a class method and can't reference a client associated with an instance.

jackyhu-db (approved):

PR #731, opened 2024-07-12
#676 pk constraint

### LINKED ISSUE #676: [bug] Composite primary key constraint fails
### Describe the bug
Unable to apply a multiple column primary key

### Steps To Reproduce
Simple model:
```
select
'a' as col_1,
'b' as col_2
```

adding col_1 as the PK works as expected.
```
models:
  - name: model_a
    config:
      contract:
        enforced: true
    columns:
      - name: col_1
        data_type: string
        constraints:
          - type: not_null
          - type: primary_key
      - name: col_2
        data_type: string
        constraints:
          - type: not_null
```

```
models:
  - name: model_a
    config:
      contract:
        enforced: true
    columns:
      - name: col_1
        data_type: string
        constraints:
          - type: not_null
      - name: col_2
        data_type: string
        constraints:
          - type: not_null
    constraints:
      - type: primary_key
        columns: [col_1, col_2]   
```

However, updating to try and use a composite primary key does not work.  In the second scenario it tries to create the primary key 
>alter table `model_a` add constraint 0a36b673f037eb84f703cefef2531521 primary key(col_1, col_2);

However, it is not setting the two columns as not_null first and so creation of the PK fails.
>: Cannot create the primary key `0a36b673f037eb84f703cefef2531521` because its child column(s) `col_1` is nullable. Please change the column nullability and retry.

### Expected behavior
The columns are set to NOT NULL before the primary key is applied

```
`       create or replace table `model_a`
 
    using delta
      as
    select col_1, col_2
    from (
select
'a' as col_1,
'b' as col_2
    ) as model_subq`

alter table `model_a` add constraint 0a36b673f037eb84f703cefef2531521 primary key(col_1, col_2);
      
18:59:38 Databricks adapter: Cursor(session-id=01ef186d-6818-15d1-b37f-1642bf4c4ff9, command-id=Unknown) - Closing cursor
18:59:38 Databricks adapter: Exception while trying to execute query
/* {"app": "dbt", "dbt_version": "1.7.14", "dbt_databricks_version": "1.7.15", "databricks_sql_connector_version": "2.9.6", "profile_name": "user", "target_name": "dev", "node_id": "model.databricks_analytics_dbt.model_a"} */

        alter table `model_a` add constraint 0a36b673f037eb84f703cefef2531521 primary key(col_1, col_2);
      
: Cannot create the primary key `0a36b673f037eb84f703cefef2531521` because its child column(s) `col_1` is nullable. Please change the column nullability and retry.
```

### System information
```
18:59:18 Running with dbt=1.7.14
18:59:18 Registered adapter: databricks=1.7.15
```

DBT Cloud

frankivo:
Resolves #676 

### Description

This PR makes sure that the column constraints are added before the table constraints.
The order matters for the primary_key constraint, which requires the column to have a not_null constraint. At a later moment, I suppose, the unique constraint should be present too.

I don't think the order matters for other (supported) contraints.

### Checklist

I am not sure how to build and test this yet. Existing unit tests seem to still work.

### DIFF
--- a/dbt/include/databricks/macros/relations/constraints.sql
+++ b/dbt/include/databricks/macros/relations/constraints.sql
@@ -19,8 +19,8 @@
       {# Constraints are not applied for incremental updates. This point in the code should not have been reached #}
       {{ exceptions.raise_compiler_error("Constraints are not applied for incremental updates. Full refresh is required to update constraints.") }}
     {% else %}
-      {% do alter_table_add_constraints(relation, model) %}
       {% do alter_column_set_constraints(relation, model) %}
+      {% do alter_table_add_constraints(relation, model) %}
     {% endif %}
   {% endif %}
 {% endmacro %}
--- a/dbt/include/databricks/macros/relations/constraints.sql
+++ b/dbt/include/databricks/macros/relations/constraints.sql
@@ -79,8 +79,7 @@
 
 {% macro get_constraints_sql(relation, constraints, model, column={}) %}
   {% set statements = [] %}
-  -- Hack so that not null constraints will be applied before primary key constraints
-  {% for constraint in constraints|sort(attribute='type') %}
+  {% for constraint in constraints %}
     {% if constraint %}
       {% set constraint_statements = get_constraint_sql(relation, constraint, model, column) %}
       {% for statement in constraint_statements %}
--- a/dbt/include/databricks/macros/relations/constraints.sql
+++ b/dbt/include/databricks/macros/relations/constraints.sql
@@ -79,7 +79,8 @@
 
 {% macro get_constraints_sql(relation, constraints, model, column={}) %}
   {% set statements = [] %}
-  {% for constraint in constraints %}
+  -- Hack so that not null constraints will be applied before primary key constraints
+  {% for constraint in constraints|sort(attribute='type') %}
     {% if constraint %}
       {% set constraint_statements = get_constraint_sql(relation, constraint, model, column) %}
       {% for statement in constraint_statements %}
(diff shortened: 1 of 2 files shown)


dbt/include/databricks/macros/relations/constraints.sql
```diff
@@ -79,8 +79,7 @@
 
 {% macro get_constraints_sql(relation, constraints, model, column={}) %}
   {% set statements = [] %}
-  -- Hack so that not null constraints will be applied before primary key constraints
-  {% for constraint in constraints|sort(attribute='type') %}
```
benc-db, line 83: And you've tested that we don't need to order them after your change?

benc-db:
I'm going to pull-push to run the integration tests

benc-db:
This fails an integration test:

```
  Cannot create the primary key `475b7994e83a4721f278bf15984ac8d8` because its child column(s) `n` is nullable. Please change the column nullability and retry.
```

dbt/include/databricks/macros/relations/constraints.sql
frankivo: Not quite yet. But: the only two supported constraints are check and not_null. The order in that should not matter.
Will also see if I can test this.

frankivo:
The failing test is the exact thing I'm trying to fix. Will investigate.

frankivo:
> The failing test is the exact thing I'm trying to fix. Will investigate.

It's not actually. It's a FK constraint that is failing now.

dbt/include/databricks/macros/relations/constraints.sql
frankivo: I was wrong to assume this hack was for pk/not_null only. In the meantime I've been able to get the integration-tests working, so it _should_ be fine now.

frankivo:
@benc-db Could you re-run the integration-tests?

benc-db (dismissed):

benc-db:
Assuming you are done, I'm ready to merge.  Thanks again.

frankivo:
> Assuming you are done, I'm ready to merge. Thanks again.

If  you are happy, go ahead :)

PR #739, opened 2024-07-23
Extend Merge Capabilities

### LINKED ISSUE #245: Advanced MERGE incremental settings
### Describe the feature
Currently merge **conditions** are not supported in merge strategy.

If following the Databricks terminology, merge conditions are `matched_conditions` and `not_matched_conditions` in `MERGE` definition.

Taken from DBricks docs:
```sql
MERGE INTO target_table_name [target_alias]
   USING source_table_reference [source_alias]
   ON merge_condition
   { WHEN MATCHED [ AND matched_condition ] THEN matched_action |
     WHEN NOT MATCHED [ AND not_matched_condition ] THEN not_matched_action } [...]
```

As for now the full power of `MERGE` statement is not supported.

Writing customized adapter with own `merge` macro to perform such operations. Add conditions to configuration for model.

### Additional context
Databricks is more and more often used as a tool to support a variety of data operations that was previously relied on some analytical MPPs. Specifically preparing some DWH-like normalized models for long-term entity-oriented storage and creating non-denormalized data marts.

One of the drawbacks of the merge is that update operation is basically DELETE + INSERT, so it creates some additional expenses for each record that you are updating. So it is widely used techniques to minimise such updates by additional conditions which may apply only for already observed records, and thus can't be defined as `merge predicates` (in dbt terminology).

Moreover, there are business cases that require that. For example, I am tracking some object and its properties as an entry of the form `<o, P>`. If I observe some specific instance of such record (`<1, A>`) I may want to skip all consequent updates if the properties are not changing. And do the update only if I see `<1, B>` (for example).

### Who will this benefit?
Anyone who uses MERGE as a strategy for model updates and requires additional business checks for the records before making updates / inserts.

Probably, but due to variety of possible solutions I will appreciate the solution design suggestion to be discussed with maintainers.

### LINKED ISSUE #707: Support for `MERGE WITH SCHEMA EVOLUTION`
### Describe the feature
Add support for the new `MERGE WITH SCHEMA EVOLUTION` syntax available on the current and preview channel SQL Warehouse versions. https://docs.databricks.com/en/delta/update-schema.html#schema-evolution-syntax-for-sql.

This is especially useful if you need schema evolution on struct data types.

The change is quite easy SQL wise, but I don't know how you would want to expose the option... via `on_schema_change`?

mi-volodin:
Resolves #245 
Resolves #707 

### Motivation

The current MERGE INTO capabilities available in dbt (data build tool) significantly differ from what can be achieved through a pure SQL interface. While it is sometimes possible to circumvent these limitations using complex SQL techniques, this approach often increases code complexity. Additionally, because Spark does not execute some operations identically when comparing direct merges with workaround solutions, this can lead to performance issues.

### Description

This PR aims to add the following capabilities:

- Add flag `skip_matched_step` to be able to skip matched step in merge statement. This and other mentioned flags are considered active only if they equal to a `true` string (case-insensitive).
- Add flag `skip_not_matched_step` to be able to skip not matched step in merge statement.
- Add flag `merge_with_schema_evolution` to allow new capability of MERGE to evolve schema.
- Add `matched_conditions` and `not_matched_conditions`:
    - The default conditions can be expressed as a string only, as it looks more convenient for complex conditions (mix of `and` and `or`, functions)
    - Default aliases for source and target are `src` and `tgt` respectively, but...
    - it is possible to set aliases using `target_alias` or `source_alias` parameters.
    - Rationale and background:  see [discussion](https://github.com/databricks/dbt-databricks/issues/245#issuecomment-2168416491)
- Add `delete` option for merge. This will be triggered by two parameters: `not_matched_by_source_action` which should be set to `delete`. Also it is possible to add `not_matched_by_source_condition` which defines additional predicates that should resolve to `true` for delete to happen.

#### Tests

- [x] Model that has updates with `skip_matched_step == true` updates nothing.
- [x] Model that has new records with `skip_not_matched_step == true` inserts nothing.
- [x] Model with extended schema adds the new attribute when used in the input model. When attribute is not provided in the old records - it remains `null`.
- [x] Test for `matched_conditions` and `not_matched_conditions` that covers all cases of logic (matched / not + conditions met / not) + also cover custom alias testing.
- [x] Test for `not_matched_by_source` removes the record that is not on source, keeps the record which is not on source but the `not_matched_by_source_condition` is not met.

#### Open questions

- [x] Do I need to implement exceptions for compile time for the case I have some inconsistent flag states (e.g. all steps in merge are requested to be skipped)?
  - Decision is NO, as Databricks should cover syntax errors reporting.

#### Docs

- [x] Update documentation with the new capabilities (see added `docs/databricks-merge.md`
- [x] Changelog

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
 - Add 'user_folder_for_python' config to switch writing python model notebooks to the user's folder ([706](https://github.com/databricks/dbt-databricks/pull/706))
 - Merge capabilities are extended ([739](https://github.com/databricks/dbt-databricks/pull/739)) to include the support for the follwing features:
   - `with schema evolution` clause (requires Databricks Runtime 15.2 or above);
-  - `when not matched on source` clause, only for `delete` action
-  - `matched`, `not matched` and `not matched on source` condition clauses;
+  - `when not matched by source` clause, only for `delete` action
+  - `matched`, `not matched` and `not matched by source` condition clauses;
   - custom aliases for source and target tables can be specified and used in condition clauses;
   - `matched` and `not matched` steps can now be skipped;
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 - Add support for serverless job clusters on python models ([706](https://github.com/databricks/dbt-databricks/pull/706))
 - Add 'user_folder_for_python' config to switch writing python model notebooks to the user's folder ([706](https://github.com/databricks/dbt-databricks/pull/706))
+- Merge capabilities are extended ([739](https://github.com/databricks/dbt-databricks/pull/739)) to include the support for the follwing features:
+  - `with schema evolution` clause (requires Databricks Runtime 15.2 or above);
+  - `when not matched on source` clause, only for `delete` action
+  - `matched`, `not matched` and `not matched on source` condition clauses;
+  - custom aliases for source and target tables can be specified and used in condition clauses;
+  - `matched` and `not matched` steps can now be skipped;
 
 ### Under the Hood
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Add support for serverless job clusters on python models ([706](https://github.com/databricks/dbt-databricks/pull/706))
 - Add 'user_folder_for_python' config to switch writing python model notebooks to the user's folder ([706](https://github.com/databricks/dbt-databricks/pull/706))
-- Merge capabilities are extended ([739](https://github.com/databricks/dbt-databricks/pull/739)) to include the support for the follwing features:
+- Merge capabilities are extended ([739](https://github.com/databricks/dbt-databricks/pull/739)) to include the support for the following features (thanks @mi-volodin):
   - `with schema evolution` clause (requires Databricks Runtime 15.2 or above);
   - `when not matched by source` clause, only for `delete` action
   - `matched`, `not matched` and `not matched by source` condition clauses;
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -101,9 +101,13 @@ class DatabricksConfig(AdapterConfig):
     buckets: Optional[int] = None
     options: Optional[Dict[str, str]] = None
     merge_update_columns: Optional[str] = None
+    merge_exclude_columns: Optional[str] = None
     databricks_tags: Optional[Dict[str, str]] = None
     tblproperties: Optional[Dict[str, str]] = None
     zorder: Optional[Union[List[str], str]] = None
+    skip_matched_step: Optional[bool] = None
+    skip_non_matched_step: Optional[bool] = None
+    merge_with_schema_evolution: Optional[bool] = None
 
 
 def check_not_found_error(errmsg: str) -> bool:
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -105,8 +105,14 @@ class DatabricksConfig(AdapterConfig):
     databricks_tags: Optional[Dict[str, str]] = None
     tblproperties: Optional[Dict[str, str]] = None
     zorder: Optional[Union[List[str], str]] = None
-    skip_matched_step: Optional[bool] = None
     skip_non_matched_step: Optional[bool] = None
+    skip_matched_step: Optional[bool] = None
+    matched_condition: Optional[str] = None
+    not_matched_condition: Optional[str] = None
+    not_matched_by_source_action: Optional[str] = None
+    not_matched_by_source_condition: Optional[str] = None
+    target_alias: Optional[str] = None
+    source_alias: Optional[str] = None
     merge_with_schema_evolution: Optional[bool] = None
 
 
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -90,7 +90,7 @@ select {{source_cols_csv}} from {{ source_relation }}
   {%- set not_matched_condition = config.get('not_matched_condition') -%}
 
   {%- set not_matched_by_source_action = config.get('not_matched_by_source_action') -%}
-  {%- set not_matched_by_source_condition = config.get('not_matchednot_matched_by_source_condition_condition') -%}
+  {%- set not_matched_by_source_condition = config.get('not_matched_by_source_condition') -%}
   
 
   {% if unique_key %}
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -86,8 +86,11 @@ select {{source_cols_csv}} from {{ source_relation }}
   {%- set skip_matched_step = (config.get('skip_matched_step') | lower == 'true') -%}
   {%- set skip_not_matched_step = (config.get('skip_not_matched_step') | lower == 'true') -%}
 
-  {%- set matched_conditions = config.get('matched_conditions') -%}
-  {%- set not_matched_conditions = config.get('not_matched_conditions') -%}
+  {%- set matched_condition = config.get('matched_condition') -%}
+  {%- set not_matched_condition = config.get('not_matched_condition') -%}
+
+  {%- set not_matched_by_source_action = config.get('not_matched_by_source_action') -%}
+  {%- set not_matched_by_source_condition = config.get('not_matchednot_matched_by_source_condition_condition') -%}
   
 
   {% if unique_key %}
@@ -120,19 +123,26 @@ select {{source_cols_csv}} from {{ source_relation }}
         {{ predicates | join('\n    and ') }}
     {%- if not skip_matched_step %}
     when matched
-        {%- if matched_conditions %}
-        and ({{ matched_conditions }})
+        {%- if matched_condition %}
+        and ({{ matched_condition }})
         {%- endif %}
         then update set
             {{ get_merge_update_set(update_columns, on_schema_change, source_columns, source_alias) }}
     {%- endif %}
     {%- if not skip_not_matched_step %}
     when not matched
-        {%- if not_matched_conditions %}
-        and ({{ not_matched_conditions }})
+        {%- if not_matched_condition %}
+        and ({{ not_matched_condition }})
         {%- endif %}
         then insert
             {{ get_merge_insert(on_schema_change, source_columns, source_alias) }}
+    {%- endif %}
+    {%- if not_matched_by_source_action == 'delete' %}
+    when not matched by source
+        {%- if not_matched_by_source_condition %}
+        and ({{ not_matched_by_source_condition }})
+        {%- endif %}
+        then delete
     {%- endif %}
 {% endmacro %}
 
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -146,7 +146,7 @@ select {{source_cols_csv}} from {{ source_relation }}
     {%- endif %}
 {% endmacro %}
 
-{% macro get_merge_update_set(update_columns, on_schema_change, source_columns, source_alias) %}
+{% macro get_merge_update_set(update_columns, on_schema_change, source_columns, source_alias='src') %}
   {%- if update_columns -%}
     {%- for column_name in update_columns -%}
       {{ column_name }} = {{ source_alias }}.{{ column_name }}{%- if not loop.last %}, {% endif -%}
@@ -160,7 +160,7 @@ select {{source_cols_csv}} from {{ source_relation }}
   {%- endif -%}
 {% endmacro %}
 
-{% macro get_merge_insert(on_schema_change, source_columns, source_alias) %}
+{% macro get_merge_insert(on_schema_change, source_columns, source_alias='src') %}
   {%- if on_schema_change == 'ignore' -%}
     *
   {%- else -%}
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -71,6 +71,10 @@ select {{source_cols_csv}} from {{ source_relation }}
 
 {% macro databricks__get_merge_sql(target, source, unique_key, dest_columns, incremental_predicates) %}
   {# need dest_columns for merge_exclude_columns, default to use "*" #}
+
+  {%- set target_alias = config.get('target_alias', 'tgt') -%}
+  {%- set source_alias = config.get('source_alias', 'src') -%}
+
   {%- set predicates = [] if incremental_predicates is none else [] + incremental_predicates -%}
   {%- set dest_columns = adapter.get_columns_in_relation(target) -%}
   {%- set source_columns = (adapter.get_columns_in_relation(source) | map(attribute='quoted') | list)-%}
@@ -82,17 +86,21 @@ select {{source_cols_csv}} from {{ source_relation }}
   {%- set skip_matched_step = (config.get('skip_matched_step') | lower == 'true') -%}
   {%- set skip_not_matched_step = (config.get('skip_not_matched_step') | lower == 'true') -%}
 
+  {%- set matched_conditions = config.get('matched_conditions') -%}
+  {%- set not_matched_conditions = config.get('not_matched_conditions') -%}
+  
+
   {% if unique_key %}
       {% if unique_key is sequence and unique_key is not mapping and unique_key is not string %}
           {% for key in unique_key %}
               {% set this_key_match %}
-                  DBT_INTERNAL_SOURCE.{{ key }} <=> DBT_INTERNAL_DEST.{{ key }}
+                  {{ source_alias }}.{{ key }} <=> {{ target_alias }}.{{ key }}
               {% endset %}
               {% do predicates.append(this_key_match) %}
           {% endfor %}
       {% else %}
           {% set unique_key_match %}
-              DBT_INTERNAL_SOURCE.{{ unique_key }} <=> DBT_INTERNAL_DEST.{{ unique_key }}
+              {{ source_alias }}.{{ unique_key }} <=> {{ target_alias }}.{{ unique_key }}
           {% endset %}
           {% do predicates.append(unique_key_match) %}
       {% endif %}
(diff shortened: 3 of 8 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -100,9 +100,13 @@ class DatabricksConfig(AdapterConfig):
     buckets: Optional[int] = None
     options: Optional[Dict[str, str]] = None
     merge_update_columns: Optional[str] = None
+    merge_exclude_columns: Optional[str] = None
```
mi-volodin, line 103: Used in macro, but not declared. Not sure if needs to be added.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -97,11 +100,26 @@ select {{source_cols_csv}} from {{ source_relation }}
...
+        {%- endif %}
+    into
+        {{ target }} as DBT_INTERNAL_DEST
+    using
+        {{ source }} as DBT_INTERNAL_SOURCE
+    on
+        {{ predicates | join('\n    and ') }}
+    {%- if not skip_matched_step %}
+    when matched
+        then update set
```
mi-volodin, line 115: Sorry for weird formatting, I cleaned up my experiments with match conditions, they will reside here between these two lines.

mi-volodin:
❗ My local tests are failing mostly on the snapshot steps. And the reason for it concealed in dbt base tests, which I found buggy.

It works like this:

1. First run we generate the snapshot from scratch, populating the `dbt_` fields. Also `dbt_unique_id`.
2. Next run when we update snapshot `dbt` core `snapshot` logic (see `dbt/include/global_project/macros/materializations/snapshots/helpers.sql`) executes the following 
  
    ```sql
    snapshotted_data as (

        select *,
            {{ strategy.unique_key }} as dbt_unique_key

        from {{ target_relation }}
        where dbt_valid_to is null

    ),
    ```

3. If we look in the tests definition of dbt, the `unique_key` field will be `id` of a seed. (l.123 `dbt/tests/adapter/basic/files.py`). And therefore the following sql is generated during tests

    ```sql
    snapshotted_data as (

        select *,
            id as dbt_unique_key

        from [EXISTING_SNAPSHOT]
        where dbt_valid_to is null

    ),
    ```

It cannot be resolved, because `snapshotted_data` now has duplicated `dbt_unique_key` attribute, and when it is referenced below - query fails.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
benc-db: Formatting with Jinja will typically be weird, because you have to go out of your way to get spacing to not look terrible.  I wouldn't worry too much about it.

benc-db:
> ❗ My local tests are failing mostly on the snapshot steps. And the reason for it concealed in dbt base tests, which I found buggy.
> 
> It works like this:
> 
> 1. First run we generate the snapshot from scratch, populating the `dbt_` fields. Also `dbt_unique_id`.
> 2. Next run when we update snapshot `dbt` core `snapshot` logic (see `dbt/include/global_project/macros/materializations/snapshots/helpers.sql`) executes the following
>    ```sql
>    snapshotted_data as (
>    
>        select *,
>            {{ strategy.unique_key }} as dbt_unique_key
>    
>        from {{ target_relation }}
>        where dbt_valid_to is null
>    
>    ),
>    ```
> 3. If we look in the tests definition of dbt, the `unique_key` field will be `id` of a seed. (l.123 `dbt/tests/adapter/basic/files.py`). And therefore the following sql is generated during tests
>    ```sql
>    snapshotted_data as (
>    
>        select *,
>            id as dbt_unique_key
>    
>        from [EXISTING_SNAPSHOT]
>        where dbt_valid_to is null
>    
>    ),
>    ```
> 
> It cannot be resolved, because `snapshotted_data` now has duplicated `dbt_unique_key` attribute, and when it is referenced below - query fails.

Can we clarify, is the test buggy, or our implementation?  Is this something we should raise to dbt Labs?

benc-db:
@mi-volodin if it's not too painful, can you rebase against 1.9.latest branch next week?  This feature (and a couple of others I've worked on) is big enough to increment feature version.  I'll get 1.9.latest updated today to include latest changes from main.

mi-volodin:
@benc-db sure thing. TBH never done such operation before, so would be curious how painful is this 😄 

Let me know when I can start. I see neither tag nor branch with 1.9... Maybe I am looking in a wrong place. Anyway, I'd appreciate if you ping me once the time comes.

mi-volodin:
> Can we clarify, is the test buggy, or our implementation? Is this something we should raise to dbt Labs?

For now it looks purely as dbt side bug. Which is suspicuous... In `dbt-databricks` both the snapshot code and the tests are taken from `dbt-core`. So let me finish with the development and I will finalise the investigation.

I planned to raise it in dbt-core, wrote this note here just to explain that it is not related to my changes.

mi-volodin:
@benc-db rebased

mi-volodin:
@benc-db sorry, rebased, but not tested yet. Let me fix some issues and finalise the tests before running GH actions.

benc-db:
@mi-volodin just ran the unit tests; these are super light weight, just wanted to see where we were at.  No worries :)

mi-volodin:
@benc-db Now I am done with code and testing. Before switching to the changelog and documentation - there's an open question regarding pre-checking certain conditions on compile time.

For instance, I can encode certain "impossible" cases to be checked at compile time. And if failed - throw an exception.

I am asking, because I am not sure it is really needed. Basically I will be able to put some simple checks like having at least matched/not matched/not matched on source step, but definitely not everything that possibly can be wrong.

What do you think?

benc-db:
>For instance, I can encode certain "impossible" cases to be checked at compile time. And if failed - throw an exception.  I am asking, because I am not sure it is really needed. Basically I will be able to put some simple checks like having at least matched/not matched/not matched on source step, but definitely not everything that possibly can be wrong.

This is a really good question.  In general, we allow Databricks to tell the user what went wrong; the exception is if its very easy to misconfigure something, or if the Databricks exception is misleading.  So, in those impossible cases, do you think its easy to read the error and figure out what you need to change to fix it? If not, it would be a good idea to raise compilation errors that tell the users what they need to do to configure things correctly.

mi-volodin:
Yes, in my cases the syntax error will be raised. And Databricks will point that out. I think then I will skip these exceptions for now.

I have one more stupid question: I cannot find any trace of [documentation ](https://docs.getdbt.com/reference/resource-configs/databricks-configs#incremental-models) in that repo. Is it tracked separately?

benc-db:
>I have one more stupid question: I cannot find any trace of [documentation ](https://docs.getdbt.com/reference/resource-configs/databricks-configs#incremental-models)in that repo. Is it tracked separately?

Yes, I submit a PR to dbt for that part.  If you include the doc locally, I'll make the PR with 1.9 to get the doc updated on the site.

mi-volodin:
@benc-db By including locally, do you mean writing an MD file with explanations and put in in the `./docs` folder?

I also can create a PR for updating the documentation in dbt-core repo, no problem. Just wanted to know what would be the most convenient way.

benc-db:
> @benc-db By including locally, do you mean writing an MD file with explanations and put in in the ./docs folder?

Yes, exactly.  I already will need to submit a doc update for 1.9 anyway, so I'll fold in the doc you put into the docs folder.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 - Add support for serverless job clusters on python models ([706](https://github.com/databricks/dbt-databricks/pull/706))
 - Add 'user_folder_for_python' config to switch writing python model notebooks to the user's folder ([706](https://github.com/databricks/dbt-databricks/pull/706))
+- Merge capabilities are extended ([739](https://github.com/databricks/dbt-databricks/pull/739)) to include the support for the follwing features:
+  - `with schema evolution` clause (requires Databricks Runtime 15.2 or above);
+  - `when not matched on source` clause, only for `delete` action
+  - `matched`, `not matched` and `not matched on source` condition clauses;
```
mi-volodin, line 10:
```suggestion
  - `when not matched by source` clause, only for `delete` action
  - `matched`, `not matched` and `not matched by source` condition clauses;
```

docs/databricks-merge.md
```diff
@@ -0,0 +1,96 @@
...
+If a `unique_key` is not specified, dbt will forgo match criteria and simply insert all new records (similar to `append` strategy).
+
+Specifying `merge` as the incremental strategy is optional since it's the default strategy used when none is specified.
+
+From v.1.9 onwards `merge` behavior can be tuned by setting the additional parameters.
+
+- Merge steps control parameters that tweak the default behaviour:
+  - `skip_matched_step`: if set to `true`, dbt will completely skip the `matched` clause of the merge statement.
+  - `skip_not_matched_step`: similarly if `true` the `not matched` clause will be skipped.
+  - `not_matched_on_source_action`: if set to `delete` the corresponding `when not matched on source ... then delete` clause will be added to the merge statement.
```
mi-volodin, line 20:
```suggestion
  - `not_matched_by_source_action`: if set to `delete` the corresponding `when not matched by source ... then delete` clause will be added to the merge statement.
```

tests/functional/adapter/incremental/fixtures.py
```diff
@@ -279,6 +309,140 @@
...
+skip_not_matched_model = skip_matched_model.replace(
+    "skip_matched_step = true", "skip_not_matched_step = true"
+)
+
+matching_condition_model = """
+{{ config(
+    materialized = 'incremental',
+    unique_key = 'id',
+    incremental_strategy='merge',
+    target_alias='t',
```
mi-volodin, line 350: Only `target_alias` is covered by the test, while actually we should test also `source_alias`

mi-volodin:
@benc-db let me know if I can do anything else to support the process of testing.

Meanwhile I can investigate and report the snapshot test issue.

benc-db:
Running functional tests today, then will start reviewing.  Thanks for your efforts!

mi-volodin:
@benc-db I see some tests are failing due to hardcoded "DBT_INTERNAL_*" aliases. I will take care of it. Apologies that I forgot to look into it beforehand... had a thought, but something distracted me.

mi-volodin:
- [ ] `tox -e linter`
- [ ] `tox -e unit`
- [ ] `tox -e integration_databricks_cluster:
  - [ ] `TestPersistDocsWithSeeds` **failed** with error 
     ```bash
     [DELTA_CANNOT_CREATE_LOG_PATH] Cannot create s3a://[redacted]-warehouse/mnt/dbt_databricks/seeds/persist_seed/_delta_log
     ```
     Expected, as this path does not exist. Probably should work for external testing.
  - [ ] `TestSnapshot.test_inserts_are_captured_by_snapshot` **failed** with ambigous reference error (see [above](https://github.com/databricks/dbt-databricks/pull/739#issuecomment-2252939914))
  - [ ] `TestSnapshot.test_deletes_are_captured_by_snapshot` same
  - [ ] `TestSnapshot.test_new_column_captured_by_snapshot` same
  - [ ] `TestIncrementalOnSchemaChange.test_run_incremental_ignore` failed with the following error
     ```bash
     [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter with name `field3` cannot be resolved. Did you mean one of the following? [`field1`, `field2`, `id`]. SQLSTATE: 42703; line 7 pos 40
     ```
     ❗  Looking into it.
  - [ ] `TestSnapshotCheckCols.test_snapshot_check_cols` same snapshot error
  - [ ] `TestSnapshotTimestamp.test_snapshot_timestamp` same snapshot error
  - [ ]  `TestSnapshot.test_revives_are_captured_by_snapshot` ❗  hardcoded placeholders again  + same snapshot error
  - [ ] `TestWarehousePerModel.test_wpm` ??? (probably output mixed with previous test)

mi-volodin:
@benc-db I fixed unit tests. Among tests that are failing I looked through - majority is my Snapshot issue. Would be interested to know if it also fails on your side.

There are two more tests that are failing. First one, as described above, is supposed to test "ignore_schema" behaviour. But, I think it doesn't work by design if the schema evolution is enabled by default in the environment. I.e. in that case the merge operation that happens execute only inserts, and that inserts extend the schema to have `field3, field4`. 

Later diff test generates the code to compare it against "target" which has only `field1,2` and expectedly fails while querying for `field3,4`. 

If you confirm that it should be fixed - I can fix it withing this PR.


The last test is `test_wpm` which fails, because I think it shouldn't be included in my test-set at all (I don't have warehouses and UC). Also can fix it here (exclude).

benc-db:
Apologies for delay, I'm on-call and haven't had much time to review PRs.  Will get to this shortly.

benc-db:
Rerunning functional tests

benc-db:
All functional tests pass.  Just need to find time to actually review the code :P .  Thanks so much!

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 - Add support for serverless job clusters on python models ([706](https://github.com/databricks/dbt-databricks/pull/706))
 - Add 'user_folder_for_python' config to switch writing python model notebooks to the user's folder ([706](https://github.com/databricks/dbt-databricks/pull/706))
+- Merge capabilities are extended ([739](https://github.com/databricks/dbt-databricks/pull/739)) to include the support for the follwing features:
```
benc-db, line 7: Can you add a thank you call out (search 'thanks for examples) to give yourself credit for this work :)?

dbt/adapters/databricks/impl.py
```diff
@@ -101,9 +101,19 @@ class DatabricksConfig(AdapterConfig):
...
     databricks_tags: Optional[Dict[str, str]] = None
     tblproperties: Optional[Dict[str, str]] = None
     zorder: Optional[Union[List[str], str]] = None
+    skip_non_matched_step: Optional[bool] = None
+    skip_matched_step: Optional[bool] = None
+    matched_condition: Optional[str] = None
+    not_matched_condition: Optional[str] = None
+    not_matched_by_source_action: Optional[str] = None
+    not_matched_by_source_condition: Optional[str] = None
+    target_alias: Optional[str] = None
```
benc-db, line 114: I wonder if these should be nested in a 'merge_options' config?  Mostly thinking about this because target_alias and source_alias might have meaning in other scenarios, but we want users to know that this is a mechanism for tailoring the merge SQL.

dbt/adapters/databricks/impl.py
```diff
@@ -101,9 +101,19 @@ class DatabricksConfig(AdapterConfig):
...
+    source_alias: Optional[str] = None
+    merge_with_schema_evolution: Optional[bool] = None
```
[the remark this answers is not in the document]
benc-db: The way python models work, I would expect these to apply there too, as python models are created by first making a table in python from a DataFrame, and then running SQL against that table to do things like incremental merges.

benc-db (commented):
See comment about having merge_options as a top level config that the other configs nest under (similar to what I do for schedule, which has two possible components).  Otherwise, looks good.

dbt/adapters/databricks/impl.py
mi-volodin: That's possible, but we have to possibly do the same with some of the old configuration parameters.

So right now it looks like this

```bash
└── config
    ├── target_alias
    ├── source_alias
    ├── predicates
    ├── dest_columns
    ├── source_columns
    ├── merge_update_columns
    ├── merge_exclude_columns
    ├── merge_with_schema_evolution
    ├── on_schema_change
    ├── update_columns
    ├── skip_matched_step
    ├── skip_not_matched_step
    ├── matched_condition
    ├── not_matched_condition
    ├── not_matched_by_source_action
    └── not_matched_by_source_condition
```

If I move only added options to the `merge_options` subdict it will look like this

```bash
└── config
    ├── predicates
    ├── dest_columns
    ├── source_columns
    ├── merge_update_columns
    ├── merge_exclude_columns
    ├── on_schema_change
    ├── update_columns
    └── merge_options/
        ├── target_alias
        ├── source_alias
        ├── merge_with_schema_evolution
        ├── skip_matched_step
        ├── skip_not_matched_step
        ├── matched_condition
        ├── not_matched_condition
        ├── not_matched_by_source_action
        └── not_matched_by_source_condition
```

where at least for me the rest of the options are misplaced, as they are also `merge_options`.

## Options

Though I don't mind to change the code, I think it is a quite important design solution. So let me outline my opinion for the options that I see here.

With a **flat config map** there's a hard requirement to have all options uniquely identified to exclude misinterpretation. Also the documentation will be absolutely required.
However, If some config parameter will be misused on `dbt-databricks` development side - it will likely fail the tests. If not - it should be safe to use it.
Same time flat map simplifies the structure and kind of make it more straightforward for users, especially vs the case when some parameters should be nested and some not. Also, if some parameters should be used in several contexts - it is okay to use existing config key.

With a **merge options subdict** we group some of the options highlighting the usage context. But, if some identical parameter (we are using SQL, where many clauses are also coincide between statements) should be used for a different context - we will have to create another subdict or introduce a root `config` key. 
Also, I think in this case we should migrate all existing parameters to `merge_options` as well, which will create the need for deprecation process.

So in the context of the current state of the `config` and proposed extension I perceive the variant with `merge_options` subdict as an unnecessary complexity. 

Of course, I will fix everything to the state that you think is better for the future of the product.

mi-volodin: There's a middle way. We can keep flat map, but highlight some critical config keys that they are exclusively for the merge context. By adding a `merge_` prefix, for instance.

```bash
└── config
    ├── merge_target_alias
    ├── merge_source_alias
    ├── predicates
    ├── dest_columns
    ├── merge_source_columns
    ├── merge_update_columns
    ├── merge_exclude_columns
   ...
```

Bad thing is that still some old parameters seem to have identical problem (`predicates`, `dest_columns`) and if we have to deprecate them in favor of a new name - then we can probably do this later 😄

dbt/adapters/databricks/impl.py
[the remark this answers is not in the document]
mi-volodin: Hm... should I add some python model test? I have no idea if it will work now.

CHANGELOG.md
mi-volodin:
```suggestion
- Merge capabilities are extended ([739](https://github.com/databricks/dbt-databricks/pull/739)) to include the support for the following features (thanks @mi-volodin):
```

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
...
+  - `with schema evolution` clause (requires Databricks Runtime 15.2 or above);
+  - `when not matched by source` clause, only for `delete` action
```
mi-volodin, line 9:
```suggestion
  - `when not matched by source` clause, only for `delete` action;
```

dbt/adapters/databricks/impl.py
benc-db: So, you can't see it here, because I'm just reading it out of extras, but I'm already using nesting; I'll accept flat for now, and we'll see if any conflicts come up later.

dbt/adapters/databricks/impl.py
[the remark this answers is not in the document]
benc-db: I think it would be good to add a python test.  Skip for databricks_uc_sql_endpoint and databricks_cluster.  It should just work, because the merge is between SQL tables, even when using a python model.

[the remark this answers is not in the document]
mi-volodin: Ok, I'll look into it

benc-db:
@mi-volodin thank you for your hard work.  I will be going on vacation this week, and will start working on the 1.9 release when I return in September.  Just wanted to let you know.

mi-volodin:
@benc-db thanks! And I have just returned from mine 😄 Very happy to see it merged, and I still plan to look into python models in tests.
Also please loop me in if any issues are be uncovered. Have a nice vacation!

benc-db:
@mi-volodin I'm going to be opening a new PR to make the default match the prior aliases.  In beta testing, some users have macro overrides that assume the previous naming, leading to failures.

PR #742, opened 2024-07-24
Fix capitalization issue

benc-db:
### Description

Incremental tables were potentially not running incrementally if components were specified with capitalization, as in information_schema they are always present in lowercase.  Talking to dbt to see if there is a better way to do this, but for now, ensure lower case when used in queries.

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/adapters/catalog.sql
+++ b/dbt/include/databricks/macros/adapters/catalog.sql
@@ -45,7 +45,7 @@
         'The timestamp for last update/change' as `stats:last_modified:description`,
         (last_altered is not null and table_type not ilike '%VIEW%') as `stats:last_modified:include`
     from `system`.`information_schema`.`tables`
-    where table_catalog = '{{ information_schema.database }}'
+    where table_catalog = '{{ information_schema.database|lower }}'
 {%- endmacro %}
 
 {% macro databricks__get_catalog_columns_sql(information_schema) -%}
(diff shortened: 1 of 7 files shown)


dbt/include/databricks/macros/adapters/catalog.sql
```diff
@@ -69,23 +69,23 @@
...
 {% macro databricks__get_catalog_relations_where_clause_sql(catalog, relations) -%}
-    where table_catalog = '{{ catalog }}' and (
+    where table_catalog = '{{ catalog|lower }}' and (
         {%- for relation in relations -%}
             {% if relation.schema and relation.identifier %}
                 (
-                    table_schema = lower('{{ relation.schema }}')
-                    and table_name = lower('{{ relation.identifier }}')
+                    table_schema = '{{ relation.schema|lower }}'
+                    and table_name = '{{ relation.identifier|lower }}'
```
jackyhu-db, line 84: what if the table name is configured as case-sensitive

benc-db: Can you link to ref?  This would be a considerable monkey wrench if so.

jackyhu-db: nvm. Just found Databricks table name is always case insenstive.

jackyhu-db (dismissed):

benc-db:
Going to merge (after changelog) and if dbt has a better option, will do that as a separate PR.

PR #750, opened 2024-07-30
Fix table comment updates

### LINKED ISSUE #732: Relation comments are not being updated for incremental models
### Describe the bug
Relation comments for incremental models are only persisted when the relation is first being created. On subsequent incremental runs an updated model description will not be persisted to the relation.

### Steps To Reproduce
1) Create an incremental model with:

`model.sql`
```sql
{{
    config(
        materialized="incremental",
        persist_docs={"relation": true},
    )
}}
select 1 as id
```
`schema.yml`
```yml
version: 2
models:
  - name: model
    description: "Model Description"
```
2) `dbt run`
3) Change the model description
4) `dbt run`
5) Check the table comment: `describe detail model`. The table still has the old model description as table comment.

### Expected behavior
After the second `dbt run` the new model description should be persisted as the table comment.

### System information
```
Core:
  - installed: 1.8.3
  - latest:    1.8.3 - Up to date!

Plugins:
  - databricks: 1.8.3 - Up to date!
```

Ubuntu 22.04

Python 3.10.12

### Additional context
I'm interested in contributing to this.

henlue:
Resolves #732

### Description

 - Update table comments for incremental models on incremental runs
 - Update table comments for snapshots on incremental runs
 - Create table comments during dbt clone
 - Check for `config.persist_relation_docs()` in `persist_docs`
 - Align black versions in `black.ini` and `.pre-commit-config.yaml` to fix failing `pre-commit` runs

The issue is only about the incremental models. While working on the PR I realized table comments are also not being updated for snapshots and are not created with `dbt clone`.

I think comments are also not being updated in materialized views and streaming tables. Since I don't have a good understanding of those materializations, I kept the original behavior by setting `for_relation=False`. The tests would also run through with `for_relation=True`, but this might lead to unnecessary queries that are being send to databricks on the first run of those relations.

Generally, the code would be simpler if relation comments are only persisted in `persist_docs` (using `comment on ...` queries) and not as part of `create table` statements. I assume they are persisted as part of the `create` statement to improve performance?

I've ran:
`tox -e unit`
`tox -e integration-databricks-uc-sql-endpoint`
`tox -e integration-databricks-uc-cluster`

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       - id: check-yaml
       - id: check-added-large-files
   - repo: https://github.com/psf/black-pre-commit-mirror
-    rev: 22.12.0
+    rev: 24.3.0
     hooks:
       - id: black
         language_version: python3.11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Alter column statements are now done before the alter table statement (thanks @frankivo!). ([731](https://github.com/databricks/dbt-databricks/pull/731))
 - Always use lower case when gathering metadata (since objects are stored internally as lower case regardless of how we create them) ([742](https://github.com/databricks/dbt-databricks/pull/742))
 - Persist table comments for python models ([743](https://github.com/databricks/dbt-databricks/pull/743))
+- Persist table comments for incremental models, snapshots and dbt clone ([750](https://github.com/databricks/dbt-databricks/pull/750))
 - Stop cursor destructor warnings ([744](https://github.com/databricks/dbt-databricks/pull/744))
 - Race condition on cluster creation. (thanks @jurasan!) ([751](https://github.com/databricks/dbt-databricks/pull/751))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
-## dbt-databricks next
+## dbt-databricks 1.8.6 (TBD)
+
+### Fixes
+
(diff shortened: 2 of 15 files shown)


benc-db:
> I think comments are also not being updated in materialized views and streaming tables. Since I don't have a good understanding of those materializations, I kept the original behavior by setting for_relation=False. The tests would also run through with for_relation=True, but this might lead to unnecessary queries that are being send to databricks on the first run of those relations.

They should already for MV/ST, but the mechanism is very different.  Look at alter.sql for either.  Core reason is that for MV/ST we have a mechanism for declaring how we want to handle changes (on_config_change), but I haven't convinced dbt-core yet that we should have the same mechanism for incremental tables.

benc-db:
Basically, in the absence of on_config_change for other materialization types, you are assuming users want apply.  I think that is the correct assumption.  Thanks for PR, will review in the next day or so.

benc-db:
Running functional tests now.

benc-db:
Please take a look at the non-uc cluster tests; there are few that are failing.

benc-db:
Rerunning functional tests

henlue:
I forgot to sign my latest commit and had to force push it again.

benc-db:
> I forgot to sign my latest commit and had to force push it again now.

Apologies for delay, I'm on-call and haven't had much time to review PRs.  Will get to this shortly.

benc-db:
Does this by any chance fix: https://github.com/databricks/dbt-databricks/issues/763?

benc-db (dismissed):

benc-db (approved):

henlue:
> Does this by any chance fix: #763?

No that issue still persists

PR #751, opened 2024-07-31
fix: race condition on cluster creation

### LINKED ISSUE #347: Python Clusters not starting correctly: race conditions?
### Describe the bug
Hi, I'm experiencing an issue when running python models from terminated clusters.  Feels related to https://github.com/databricks/dbt-databricks/pull/306#issuecomment-1520405748 and #306 in general.  Tagging  for visibility since you've responded on the other thread.

I use python models in conjunction with SQL models (no surprise there) but I configure each python model to have its their own `http_path` so that I can run the python models in conjunction with the SQL models from a single `dbt run --target <sql_warehouse>` command.  So the top level "run" points at SQL and each python model points at python.

Some of the python models run, others experience an error like:
```
{'error': 'ClusterNotReadyException: Cluster <cluster_id> not currently ready for driver client (currently Pending)'}
```

### Steps To Reproduce
Not 100% sure, but I would say:

- Run `dbt run` against a SQL warehouse with python-model level `http_path`s pointing to an all purpose cluster.
- Ensure some SQL models run first.
- Ensure 5+ python models all get hit in the dag around the same time (within the window of terminated -> running, say 2 minutes).

### Expected behavior
I expect for the all purpose cluster to boot when it encounters the first python model and then for all other simultaneous models to realize its booting and wait until the cluster is ready.

See bug description.

### System information
core is 1.5.0
spark is 1.5.0
databricks is 1.5.0

Linux

3.10.4

### Additional context

All the python models point to the same `http_path`.  This feels like a recipe for the race condition @susodapop described in #306 where by:

1.  the first python model to run gets a `TERMINATED` cluster and then runs [`start_cluster`](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/python_submissions.py#LL262C9-L262C22) as intended, then
2. the second one sees that the cluster is neither [`TERMINATED` nor `TERMINATING`](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/python_submissions.py#L210) and then skips the if statement, stumbles in to the [create context](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/python_submissions.py#L215) call and dies a death a la: `Cluster <cluster_id> not currently ready for driver client (currently Pending)`.
3. This seems to be re-enforced by the fact that a quick search shows the "pending" status does not appear in any of the python code, so I'm guessing unhandled at the moment?

I think an elegant solution might be to shift the waiting part ([seen here](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/python_submissions.py#L282-L298)) out of `DBContext.start_cluster` into its own function `DBContext.wait_for_running_cluster` and then call this in `DBContext.create` directly after (but outside) the `TERMINATED` conditional:

1. This adds at least one `get_cluster_status` call to each context creation, but that feels right since we need to know its `RUNNING` before context can be created.
2. This also delegates the "shared state" of the separate threads (is my cluster ready or not) to the API, which is easy to share amongst the threads. 
3. That said, I've only skimmed the code, could be the wrong approach. 

My work around for now is to boot the cluster at the start of the job in a separate API call.  This assumes the cluster doesn't auto-terminate before the python models run, which isn't great.

jurasan:
Resolves #347

### Description

When multiple threads attempt to start a cluster, it's possible that the check [here](https://github.com/databricks/dbt-databricks/blob/37e16a5b5238f0d8cb7b99777db9fb6e8efeb7f3/dbt/adapters/databricks/python_submissions.py#L334) is True, but by the time it tries to trigger cluster start [here](https://github.com/databricks/dbt-databricks/blob/37e16a5b5238f0d8cb7b99777db9fb6e8efeb7f3/dbt/adapters/databricks/python_submissions.py#L392), the cluster has already been started by another thread and is in "RUNNING" or "PENDING" status. 

In such cases API returns an error "Cluster [cluster_id] is in unexpected state Pending." or "Cluster [cluster_id] is in unexpected state Running."
Current logic throws exceptions in this case. This fix allows it to continue and wait for cluster start if needed. 

This could probably be also backported to versions 1.6 and 1.7

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - Always use lower case when gathering metadata (since objects are stored internally as lower case regardless of how we create them) ([742](https://github.com/databricks/dbt-databricks/pull/742))
 - Persist table comments for python models ([743](https://github.com/databricks/dbt-databricks/pull/743))
 - Stop cursor destructor warnings ([744](https://github.com/databricks/dbt-databricks/pull/744))
-- Race condition on cluster creation([751](https://github.com/databricks/dbt-databricks/pull/751))
+- Race condition on cluster creation. (thanks @jurasan!) ([751](https://github.com/databricks/dbt-databricks/pull/751))
 
 ## dbt-databricks 1.8.4 (July 17, 2024)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Always use lower case when gathering metadata (since objects are stored internally as lower case regardless of how we create them) ([742](https://github.com/databricks/dbt-databricks/pull/742))
 - Persist table comments for python models ([743](https://github.com/databricks/dbt-databricks/pull/743))
 - Stop cursor destructor warnings ([744](https://github.com/databricks/dbt-databricks/pull/744))
+- Race condition on cluster creation([751](https://github.com/databricks/dbt-databricks/pull/751))
 
 ## dbt-databricks 1.8.4 (July 17, 2024)
 
--- a/dbt/adapters/databricks/python_submissions.py
+++ b/dbt/adapters/databricks/python_submissions.py
@@ -395,7 +395,9 @@ class DBContext:
             json={"cluster_id": self.cluster_id},
         )
         if response.status_code != 200:
-            raise DbtRuntimeError(f"Error starting terminated cluster.\n {response.content!r}")
+            current_status = self.get_cluster_status().get("state", "").upper()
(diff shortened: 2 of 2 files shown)


benc-db:
Looks good, running functional tests now.

benc-db:
Can you either give me edit, or add your 'thank you' to the changelog?  Just want to make sure you're credited in there :)

benc-db:
I was only planning to release in 1.8.5, but if you are on an earlier version, I will backport.

Issue #756, opened 2024-08-06
Run dbt python models as full-fledged jobs (not one-time job submission)


kdazzle:
### Describe the feature

I'd like to leverage Databricks' job running functionality but still use dbt to manage the orchestration and development workflow. So, from the Databricks side, instead of running my python jobs using the one-time submission API, I would want a full-fledged Workflow Job to be created for certain dbt python models. This would be a new submission_method. I think it will fill in some gaps in the dbt-databricks integration and make it easier for heavy Databricks users to migrate pipelines into dbt/dbt cloud.

One of the reasons is that Dbt Cloud's job management features are limited compared to Databricks. Ex: natively, dbt doesn't offer retries out of the box, timeout warnings, runs timeout after a max of 24 hours, their notifications need a lot of work, the Databricks run history is great, etc. All things that Databricks has nailed down pretty well over the years.

Another reason is that, even if dbt had feature parity, there are a bunch of people who are using these features from the Databricks side, and who are happy with it. I'd like to leverage dbt for the dev lifecycle/orchestration improvements, and continue to leverage Databricks for other things. Allowing models to materialize as Workflows accomplishes both of those goals. This makes it easier for people to move their workloads into dbt.

#### Implementation

See #762 

TLDR; Creates a Databricks Workflow for each model called `my-database-my-schema_my-model__dbt`. It gets triggered to run in the usual dbt manner (`dbt run --select=my_model`).

Also allows `post_hook_tasks` to be added to the workflow.

### Describe alternatives you've considered

The current setup, via the dbt-databricks all-purpose_cluster/job_cluster methods. But these are somewhat limiting because:

- One-time runs ([docs](https://docs.databricks.com/api/gcp/workspace/jobs/submit)) are difficult to find in the UI, which makes debugging more difficult
- Backfills + initial runs - due to a maximum dbt job runtime of 24 hours, it is difficult to run jobs that may take several days to complete
- Using a workflow prevents the same model from running twice at the same time (where the same model is running in two different pipelines - ex: in a modified+ and in a scheduled job)
- We use Databricks' built-in job run history quite a bit to see how well a job performs over time.
- People often want to run their job outside of a schedule for a variety of reasons (some good, some less so). Being able to do this in Databricks by hitting "play" is much easier than going into dbt to create an ad hoc job each time this comes up.
- Alerting/retries - built-in email notification, retry, and tagging support for workflows 
[...]

PR #762, opened 2024-08-08
Draft: #756 - implement python workflow submissions

kdazzle:
WIP - Stubs out implementation for #756

This pretty much implements what a workflow job submission type would look like, though I'm sure I'm missing something. Tests haven't been added yet.

### Sample

Outside of the new submission type, models are the same. Here is what one could look like:

```python
# my_model.py
import pyspark.sql.types as T
import pyspark.sql.functions as F


def model(dbt, session):
    dbt.config(
        materialized='incremental',
        submission_method='workflow_job'
    )

    output_schema = T.StructType([
        T.StructField("id", T.StringType(), True),
        T.StructField("odometer_meters", T.DoubleType(), True),
        T.StructField("timestamp", T.TimestampType(), True),
    ])
    return spark.createDataFrame(data=spark.sparkContext.emptyRDD(), schema=output_schema)
```

The config for a model could look like (forgive my jsonification...yaml data structures still freak me out):

```yaml
models:
  - name: my_model
      workflow_job_config:
        email_notifications: {
          on_failure: ["reynoldxin@databricks.com"]
        }
        max_retries: 2
        timeout_seconds: 18000
        existing_job_id: 12341234  # not part of Databricks API (+ optional)
        additional_task_settings: {  # not part of Databricks API (+ optional)
          "task_key": "my_dbt_task"
        }
        post_hook_tasks: [{  # not part of Databricks API (+ optional)
          "depends_on": [{ "task_key": "my_dbt_task" }],
          "task_key": 'OPTIMIZE_AND_VACUUM',
          "notebook_task": {
            "notebook_path": "/my_notebook_path",
            "source": "WORKSPACE",
          },
        }]
        grants:  # not part of Databricks API (+ optional)
          view: [
            {"group_name": "marketing-team"},
          ]
          run: [
            {"user_name": "alighodsi@databricks.com"}
          ]
          manage: []
      job_cluster_config:
        spark_version: "15.3.x-scala2.12"
        node_type_id: "rd-fleet.2xlarge"
        runtime_engine: "STANDARD"
        data_security_mode: "SINGLE_USER"
        autoscale: {
          "min_workers": 1,
          "max_workers": 4
        }
```

### Explanation

For all of the dbt configs that I added (in addition to the Databricks API attributes), I tried to roughly mediate between the dbt convention of requiring minimal configuration, but also allowing for the full flexibility of the Databricks API. Attribute names were trying to split the difference between the Databricks API and the dbt API. Happy to change the approach for anything.

- added `existing_job_id` in case users want to reuse an existing workflow. If no `name` is provided in this config, it will get renamed to the default job name (currently `f"dbt__{self.database}-{self.schema}-{self.identifier}"`)
- Job names must be unique unless `existing_job_id` is also provided
- The task key for the model run task is hardcoded as `task_a` - configurable in `additional_task_settings`
- Allow for "post_hook tasks"
  - Can specify a different cluster type using Databricks' `new_cluster` or `existing_cluster_id`. Leaving blank is serverless
  - `post_hook` might be a misnomer, because you could technically set the dbt model to depend on one of these tasks, making it also a pre hook
- `grants` - allow for permissions to be set on the workflow so that additional users/teams can run the job ad hoc if needed (for initial runs/backfills, etc). The owner is the user/service principal that deployed, and the format needs to follow the Databricks API where you specify whether the user is a user, group, or SP.
- `additional_task_settings` to add to/override the default dbt model task

### Todo:

- [ ] Reuse `all_purpose_cluster` attribute, similar to `job_cluster_config`?
- [x] Can I use a serverless job cluster? (by not defining any cluster)
- [x] Fix the run tracker
- [x] What happens if the workflow is already running?
  - I'd like the new dbt job run to start tracking the current Databricks workflow run, rather than failing
- [ ] Log if workflow permissions are being changed? (Kind of mimicking TF apply logs, which have been helpful in the past when table permissions had been unexpectedly broadened)

### Checklist


dbt/adapters/databricks/python_submissions.py
```diff
@@ -86,6 +86,9 @@ def insert_command(cls, command: CommandExecution) -> None:
 
     @classmethod
     def cancel_runs(cls, session: Session) -> None:
+        # TODO: let Databricks Workflows handle timeouts if a timeout is defined
```
kdazzle, line 89: Not sure if there's a good way to fit this in with the whole dbt workflow. My main use-case was the initial run/backfilling lots of data, but I think just pressing "play" on the job by hand and getting out of the dbt limitations would be good enough.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -162,3 +162,265 @@ def submit(self, compiled_code: str) -> None:
...
+                "url": f"https://{self.credentials.host}/api/2.1/jobs/runs/get?run_id={run_id}",
+                "headers": self.extra_headers,
+            },
+            get_state_func=lambda response: response.json()["state"]["life_cycle_state"],
+            terminal_states=("TERMINATED", "SKIPPED", "INTERNAL_ERROR"),
+            expected_end_state="TERMINATED",
+            get_state_msg_func=lambda response: response.json()["state"]["state_message"],
+        )
+
+        run_output = self.api_client.session.get(
```
kdazzle, line 251: all of these API calls are broken for now. I also need to integrate with the new polling tracker. Will get to that on Tuesday.

kdazzle: Should be fixed with latest commit

benc-db:
Looks like some syntax you're using does not work with python 3.8 based on unit test failures.

dbt/adapters/databricks/api_client.py
```diff
@@ -356,6 +361,93 @@ def cancel(self, run_id: str) -> None:
...
+        if active_runs_response.status_code != 200:
+            raise DbtRuntimeError(f"Error getting active runs.\n {active_runs_response.content!r}")
+
+        return active_runs_response.json().get("runs", [])
+
+class JobPermissionsApi(DatabricksApi):
+    def __init__(self, session: Session, host: str):
+        super().__init__(session, host, "/api/2.0/permissions/jobs")
+
+    def put(self, job_id, access_control_list):
```
benc-db, line 379: Should this be part of the operations that dbt handles?  If so, should there be an equivalent for the job runs approach?

dbt/adapters/databricks/api_client.py
```diff
@@ -356,6 +361,93 @@ def cancel(self, run_id: str) -> None:
...
+        """
+        response = self.session.post("/create", json=job_spec)
+
+        if response.status_code != 200:
+            raise DbtRuntimeError(f"Error creating Workflow.\n {response.content!r}")
+
+        logger.info(f"Workflow creation response={response.json()}")
+        return response.json()["job_id"]
+
+    def update_by_reset(self, job_id, job_spec):
```
benc-db, line 425: What is this intended for?  It's obscure enough that a doc string might be useful.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -162,3 +162,161 @@ def submit(self, compiled_code: str) -> None:
 class ServerlessClusterPythonJobHelper(BaseDatabricksHelper):
     def submit(self, compiled_code: str) -> None:
         self._submit_through_notebook(compiled_code, {})
+
+
+class WorkflowPythonJobHelper(BaseDatabricksHelper):
+
+    @property
+    def default_job_name(self) -> str:
+        return f"{self.database}-{self.schema}-{self.identifier}__dbt"
```
benc-db, line 171: Think it might be helpful to put the dbt up front.  That way if you sort in the UI, all of the dbt jobs are together.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -162,3 +162,161 @@ def submit(self, compiled_code: str) -> None:
...
+
+    @property
+    def notebook_path(self) -> str:
+        return f"{self.notebook_dir}/{self.identifier}"
+
+    @property
+    def notebook_dir(self) -> str:
+        return f"/Shared/dbt_python_model/{self.database}/{self.schema}"
```
benc-db, line 179: We should probably use the folder api logic.  I've been told that using Shared is an anti-pattern and that we might get rid of it as we continue with improved governance.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -162,3 +162,161 @@ def submit(self, compiled_code: str) -> None:
...
+        workflow_spec["tasks"] = [notebook_task] + post_hook_tasks
+        return workflow_spec
+
+    def _submit_through_workflow(self, compiled_code: str, workflow_spec) -> None:
+        self.api_client.workspace.upload_notebook(self.notebook_path, compiled_code)
+
+        job_id, is_new = self._get_or_create_job(workflow_spec)
+
+        if not is_new:
+            self.api_client.workflows.update_by_reset(job_id, workflow_spec)
```
benc-db, line 231: Even though the API is 'reset', I think update_job_settings is probably more apt.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -162,3 +162,161 @@ def submit(self, compiled_code: str) -> None:
...
+                        return principal['service_principal_name'], 'service_principal_name'
+
+        raise DbtRuntimeError(f"Error getting current owner for Databricks workflow.\n {response.content!r}")
+
+    def _get_or_trigger_job_run(self, job_id: str):
+        """
+        :return: the run id
+        """
+        active_runs = self.api_client.job_runs.list_active_runs_for_job(job_id)
+        if len(active_runs) > 0:
```
benc-db, line 339: If the workflow is already running, I think we still need to schedule a run after it completes...this is one of the edge-cases that Amy is worried about.  Consider that in the dbt run, the tables that this model depends on may have been updated after that job run started; in order to ensure that downstream tables get those updates, we need to run the job again.

benc-db (changes requested):
Mostly minor.  Biggest thing is that I think if there is already an active run, we need to wait and then execute another one.

dbt/adapters/databricks/api_client.py
kdazzle: I'd argue it should be since dbt creates these workflows, so they should also at least allow for the possibility of managing permissions. Otherwise these objects are kind of orphaned and there has to be some other process. Dbt abdicates that responsibility on schemas, which puts everyone in an awkward position.

> If so, should there be an equivalent for the job runs approach?

I'm not sure it's as necessary for job runs, since those are usually just a one-time thing. But I'm happy to add it in there if you think that makes the most sense.

dbt/adapters/databricks/api_client.py
kdazzle: Yeah good call. It's basically a PUT, but it's called a reset in the Databricks API. It's to make the job reflect the config, with any changes.

dbt/adapters/databricks/api_client.py
benc-db: Good point about schema.  That's in my backlog of pain points :P

dbt/adapters/databricks/python_models/python_submissions.py
kdazzle: good point, thanks Ben

dbt/adapters/databricks/python_models/python_submissions.py
kdazzle: Gotcha - that's a good point. That shouldn't be too hard to work in

dbt/adapters/databricks/python_models/python_submissions.py
kdazzle: Yeah, I like that

dbt/adapters/databricks/api_client.py
kdazzle: I'll take a look at pulling that into job runs too

dbt/adapters/databricks/api_client.py
kdazzle: Changed method name to `update_job_settings` per [your other comment](https://github.com/databricks/dbt-databricks/pull/762#discussion_r1783546514)

dbt/adapters/databricks/python_models/python_submissions.py
kdazzle: changed

dbt/adapters/databricks/python_models/python_submissions.py
kdazzle: updated to use the new API classes

dbt/adapters/databricks/python_models/python_submissions.py
kdazzle: changed

dbt/adapters/databricks/python_models/python_submissions.py
kdazzle: Didn't realize there was an option for that in the Databricks API already - that made things easy. Done.

benc-db:
Going to pull/push to origin to run the existing functional tests.  We should add one for this new code.  Let me know if you need help with that.

tests/functional/adapter/python_model/test_python_model.py
```diff
@@ -144,3 +144,41 @@ def test_expected_handling_of_complex_config(self, project):
...
+class TestWorkflowJob:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {
+            "schema.yml": override_fixtures.workflow_schema,
+            "my_workflow_model.py": override_fixtures.simple_python_model,
+        }
+
+    def test_workflow_run(self, project):
+        util.run_dbt(["run", "-s", "my_workflow_model"])
```
kdazzle, line 160: Running this on my local stopped working for some reason, so this might fail

docs/workflow-job-submission.md
```diff
@@ -0,0 +1,180 @@
...
+  ]
+  manage: []
+```
+
+#### Post hooks
+
+It is possible to add in python hooks by using the `config.workflow_job_config.post_hook_tasks` 
+attribute. If there is a cluster defined in `config.job_cluster_config`, that will get pulled 
+into the hook as a new cluster. For reusable job clusters or an existing job cluster, you will 
+have to define that in the post hook task definition manually.
```
kdazzle, line 180: @benc-db - I've been going back and forth on this behavior. On the one hand, defaulting to separate job clusters might not be ideal if their definitions are the same. On the other hand, this code seems complex enough as it is and doesn't need _too_ much dbt automagic. If you have an opinion, let me know and I can make any changes, otherwise I'll probably just leave it.

kdazzle: Actually, I'm going to scratch that and make clusters be defined for the post hooks. Otherwise you can't use a job cluster for the main notebook and serverless for a followup. And it will take out some magic.

tests/unit/python/test_python_submissions.py
```diff
@@ -247,97 +247,3 @@ def test_build_job_spec_with_post_hooks(self, mock_api_client):
         assert len(result["tasks"]) == 2
         assert result["tasks"][1]["task_key"] == "task_b"
         assert result["tasks"][1]["new_cluster"]["spark_version"] == "14.3.x-scala2.12"
-
-    @patch("dbt.adapters.databricks.python_models.python_submissions.DatabricksApiClient")
-    def test_build_job_spec_with_post_hooks_serverless_job_cluster(self, mock_api_client):
```
kdazzle: Removing these since the logic to muck around with the cluster settings in additional tasks was removed [here](https://github.com/databricks/dbt-databricks/pull/762/commits/38c3ce48fd0f76bb1cad5604180de7f174dea8db)

benc-db (approved):

benc-db:
Going to merge in 1.9.latest changes (which is basically only 1.8 changes), ensure tests still pass, then merge.

Issue #760, opened 2024-08-07
tblproperties and Comment are not updated for incremental


elca-anh:
### Describe the bug
My use case is a model with materialization 'incremental'. Model already exists. I am adding tblproperties option to be able to set delta.dataSkippiongStatsColumn. This property is not set.
However, if I run a full refresh, the property is set.
Since I checked similar behavior for the Comment that is not updated in Unity Catalog.

### Steps To Reproduce
- Set or update tblproperties/comment on an existing incremental model
- dbt run
- Check in Databricks (DESC DETAIL table) or in the Databricks UI (for the comment) that the property is not set/updated

### Expected behavior
Detect creation or update of the tblproerties and apply those.

Looking at dbt-databricks\dbt\adapters\databricks\relation_configs\incremental.py, config_components only contains TagsProcessor, 

But: 
- the materialized_view's config_components contains [PartitionedByProcessor, CommentProcessor, TblPropertiesProcessor, RefreshProcessor,    QueryProcessor], 
- the streaming_table's  config_components contains [PartitionedByProcessor, CommentProcessor,        TblPropertiesProcessor,  RefreshProcessor]. 

But I am not sure how those are applied. For example I see that liquid clustering is applied specifically by  dbt-databricks\dbt\include\databricks\macros\relations\liquid_clustering.sql
I also see that tblproperties are only updated for Python (apply_tblproperties_python) in dbt-databricks\dbt\include\databricks\macros\materializations\incremental\incremental.sql

### System information
**The output of `dbt --version`:**
```
Core: 1.8.4
Databricks: 1.8.4
Spark: 1.8.0
```

**The operating system you're using:**

Windows

**The output of `python --version`:**

Python 3.10

benc-db:
I thought I had already addressed this for tblproperties, but apparently not.  I believe there is currently a PR for fixing it for table comments (should already work for column comments).

benc-db:
Thanks for reporting.

PR #765, opened 2024-08-12
Update tblproperties on incremental runs

benc-db:
Addresses #760 (doesn't fully resolve because 760 also discusses comment updates)

### Description

Incremental tables have not been updating tblproperties on incremental runs.  This PR adds support for add/replace tblproperties on incremental run.  Delete was tested, but too risky/complicated, as Databricks sets some tblproperties for tables on its own.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -817,4 +817,5 @@ class IncrementalTableAPI(RelationAPIBase[IncrementalTableConfig]):
 
         if not relation.is_hive_metastore():
             results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
+        results["show_tblproperties"] = adapter.execute_macro("fetch_tbl_properties", kwargs=kwargs)
         return results
--- a/dbt/adapters/databricks/relation_configs/incremental.py
+++ b/dbt/adapters/databricks/relation_configs/incremental.py
@@ -5,10 +5,11 @@ from dbt.adapters.databricks.relation_configs.base import DatabricksComponentCon
 from dbt.adapters.databricks.relation_configs.base import DatabricksRelationChangeSet
 from dbt.adapters.databricks.relation_configs.base import DatabricksRelationConfigBase
 from dbt.adapters.databricks.relation_configs.tags import TagsProcessor
+from dbt.adapters.databricks.relation_configs.tblproperties import TblPropertiesProcessor
 
 
 class IncrementalTableConfig(DatabricksRelationConfigBase):
-    config_components = [TagsProcessor]
+    config_components = [TagsProcessor, TblPropertiesProcessor]
 
     def get_changeset(
         self, existing: "IncrementalTableConfig"
--- a/dbt/adapters/databricks/relation_configs/tags.py
+++ b/dbt/adapters/databricks/relation_configs/tags.py
@@ -22,7 +22,9 @@ class TagsConfig(DatabricksComponentConfig):
         for k in other.set_tags.keys():
             if k not in self.set_tags:
                 to_unset.append(k)
-        return TagsConfig(set_tags=self.set_tags, unset_tags=to_unset)
+        if self.set_tags or to_unset:
+            return TagsConfig(set_tags=self.set_tags, unset_tags=to_unset)
+        return None
 
 
 class TagsProcessor(DatabricksComponentProcessor[TagsConfig]):
--- a/dbt/include/databricks/macros/materializations/incremental/incremental.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/incremental.sql
@@ -38,6 +38,10 @@
     {%- endcall -%}
     {% do persist_constraints(target_relation, model) %}
     {% do apply_tags(target_relation, tags) %}
+    {%- if language == 'python' -%}
+      {%- do apply_tblproperties(target_relation, tblproperties) %}
+    {%- endif -%}
+
     {% do persist_docs(target_relation, model, for_relation=language=='python') %}
   {%- elif existing_relation.is_view or existing_relation.is_materialized_view or existing_relation.is_streaming_table or should_full_refresh() -%}
     {#-- Relation must be dropped & recreated --#}
@@ -96,9 +100,13 @@
     {%- endif -%}
     {% do apply_liquid_clustered_cols(target_relation) %}
     {% if _configuration_changes is not none %}
-      {% set tags = _configuration_changes.changes["tags"] %}
+      {% set tags = _configuration_changes.changes.get("tags", None) %}
+      {% set tblproperties = _configuration_changes.changes.get("tblproperties", None) %}
       {% if tags is not none %}
         {% do apply_tags(target_relation, tags.set_tags, tags.unset_tags) %}
+      {%- endif -%}
+      {% if tblproperties is not none %}
+        {% do apply_tblproperties(target_relation, tblproperties.tblproperties) %}
       {%- endif -%}
     {%- endif -%}
     {% do persist_docs(target_relation, model, for_relation=True) %}
@@ -106,7 +114,6 @@
 
   {% set should_revoke = should_revoke(existing_relation, full_refresh_mode) %}
   {% do apply_grants(target_relation, grant_config, should_revoke) %}
-  {% do apply_tblproperties_python(target_relation, tblproperties, language) %}
   {% do optimize(target_relation) %}
 
   {{ run_hooks(post_hooks) }}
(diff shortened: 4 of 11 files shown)


benc-db:
Adding Eric, Yunbo to review since Jacky is out.  I will comment the code so that you have some context.

dbt/adapters/databricks/impl.py
```diff
@@ -817,4 +817,5 @@ def _describe_relation(
 
         if not relation.is_hive_metastore():
             results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
+        results["show_tblproperties"] = adapter.execute_macro("fetch_tbl_properties", kwargs=kwargs)
```
benc-db, line 820: Core idea with MV/ST and now incremental is to check for what's in Databricks and compare against what is configured in dbt project.  This is grabbing the tblproperties from the table that has already been deployed to check if they have changed.

dbt/adapters/databricks/relation_configs/incremental.py
```diff
@@ -5,10 +5,11 @@
 from dbt.adapters.databricks.relation_configs.base import DatabricksRelationChangeSet
 from dbt.adapters.databricks.relation_configs.base import DatabricksRelationConfigBase
 from dbt.adapters.databricks.relation_configs.tags import TagsProcessor
+from dbt.adapters.databricks.relation_configs.tblproperties import TblPropertiesProcessor
 
 
 class IncrementalTableConfig(DatabricksRelationConfigBase):
-    config_components = [TagsProcessor]
+    config_components = [TagsProcessor, TblPropertiesProcessor]
```
benc-db, line 12: Previously we only handled Tags this way (i.e. this is a new approach to updating the metadata of incremental tables).  This structure checks for each configured component to see if there are any metadata changes.

dbt/adapters/databricks/relation_configs/tags.py
```diff
@@ -22,7 +22,9 @@ def get_diff(self, other: "TagsConfig") -> Optional["TagsConfig"]:
         for k in other.set_tags.keys():
             if k not in self.set_tags:
                 to_unset.append(k)
-        return TagsConfig(set_tags=self.set_tags, unset_tags=to_unset)
+        if self.set_tags or to_unset:
```
benc-db, line 25: I realized while testing that we were potentially issuing alter statements that were unneeded.  This says only return a diff if we are either setting or unsetting some tags.

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -38,6 +38,10 @@
     {%- endcall -%}
     {% do persist_constraints(target_relation, model) %}
     {% do apply_tags(target_relation, tags) %}
+    {%- if language == 'python' -%}
```
benc-db, line 41: If were a creating a new table (rather than incremental), we need to apply tblproperties to python models, since for now there is no python API to use.

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -94,16 +98,19 @@
     {%- endif -%}
     {% do apply_liquid_clustered_cols(target_relation) %}
     {% if _configuration_changes is not none %}
-      {% set tags = _configuration_changes.changes["tags"] %}
+      {% set tags = _configuration_changes.changes.get("tags", None) %}
```
benc-db, line 101: Now that we have two possible changes, we need to be able to check for presence of one or the other in the change set.

dbt/include/databricks/macros/materializations/table.sql
```diff
@@ -29,8 +29,9 @@
 
   {% set should_revoke = should_revoke(old_relation, full_refresh_mode=True) %}
   {% do apply_grants(target_relation, grant_config, should_revoke) %}
-  {% do apply_tblproperties_python(target_relation, tblproperties, language) %}
-
+  {% if language=="python" %}
+    {% do apply_tblproperties(target_relation, tblproperties) %}
```
benc-db, line 33: Previously we only used alter statements to apply tblproperties to python models.  Now we reuse this macro for other things.

dbt/include/databricks/macros/relations/tblproperties.sql
```diff
@@ -13,10 +13,14 @@
   {%- endif %}
 {%- endmacro -%}
 
-{% macro apply_tblproperties_python(relation, tblproperties, language) -%}
-  {%- if tblproperties and language == 'python' -%}
-    {%- call statement('python_tblproperties') -%}
-      alter table {{ relation }} set {{ tblproperties_clause() }}
+{% macro apply_tblproperties(relation, tblproperties) -%}
```
benc-db, line 16: Construct the ALTER statement SQL

tests/functional/adapter/incremental/fixtures.py
```diff
@@ -1,3 +1,6 @@
+from dbt.adapters.databricks.relation_configs import tblproperties
```
benc-db, line 1: This file has test fixtures for incremental tests.

tests/functional/adapter/incremental/fixtures.py
```diff
@@ -75,6 +78,36 @@
         - name: color
 """
 
+tblproperties_a = """
```
benc-db, line 81: Tblproperties tests/fixtures looks nearly identical to Tags tests/fixtures because their syntax and behavior is currently the same.

tests/functional/adapter/incremental/test_incremental_tblproperties.py
```diff
@@ -0,0 +1,51 @@
...
+    # For now, only add/change, no deletes
+    def test_changing_tblproperties(self, project):
+        util.run_dbt(["run"])
+        util.write_file(fixtures.tblproperties_b, "models", "schema.yml")
+        util.run_dbt(["run"])
+        results = project.run_sql(
+            "show tblproperties {database}.{schema}.merge_update_columns_sql",
+            fetch="all",
+        )
+        results_dict = {}
```
benc-db, line 23: This test only asserts that the tblproperties that we set have been added/updated.  Databricks has been increasing the number of tblproperties that get set automatically, so not safe to assert that these are the only tblproperties on the table.

tests/functional/adapter/incremental/test_incremental_tblproperties.py
```diff
@@ -0,0 +1,51 @@
...
+        for result in results:
+            results_dict[result.key] = result.value
+        assert results_dict["c"] == "e"
+        assert results_dict["d"] == "f"
+
+
+@pytest.mark.skip_profile("databricks_cluster")
+class TestIncrementalPythonTblproperties:
```
benc-db, line 31: Verify that we haven't broken anything with Python model tblproperty support.

dbt/adapters/databricks/relation_configs/incremental.py
elca-anh: Shouldn't we also fix the update of the table comment?

benc-db: I need to double check, but I think the PR that I merged from @henlue yesterday could fix the incremental table comment issue.  If not, I'll send another PR to address that.

henlue: Yes the incremental table comments were fixed as part of my PR. However, they are updated on every incremental run, whether or not they are already present in databricks.

benc-db: I think that is fine.  We check for diffs on column comments because we can take one action to save us from many (updating every column).  For tblproperties, we could similarly update without checking, but if I ever do move to add deletes (would kind of require Databricks to stop writing tblproperties internally), it will be good to have the checking logic in place.

dbt/include/databricks/macros/materializations/table.sql
```diff
@@ -29,8 +29,9 @@
...
+  {% endif %}
   {%- do apply_tags(target_relation, tags) -%}
 
   {% do persist_docs(target_relation, model, for_relation=language=='python') %}
```
coetzeevs, line 37: @benc-db sorry for going back into a closed PR - but I'm curious, why only do persist_docs if the language is 'python' and the materialisation is table?

coetzeevs: To more clear about the "why" behind my question: we're materialising a model as a table in Databricks and I've been trying to understand why the persist_docs behaviour isn't yielding what I'd expect (i.e. persisted documentation).

benc-db: for_relation flag says apply a model-level comment.  For SQL models, we apply the model-level comment in the table create SQL, so we only need to persist it after table create for python models.

coetzeevs: Thank you for clarifying, really appreciate it.

PR #766, opened 2024-08-13
Eric wang 1990/azure ad

eric-wang-1990:
Resolves #

### Checklist


dbt/adapters/databricks/api_client.py
```diff
@@ -238,170 +84,88 @@ class CommandExecution(object):
...
-                    )
-
-    def cancel(self, run_id: str) -> None:
+    def cancel(self, run_id: int) -> None:
         logger.debug(f"Cancelling run id {run_id}")
-        response = self.session.post("/cancel", json={"run_id": run_id})
-
-        if response.status_code != 200:
-            raise DbtRuntimeError(f"Cancel run {run_id} failed.\n {response.content!r}")
+        self.client.jobs.cancel_run_and_wait(run_id)
```
eric-wang-1990, line 155: Do we need to stick to the old way which is cancel Async? https://docs.databricks.com/api/workspace/jobs/cancelrun

dbt/adapters/databricks/api_client.py
```diff
@@ -143,91 +38,42 @@ def get_folder(self, _: str, schema: str) -> str:
...
-        response = self.session.post(
-            "/import",
-            json={
-                "path": path,
-                "content": b64_encoded_content,
-                "language": "PYTHON",
-                "overwrite": True,
-                "format": "SOURCE",
-            },
+        self.client.workspace.import_(
```
eric-wang-1990, line 70: Seems like the SDK does not return a response, does it ever throw? https://databricks-sdk-py.readthedocs.io/en/latest/workspace/workspace/workspace.html#databricks.sdk.service.workspace.WorkspaceExt.import_

dbt/adapters/databricks/api_client.py
```diff
@@ -238,170 +84,88 @@ class CommandExecution(object):
...
-            "/execute",
-            json={
-                "clusterId": cluster_id,
-                "contextId": context_id,
-                "language": SUBMISSION_LANGUAGE,
-                "command": command,
-            },
+    def execute(self, cluster_id: str, command: str) -> CommandExecution:
+        self.client.clusters.ensure_cluster_is_running(cluster_id)
+        context = self.client.command_execution.create_and_wait(
```
eric-wang-1990, line 100: Why create context here not before?

dbt/adapters/databricks/api_client.py
```diff
@@ -238,170 +84,88 @@ class CommandExecution(object):
...
+                self.client.command_execution.destroy(cluster_id, context.id)
+            raise e
 
     def cancel(self, command: CommandExecution) -> None:
         logger.debug(f"Cancelling command {command}")
-        response = self.session.post("/cancel", json=command.model_dump())
-
-        if response.status_code != 200:
-            raise DbtRuntimeError(f"Cancel command {command} failed.\n {response.content!r}")
+        self.client.command_execution.cancel_and_wait()
```
eric-wang-1990, line 122: Should we just call cancel which is Async?

dbt/adapters/databricks/api_client.py
```diff
@@ -238,170 +84,88 @@ class CommandExecution(object):
...
 class DatabricksApiClient:
     def __init__(
         self,
-        session: Session,
-        host: str,
-        polling_interval: int,
+        workspace_client: WorkspaceClient,
         timeout: int,
-        use_user_folder: bool,
+        use_user_folder: bool = False,
```
eric-wang-1990, line 163: What is this for? Is it needed?

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -142,12 +142,9 @@ def submit(self, compiled_code: str) -> None:
                 config["existing_cluster_id"] = self.cluster_id
             self._submit_through_notebook(compiled_code, self._update_with_acls(config))
         else:
-            context_id = self.api_client.command_contexts.create(self.cluster_id)
```
eric-wang-1990, line 145: OK so the context is created for every submission of a compiled code so probably fine to just move to commands execution?

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -156,7 +153,6 @@ def submit(self, compiled_code: str) -> None:
             finally:
                 if command_exec:
                     self.tracker.remove_command(command_exec)
-                self.api_client.command_contexts.destroy(self.cluster_id, context_id)
```
eric-wang-1990, line 159: Should we keep this to destroy the context?

tests/unit/test_auth.py
```diff
@@ -55,6 +54,7 @@ def test_u2m(self):
         assert headers == headers2
 
 
+@pytest.mark.skip(reason="Broken after rewriting auth")
```
eric-wang-1990, line 57: Should I address these or is it fine to be broken?

dbt/adapters/databricks/python_models/run_tracking.py
```diff
@@ -8,20 +8,20 @@
 
 
 class PythonRunTracker(object):
-    _run_ids: Set[str] = set()
+    _run_ids: Set[int] = set()
```
eric-wang-1990, line 11: I'm not seeing how this is used, but is it intended to move to int?

tests/unit/test_auth.py
benc-db: We want to test token auth, but we no longer use ._provider_from_dict, so we need a new test.

dbt/adapters/databricks/python_models/run_tracking.py
benc-db: this is because with migrating to the workspace client, we get a real type, as opposed to strings.

dbt/adapters/databricks/api_client.py
benc-db: Yes, it will throw, and generally these sorts of exceptions we just allow to bubble up to the user.

PR #772, opened 2024-08-19
fix dbt-databricks scopes

eric-wang-1990:
### Description

This change only affect U2M OAuth for dbt-databricks.
The dbt-databricks oauth app has changed its scope to all-apis, offline_access, the redirect url has been changed to http://localhost:8020, http://localhost:8050. This will enable customer being able to use the default dbt-databricks oauth app to run python models.
The scope/redirect_url changes have already landed in all AWS & GCP regions, for Azure the govCloud regions are still pending.
But since for Azure there is another bug that we need to fix to use this dbt-databricks client, it is fine for this PR to land without fully ramped up in Azure.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ### Fixes
 
-- Fix the default scope for `dbt-databricks` to `all-apis` so it can run python models
+- Fix the default scope for `dbt-databricks` to `all-apis` so it can run python models([772](https://github.com/databricks/dbt-databricks/pull/772))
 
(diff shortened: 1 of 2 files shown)


benc-db:
@eric-wang-1990 These changes should maybe go to main, so they don't need to wait for 1.9.  We might be able to get these out this week if they're in main (depending on if your changes have rolled to all environments in time).

benc-db:
I merged your other PR, and I'm going to merge main in 1.9, so you won't need this PR anymore.

PR #774, opened 2024-08-20
Issue 747 primary key autoname

### LINKED ISSUE #747: Auto-named constraints does not take into account for the model/table name
### Describe the feature
Column constraints can be named explicitly or automatically in the YAML model documentation. 
When done automatically, the name is created by hashing the name of the column. Since constraint name shall be unique per schema, this is far from ideal as multiple columns might have the same name.
Explicit naming is not magic either since there can be no expansion of the name. In our case, we use prefixes to have several versions of the models in the dev environment. Since there are no macros within the YAML we cannot use the prefix in the constraint name.

Current code in dbt-databricks\dbt\include\databricks\macros\relations\constraints.sql:107:
{% set name = constraint.get("name") %}
    {% if not name and local_md5 %}
      {{ exceptions.warn("Constraint of type " ~ type ~ " with no `name` provided. Generating hash instead.") }}
      {%- set name = local_md5 (column.get("name", "") ~ ";" ~ expression ~ ";") -%}
    {% endif %}

Add the model name as part of the hash

### Additional context
Constraint documentation on models within YAML file.

### Who will this benefit?
Model documentation

Yes, I can contribute on the validation of this feature and other tests

elca-anh:
Resolves #747 

### Description

- Add the relation identifier (i.e. table name) to the auto generated name of constraints
- For foreign-key constraints, add either the statement of the list of columns including the parent column

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Fixes
 
 - Persist table comments for incremental models, snapshots and dbt clone (thanks @henlue!) ([750](https://github.com/databricks/dbt-databricks/pull/750))
-- Add relation identifier (i.e. table name) in auto generated constraint names, also adding the statement of table list for foreign keys
+- Add relation identifier (i.e. table name) in auto generated constraint names, also adding the statement of table list for foreign keys  ([774](https://github.com/databricks/dbt-databricks/pull/774))
 
 ## dbt-databricks 1.8.5 (August 6, 2024)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Fixes
 
 - Persist table comments for incremental models, snapshots and dbt clone (thanks @henlue!) ([750](https://github.com/databricks/dbt-databricks/pull/750))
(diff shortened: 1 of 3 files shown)


benc-db (dismissed):

benc-db (approved):

PtrckGdwn:
Hi, I am really keen to use the non-expression method for foreign keys. Unfortunately, when I try it, I get an error message like:  "No parent table defined for foreign key: " followed by a trace. 

I have configured the schema for the model like so:

```
  - name: example__child_table
    constraints:
      - type: primary_key
        name: pk_example__child_table
        columns: ["id"]
      - type: not_null
        columns: ["id", "name", "parent_id"]
      - type: foreign_key
        name: fk_example__child_table_1
        columns: ["parent_id"]
        parent: parent_table
        parent_columns: ["id"]
    columns:
      - name: id
      - name: name
      - name: parent_id
```

Any tips on how to get this working? I've had a look at the macro code, and it appears the constraint.get("parent") call is not returning anything, hence the error message, but I can't see anything obvious I am doing wrong.

Running with adaptor version 1.8.5

PR #789, opened 2024-09-13
Fix foreign key constraints

### LINKED ISSUE #784: Foreign Key constraint error: No parent table defined for foreign key
### Describe the bug
When trying to create a foreign key constraint on a relation using the `parent` and `parent_columns` constraint attributes, I am getting an error at run time: `No parent table defined for foreign key` which is followed by a trace.

### Steps To Reproduce
I have configured the model's schema like so:

```
  - name: example__child_table
    constraints:
      - type: primary_key
        name: pk_example__child_table
        columns: ["id"]
      - type: not_null
        columns: ["id", "name", "parent_id"]
      - type: foreign_key
        name: fk_example__child_table_1
        columns: ["parent_id"]
        parent: parent_table
        parent_columns: ["id"]
    columns:
      - name: id
      - name: name
      - name: parent_id
```

When running the project this results in the error message described above.

### Expected behavior
The foreign key sql should be generated, and assume all else is good (e.g., the parent_table has a primary key) the foreign key should be created.

### System information
```
Running with dbt=1.8.6
Registered adapter: databricks=1.8.5
```

Running as a dbt task in a Databricks workflow. 
Compute: Single node: Standard_DS3_v2 · DBR: 14.3.x-photon-scala2.12

### Additional context
The traditional `expression` method of creating a foreign key works as documented, but I would like to avoid using it because it has a hard coded schema name. The `parent` + `parent_columns` method assumes that the parent table is in the same schema as the child, which for me is true.

benc-db:
Resolves #784 

### Description

Looking into #784, constraints strip off config that dbt hasn't specified, so parent and parent_columns were not making it to the macro.  It looks as though 'to' and 'to_columns' can fill this role instead.  Will need to update any documentation out there that references parent and parent_columns.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add relation identifier (i.e. table name) in auto generated constraint names, also adding the statement of table list for foreign keys (thanks @elca-anh!) ([774](https://github.com/databricks/dbt-databricks/pull/774))
 - Update tblproperties on incremental runs. Note: only adds/edits. Deletes are too risky/complex for now ([765](https://github.com/databricks/dbt-databricks/pull/765))
 - Update default scope/redirect Url for OAuth U2M, so with default OAuth app user can run python models ([776](https://github.com/databricks/dbt-databricks/pull/776))
+- Fix foreign key constraints by switching from `parent` to `to` and `parent_columns` to `to_columns` ([789](https://github.com/databricks/dbt-databricks/pull/789))
 
 ## dbt-databricks 1.8.5 (August 6, 2024)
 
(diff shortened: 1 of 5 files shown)


eric-wang-1990:
Need a little more context to understand, seems we are only changing what we expect the users to provide for the foreign key table and column. You mention parent and parent_columns are not working because of dbt will strip it out? Does to and to_columns work because that's a supported keyword by dbt? Any documents reference?

benc-db:
> Need a little more context to understand, seems we are only changing what we expect the users to provide for the foreign key table and column. You mention parent and parent_columns are not working because of dbt will strip it out? Does to and to_columns work because that's a supported keyword by dbt? Any documents reference?

Exactly.  And no, these columns are not referenced in the docs I can find, but are public fields on the ModelConstraint class.

eric-wang-1990 (approved):
Maybe add a reference to the dbt feature in our changelog as of why we change to this?
https://github.com/dbt-labs/dbt-common/commit/8edae92ac4e1c5814be028281215fd90ab29d9c6

PR #794, opened 2024-09-17
Add column comments for DLT tables using dbt schema 

### LINKED ISSUE #793: Add support for auto-generating column comments in Delta Live Tables from dbt model descriptions
### **Describe the feature**
The feature adds the ability to automatically generate SQL column comments in Databricks Delta Live Tables (DLT). Specifically, it updates the `CREATE STREAMING TABLE` SQL generation to include column comments using the `COMMENT` clause, which can be pulled from the dbt column description metadata.

This allows users to define meaningful metadata for their columns directly in dbt models and have those automatically applied as SQL column comments when creating streaming tables in Databricks.

### **Describe alternatives you've considered**
1. **Manual Column Commenting**: One alternative is manually adding comments in SQL when defining the table schema. This approach, however, can be prone to inconsistencies and might result in duplicate efforts if the comments are already defined in dbt models.
2. **Post-Creation ALTER Statements**: Another alternative is adding comments using `ALTER TABLE` statements after table creation, but this would require extra steps and could lead to performance issues in large streaming tables.

### **Additional context**
Here's an example of the newly implemented code:
```jinja
{%- macro get_create_column_comment(model) -%}
  {% set node = model %}
  {% if node.columns %}
    {% set column_definitions = [] %}
    {% for column_name, column in node.columns.items() %}
      {% set data_type = column.data_type or 'STRING' %} 
      {% set column_line = column_name ~ ' ' ~ data_type %}
      {% if column.description %}
        {% set column_line = column_line ~ ' COMMENT ' ~ "'" ~ column.description ~ "'" %}
      {% endif %}
      {% do column_definitions.append(column_line) %}
    {% endfor %}
    {{ column_definitions | join(',\n    ') }}
  {% endif %}
{%- endmacro -%}

{% macro databricks__get_create_streaming_table_as_sql(relation, sql) -%}
  CREATE OR REFRESH STREAMING TABLE {{ relation }} 
    ( {{ get_create_column_comment(config.model) }} )
    -- Additional clauses like partitioning and properties
    AS {{ sql }}
{% endmacro %}
```

The code leverages dbt metadata (`node.columns`) to automatically generate column comments during table creation. This ensures consistency and reduces manual overhead for managing schema documentation.

### **Who will this benefit?**
This feature will benefit:
- Data engineers who use Databricks and dbt for building streaming tables, allowing them to include meaningful column descriptions directly from dbt models.
- Teams working with large data sets where maintaining schema documentation is critical. This ensures schema changes and documentation are kept in sync automatically.

**Example Use Case**: A team might have a dbt model that defines columns with descriptions. When they deploy this model to a Databricks Delta Live Table, the column comments would automatically be included, helping others understand the purpose and meaning of each column.

### **Are you interested in contributing this feature?**
Yes, I'm interested in contributing to this feature. Please let me know if additional steps are required to prepare it for contribution or if any changes are needed. I am happy to discuss the implementation further.

Amin-Siddique:
Resolves #793

### Description

Add support for column comments in Delta Live Tables (ST/MV) from dbt model schema descriptions 

### Checklist

- [] I have run this code in development and it appears to resolve the stated issue
- [] This PR includes tests, or tests are not required/relevant for this PR
- [] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section.

### DIFF
--- a/dbt/include/databricks/macros/relations/components/comment.sql
+++ b/dbt/include/databricks/macros/relations/components/comment.sql
@@ -1,5 +1,15 @@
-{%- macro get_create_sql_comment(comment) -%}
-{% if comment is string -%}
-  COMMENT '{{ comment }}'
-{%- endif -%}
+{%- macro get_create_column_comment(model) -%}
+  {%- if model.columns -%}
+    {%- set column_definitions = [] -%}
+    {%- for column_name, column in model.columns.items() -%}
+      {%- set data_type = column.data_type or 'STRING' -%}
+      {%- if column.description -%}
+        {%- set column_line = column_name ~ ' ' ~ data_type ~ ' COMMENT ' ~ "'" ~ column.description ~ "'" -%}
+      {%- endif -%}
+
+      {%- do column_definitions.append(column_line) -%}
+    {% endfor %}
+
+    {{ column_definitions | JOIN(',\n ') }}
+  {%- endif -%}
 {%- endmacro -%}
--- a/dbt/include/databricks/macros/relations/components/comment.sql
+++ b/dbt/include/databricks/macros/relations/components/comment.sql
@@ -1,3 +1,10 @@
+{%- macro get_create_sql_comment(comment) -%}
+{% if comment is string -%}
+  COMMENT '{{ comment }}'
+{%- endif -%}
+{%- endmacro -%}
+
+
 {%- macro get_create_column_comment(model) -%}
   {%- if model.columns -%}
     {%- set column_definitions = [] -%}
@@ -9,7 +16,6 @@
 
       {%- do column_definitions.append(column_line) -%}
     {% endfor %}
-
     {{ column_definitions | JOIN(',\n ') }}
   {%- endif -%}
 {%- endmacro -%}
--- a/dbt/include/databricks/macros/relations/components/comment.sql
+++ b/dbt/include/databricks/macros/relations/components/comment.sql
@@ -6,16 +6,48 @@
 
 
 {%- macro get_create_column_comment(model) -%}
-  {%- if model.columns -%}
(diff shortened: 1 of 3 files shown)


dbt/include/databricks/macros/relations/components/comment.sql
```diff
@@ -3,3 +3,19 @@
   COMMENT '{{ comment }}'
 {%- endif -%}
 {%- endmacro -%}
+
+
+{%- macro get_create_column_comment(model) -%}
+  {%- if model.columns -%}
+    {%- set column_definitions = [] -%}
+    {%- for column_name, column in model.columns.items() -%}
+      {%- set data_type = column.data_type or 'STRING' -%}
```
benc-db, line 12: Does this actually work?  In general we don't know the column data type, and I'm concerned that specifying as string will break when the columns are string type.

Amin-Siddique: Yes you're right, it breaks! Unfortunately, for this to work it we need to explicitly define data_types in the schema file of the respective model.

Amin-Siddique: @benc-db I have added a workaround where we initially create a temp view to infer the schema and utilize that to get the data type if undefined. Please let me know your thoughts! 😄  I have tested it in my local setup (1. with schema file and defined data_types 2. with schema file without defining any data types 3. without schema file) and works for both ST/MV

benc-db: This is an interesting approach.  I want to talk to dbt Labs about this to get their perspective, and I definitely will want to put behind a behavior flag (new tech recently introduced into dbt) because this will add some latency, but might be worth the trade off.

PR #796, opened 2024-09-17
Fix for describe column types

### LINKED ISSUE #779: Databricks truncates datatypes returned via `DESCRIBE EXTENDED` which is used by get_columns_in_relation()
### Describe the bug
Couldn't tell you the full impact of this bug but where I encountered it was while using `on_schema_change="sync_all_columns"`. 

Basically the bug led to truncated results that feed queries involved in handling alter statements when there are data type changes in a dataset. 

---
#### Current Behaviour

This is because running the below truncates the data types
```sql
DESCRIBE EXTENDED <catalog>.<schema>.<table>
```
 Truncated field example using `DESCRIBE EXTENDED`
```sql 
struct<_info:struct<fieldA:string,fieldB:string>,fieldC:bigint,fieldD:string>,... 78 more fields>
```
---
#### Requested Behaviour

Ideally dbt databricks instead uses the below to acquire that information as it does not truncate data types
```sql
select
    column_name,
    full_data_type,
    comment
from <catalog>.information_schema.columns
where table_schema = <schema> and table_name = <table>
```
---
### Steps To Reproduce

1. Have a very complex (long datatype) struct field in your dataset
2. Run any operation in `dbt-databricks` that looks up the datatype of that field via `get_columns_in_relation()`
3. You will observe the struct field you created has truncated datatype

### Expected Behaviour
1. Have a very complex (long datatype) struct field in your dataset
2. Run any operation in `dbt-databricks` that looks up the datatype of that field via `get_columns_in_relation()`
3. You will observe the struct field you created does not have a truncated datatype

### System information
```bash
Core:
  - installed: 1.8.5
  - latest:    1.8.5 - Up to date!

Plugins:
  - databricks: 1.8.5 - Up to date!
```

### Additional context
- Also causes problems when using [dbt codegen](https://github.com/dbt-labs/dbt-codegen) as it utilizes the adapter to lookup the datatypes of columns

benc-db:
Resolves #779 (for UC)

### Description

`DESCRIBE EXTENDED` truncates column types.  For generating doc site, this is ok, but for running a dbt project, it can cause issues.  Where information_schema is available, we can use it to get an un-truncated column type.  Initially I was going to hide this behind a behavior flag, since both methods can have performance issues; however, during implementation I discovered that behavior flags aren't available to adapters yet, due to the current dbt-core not having flags on the RuntimeConfig object.  So, plan is to use information_schema where available, as it is more correct.  Once behavior flags are available, we can allow users to choose, so that if one way is less performant, they can switch.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -49,7 +49,7 @@ from dbt.adapters.databricks.python_submissions import (
 from dbt.adapters.databricks.python_submissions import (
     DbtDatabricksJobClusterPythonJobHelper,
 )
-from dbt.adapters.databricks.relation import DatabricksRelation, is_hive_metastore
+from dbt.adapters.databricks.relation import DatabricksRelation
 from dbt.adapters.databricks.relation import DatabricksRelationType
 from dbt.adapters.databricks.relation import KEY_TABLE_PROVIDER
 from dbt.adapters.databricks.relation_configs.base import DatabricksRelationConfig
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -170,7 +170,7 @@ class DatabricksAdapter(SparkAdapter):
     # For now does nothing
     @property
     def _behavior_flags(self) -> List[BehaviorFlag]:
-        return [{"name": "column_types_from_information_schema", "default": False}]
+        return [{"name": "column_types_from_information_schema", "default": False}]  # type: ignore
 
     # override/overload
     def acquire_connection(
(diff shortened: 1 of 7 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -376,6 +384,36 @@ def parse_describe_extended(  # type: ignore[override]
 
     def get_columns_in_relation(  # type: ignore[override]
         self, relation: DatabricksRelation
+    ) -> List[DatabricksColumn]:
+        if (
+            # We can uncomment this once behavior flags are available to adapters
+            # self.behavior.column_types_from_information_schema and # type: ignore
+            not relation.is_hive_metastore()
```
eric-wang-1990, line 391: Is there a reason we only enable it for UC now? Should we just enable for non-UC to avoid bug?

benc-db: We only have information_schema for UC last I checked.

eric-wang-1990 (approved):

PR #808, opened 2024-09-25
Behavior: Get column info from information_schema Part I

benc-db:
Partial fix for #779

### Description

This is my second attempt at addressing the issue that describe extended can truncate complex types.  With the release of dbt-core 1.8.7, we can now process behavior flags; this PR introduces the choice of using information_schema for grabbing column information in get_columns_for_relation.  I'm hiding behind a behavior flag because given the current state of UC information_schema, we have to run repair to trust that columns of recently created or altered table will be present in the information_schema, which adds overhead.  Furthermore, this trick only works for Delta tables at this time.  It is hoped that in time the sync issue with information_schema will be solved, but in the mean time, users can use this flag when they have complex types that describe extended truncates.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -93,6 +93,15 @@ SHOW_TABLES_MACRO_NAME = "show_tables"
 SHOW_VIEWS_MACRO_NAME = "show_views"
 GET_COLUMNS_COMMENTS_MACRO_NAME = "get_columns_comments"
 
+USE_INFO_SCHEMA_FOR_COLUMNS = BehaviorFlag(
+    name="use_info_schema_for_columns",
+    default=False,
+    description=(
+        "Use info schema to gather column information to ensure complex types are not truncated."
+        "  Incurs some overhead, so disabled by default."
+    ),
+)
+
 
 @dataclass
 class DatabricksConfig(AdapterConfig):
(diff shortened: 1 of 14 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -116,26 +119,6 @@ class DatabricksConfig(AdapterConfig):
     merge_with_schema_evolution: Optional[bool] = None
 
 
-def check_not_found_error(errmsg: str) -> bool:
```
benc-db, line 119: Moved to prevent circular dependency.

dbt/adapters/databricks/impl.py
```diff
@@ -175,6 +158,19 @@ class DatabricksAdapter(SparkAdapter):
         }
     )
 
+    get_column_behavior: GetColumnsBehavior
+
+    def __init__(self, config: Any, mp_context: SpawnContext) -> None:
```
benc-db, line 163: Although I don't control the invocation of init, and thus can't pass the behaviors in, doing this during init ensures we only check the behavior flag once.  I tried to make this a little more functional but couldn't figure out how to override an existing function definition that is inherited from a parent, hence the goofy class-based strategy.

dbt/include/databricks/macros/adapters/persist_docs.sql
```diff
@@ -28,6 +28,25 @@
   {% do return(load_result('get_columns_comments').table) %}
 {% endmacro %}
 
+{% macro get_columns_comments_via_information_schema(relation) -%}
+  {% call statement('repair_table', fetch_result=False) -%}
+    REPAIR TABLE {{ relation|lower }} SYNC METADATA
```
benc-db, line 33: Ensure information_schema is up to date prior to using this method.

Issue #810, opened 2024-09-26
Set unique table suffix to allow parallel incremental execution


huangxingyi-git:
### Describe the feature
For some specific cases (eg. backfill very large amount of data), we need to execute parallel multiple `dbt run` of specific incremental(`replace_where`) model in which we pass the date (or country) as var argument.
For example, we have a model we run every day using Airflow for which we pass the a date relative to the Airflow scheduler.
FYI
https://github.com/dbt-labs/dbt-athena/pull/650/files

If we want to process by batch of N days in parallel using Airflow concurrency, we need the tmp table create by each of the dbt run to be unique. Else, you are going to end up with N insert attempting to run with the same __dbt_tmp name, creating conflict and ultimately creating failure.

### Who will this benefit?
For those who uses `repalce_where` as incremental strategy.
Example Use Case: Run the same incremental model concurrently with different `--vars` in order to parallelly insert multiple data partitions

### Are you interested in contributing this feature?
I am interested in contributing to this feature if needed.

huangxingyi-git:
How can I create PR?
i tried to push my feature branch to github and met below error
``` 
% git push origin unique-temp-table-name:huangxingyi/unique-temp-table-name

remote: Permission to databricks/dbt-databricks.git denied to huangxingyi-git.
fatal: unable to access 'https://github.com/databricks/dbt-databricks.git/': The requested URL returned error: 403```

benc-db:
This issue is going to solved more comprehensively with https://github.com/dbt-labs/dbt-core/discussions/10672; however, I'll take a look at your PR, and assuming it doesn't hurt any other use case, I'm not against taking this in the short term.

PR #811, opened 2024-09-26
Set unique temp table suffix to allow parallel incremental executions

huangxingyi-git:
Resolves #

### Description
Set unique table suffix to allow parallel incremental execution
For some specific cases (eg. backfill very large amount of data), we need to execute parallel multiple dbt run of specific incremental(replace_where) model in which we pass the date (or country) as var argument.
For example, we have a model we run every day using Airflow for which we pass the a date relative to the Airflow scheduler.
FYI
https://github.com/dbt-labs/dbt-athena/pull/650/files

If we want to process by batch of N days in parallel using Airflow concurrency, we need the tmp table create by each of the dbt run to be unique. Else, you are going to end up with N insert attempting to run with the same __dbt_tmp name, creating conflict and ultimately creating failure.

[issue](https://github.com/databricks/dbt-databricks/issues/810)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 ## dbt-databricks 1.8.7 (TBD)
 
+### Features
+
+- Add config for generating unique tmp table names for enabling parralel replace-where (thanks @huangxingyi-git!) ([811](https://github.com/databricks/dbt-databricks/pull/811))
+
 ### Fixes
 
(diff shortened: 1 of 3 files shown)


benc-db:
Apologies for the delay; I've been busy getting ready for 1.9.0 release and Coalesce next week.  Code looks straightforward, so I will just verify that it does not break functional tests.

benc-db (dismissed):

PR #815, opened 2024-10-01
Simple Iceberg support

benc-db:
### Description

Adds table_format: iceberg, to match support of other dbt-adapters.  This config sets a table or incremental materialization as iceberg compatible via UniForm.  Note: once a table is made iceberg compatible, removing the table_format will not remove the tblproperties, as we currently find doing tblproperty deletes on behalf of the user to be too fragile for most use-cases.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -66,6 +66,7 @@ from dbt.adapters.databricks.relation_configs.materialized_view import (
 from dbt.adapters.databricks.relation_configs.streaming_table import (
     StreamingTableConfig,
 )
+from dbt.adapters.databricks.relation_configs.table_format import TableFormat
 from dbt.adapters.databricks.relation_configs.tblproperties import TblPropertiesConfig
 from dbt.adapters.databricks.utils import get_first_row, handle_missing_objects
 from dbt.adapters.databricks.utils import redact_credentials
@@ -80,6 +81,7 @@ from dbt.adapters.spark.impl import SparkAdapter
 from dbt_common.behavior_flags import BehaviorFlag
 from dbt_common.utils import executor
 from dbt_common.utils.dict import AttrDict
+from dbt_common.exceptions import DbtConfigError
 
 if TYPE_CHECKING:
     from agate import Row
@@ -106,6 +108,7 @@ USE_INFO_SCHEMA_FOR_COLUMNS = BehaviorFlag(
 @dataclass
 class DatabricksConfig(AdapterConfig):
     file_format: str = "delta"
+    table_format: TableFormat = TableFormat.DEFAULT
     location_root: Optional[str] = None
     partition_by: Optional[Union[List[str], str]] = None
     clustered_by: Optional[Union[List[str], str]] = None
@@ -179,6 +182,20 @@ class DatabricksAdapter(SparkAdapter):
     @property
     def _behavior_flags(self) -> List[BehaviorFlag]:
         return [USE_INFO_SCHEMA_FOR_COLUMNS]
+
+    @available.parse(lambda *a, **k: 0)
+    def check_iceberg(self, table_format: TableFormat, file_format: str, materialized: str) -> bool:
+        if table_format == TableFormat.ICEBERG:
+            if self.compare_dbr_version(14, 3) < 0:
+                raise DbtConfigError("Iceberg support requires Databricks Runtime 14.3 or later.")
+            if file_format and file_format != "delta":
+                raise DbtConfigError("When table_format is 'iceberg', cannot set file_format.")
+            if materialized not in ("incremental", "table"):
+                raise DbtConfigError(
+                    "When table_format is 'iceberg', materialized must be 'incremental' or 'table'."
+                )
+            return True
+        return False
 
     # override/overload
     def acquire_connection(
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -82,6 +82,7 @@ from dbt_common.behavior_flags import BehaviorFlag
 from dbt_common.utils import executor
 from dbt_common.utils.dict import AttrDict
 from dbt_common.exceptions import DbtConfigError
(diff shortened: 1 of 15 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -106,6 +108,7 @@
 @dataclass
 class DatabricksConfig(AdapterConfig):
     file_format: str = "delta"
+    table_format: TableFormat = TableFormat.DEFAULT
```
benc-db, line 111: In future, we may fold file_format into table_format; adding table_format for consistency with other dbt adapters.

dbt/adapters/databricks/relation_configs/tblproperties.py
```diff
@@ -74,11 +74,19 @@ def from_relation_results(cls, results: RelationResults) -> TblPropertiesConfig:
 
     @classmethod
     def from_relation_config(cls, relation_config: RelationConfig) -> TblPropertiesConfig:
```
benc-db, line 76: This file handles iceberg settings as part of tblproperties for incremental tables.  It could maybe be a separate component, but since iceberg support is accomplished via tblproperties, including here.

dbt/include/databricks/macros/relations/components/tblproperties.sql
```diff
@@ -1,4 +1,5 @@
 {% macro get_create_sql_tblproperties(tblproperties) %}
+  {%- set _ = adapter.check_iceberg(config.get('table_format'), config.get('file_format'), model.config.materialized) -%}
```
benc-db, line 2: Here we check and fail the compilation of incremental tblproperties if we are misconfigured.  Otherwise, since this is used only for incremental, the tblproperties will already have the iceberg properties included.

dbt/include/databricks/macros/relations/tblproperties.sql
```diff
@@ -3,8 +3,12 @@
 {%- endmacro -%}
 
 {% macro databricks__tblproperties_clause() -%}
-  {%- set tblproperties = config.get('tblproperties') -%}
-  {%- if tblproperties is not none %}
+  {%- set tblproperties = config.get('tblproperties', {}) -%}
+  {%- set is_iceberg = adapter.check_iceberg(config.get('table_format'), config.get('file_format'), model.config.materialized) -%}
+  {%- if is_iceberg -%}
```
benc-db, line 8: This is used by table/view (but not incremental) materialization.  We will fail compilation if the prereqs are violated, but if they are not, and iceberg is set, add the iceberg properties to the dictionary of tblproperties.

dbt/include/databricks/macros/relations/tblproperties.sql
```diff
@@ -15,6 +19,10 @@
 
 {% macro apply_tblproperties(relation, tblproperties) -%}
   {% if tblproperties %}
+    {%- set is_iceberg = adapter.check_iceberg(config.get('table_format'), config.get('file_format'), model.config.materialized) -%}
```
benc-db, line 22: Same as above, but this statement is used to alter tables constructed in other ways, such as via python models.

tests/functional/adapter/iceberg/test_iceberg_support.py
```diff
@@ -0,0 +1,35 @@
...
+        assert len(run_results) == 3
+
+
+@pytest.mark.skip_profile("databricks_cluster")
+class TestIcebergSwap:
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"first_model.sql": fixtures.basic_view}
+
+    def test_iceberg_swaps(self, project):
```
benc-db, line 28: Test that we can replace a view with iceberg table, and swap to incrementing that table with incremental.  If tblproperties weren't so hard to manage, we would also remove iceberg when the user removes the table_format config; however, here we do not.  Will need to address when I write the docs.

tests/unit/macros/relations/test_table_macros.py
```diff
@@ -178,6 +198,20 @@ def test_macros_create_table_as_tblproperties(self, config, template_bundle):
 
         assert sql == expected
 
+    def test_macros_create_table_as_tblproperties_with_iceberg(self, config, template_bundle):
+        config["tblproperties"] = {"delta.appendOnly": "true"}
+        template_bundle.context["adapter"].check_iceberg.return_value = True
+        sql = self.render_create_table_as(template_bundle)
+
+        expected = (
+            f"create or replace table {template_bundle.relation} "
```
benc-db, line 207: Test that we smoothly incorporate iceberg properties with other tblproperties.

dbt/include/databricks/macros/relations/tblproperties.sql
```diff
@@ -3,8 +3,12 @@
 {%- endmacro -%}
 
 {% macro databricks__tblproperties_clause() -%}
-  {%- set tblproperties = config.get('tblproperties') -%}
-  {%- if tblproperties is not none %}
+  {%- set tblproperties = config.get('tblproperties', {}) -%}
+  {%- set is_iceberg = adapter.check_iceberg(config.get('table_format'), config.get('file_format'), model.config.materialized) -%}
+  {%- if is_iceberg -%}
+    {%- set _ = tblproperties.update({'delta.enableIcebergCompatV2': 'true', 'delta.universalFormat.enabledFormats': 'iceberg'}) -%}
```
eric-wang-1990, line 9: It seems the update logic is duplicate, I'm not sure if it is possible but can it be extracted to a central place maybe in the adapter.updateTblProperties()?

eric-wang-1990 (commented):
It seems we need to modify many macro statements to make sure they're all covered, is there a way to know the PR has cover all the places?

dbt/include/databricks/macros/relations/tblproperties.sql
benc-db: Good call, not sure why I decided the iceberg check should be in the adapter but not updating tblproperties.

eric-wang-1990 (approved):
LGTM, I do not have context over some of the tests but functionwise looks good.

PR #822, opened 2024-10-10
Update _wait_for_cluster_to_start func to check for libraries statuses

### LINKED ISSUE #821: Python models don't wait for all-purpose-cluster to be fully ready
https://github.com/databricks/dbt-databricks/blob/38aefbf86371341f09f39b2ae93224ba2f21270d/dbt/adapters/databricks/python_submissions.py#L404

### Describe the bug
When executing python models in submission_method: all_purpose_cluster, the dbt-databricks connector is not recognizing when the cluster is still installing libraries. This produces python models to fail intermittently where libs are needed.

### Steps To Reproduce
1. Use an All Purpose Cluster running 14.3LTS
2. Configure python models to run in all_purpose_cluster mode
3. Create a python model that use a library that requires installation and configure it in the cluster created in 1.
4. Execute the python model when the All Purpose Cluster is terminated

### Expected behavior
Python models waiting until libraries configured in the cluster are installed
Screenshot 2024-10-10 at 00 15 25

Based on [this](https://github.com/databricks/dbt-databricks/blob/38aefbf86371341f09f39b2ae93224ba2f21270d/dbt/adapters/databricks/python_submissions.py#L404) function, it waits until the cluster is in a `running` state, but this doesn't mean that the cluster is ready to use as shown in the screenshot.

### System information
```
1.8.3
```

psivash:
Resolves #821 

### Description
Enhance `_wait_for_cluster_to_start` to wait for library installation

Previously, the `_wait_for_cluster_to_start` function only checked if the cluster state was 'running'. This update adds a check to ensure all cluster libraries have been successfully installed (or skipped or restored) before proceeding. This ensures the cluster is fully ready for operations that depend on these libraries.

- Added `get_cluster_libraries_status` to fetch the status of cluster libraries.
- Updated `_wait_for_cluster_to_start` to wait until all libraries are in valid statuses ('INSTALLED', 'RESTORED', 'SKIPPED').
### Checklist

- [] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section.

### DIFF
--- a/dbt/adapters/databricks/python_submissions.py
+++ b/dbt/adapters/databricks/python_submissions.py
@@ -379,6 +379,20 @@ class DBContext:
         json_response = response.json()
         return json_response
 
+    def get_cluster_libraries_status(self) -> Dict:
+        # https://docs.databricks.com/api/workspace/libraries/clusterstatus
+
+        response = self.session.get(
+            f"https://{self.host}/api/2.0/libraries/cluster-status",
+            headers=self.extra_headers,
+            json={"cluster_id": self.cluster_id},
+        )
+        if response.status_code != 200:
+            raise DbtRuntimeError(f"Error getting status of libraries of a cluster.\n {response.content!r}")
+
+        json_response = response.json()
+        return json_response
+
     def start_cluster(self) -> None:
         """Send the start command and poll for the cluster status until it shows "Running"
 
@@ -406,6 +420,7 @@ class DBContext:
         logger.info("Waiting for cluster to be ready")
 
         MAX_CLUSTER_START_TIME = 900
+        LIBRARY_VALID_STATUSES = {"INSTALLED", "RESTORED", "SKIPPED"}
         start_time = time.time()
 
         def get_elapsed() -> float:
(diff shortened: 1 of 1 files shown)


dbt/adapters/databricks/python_submissions.py
```diff
@@ -414,9 +429,11 @@ def get_elapsed() -> float:
         while get_elapsed() < MAX_CLUSTER_START_TIME:
             status_response = self.get_cluster_status()
             if str(status_response.get("state")).lower() == "running":
-                return
-            else:
-                time.sleep(5)
+                libraries_status_response = self.get_cluster_libraries_status()
```
benc-db, line 432: I'm surprised this isn't done as part of 'STARTING'.  Does every run need to wait for library installs, or just those that specify additional libraries?

psivash: Good question. I _assume_ that basic default libraries are a part of 'STARTING' indeed and only the libraries that are added explicitly (the ones exposed via Libraries API) are not and need extra time so that python models do work correctly.

benc-db: If that's the case, let's take this branch only if additional libraries have been specified.

benc-db (changes requested):
Please see failing unit tests.  Also, fyi, I'm going to be merging branch 1.9.latest into main today, so we'll probably need to rebase after that.

PR #823, opened 2024-10-11
Behavior for external path

### LINKED ISSUE #812: Add behavior to compute external path from root/catalog/schema/identifier
### Describe the feature
Provide a mechanism so that tables are written to external paths that match the structure in Databricks

Asked about the possibility of encoding this in yaml files with variables, not sure if this is currently possible

### Additional context
The lack of this is currently a big pain in testing, where sometimes UC complains that a table already exists at the location (since we only use table identifier today).

### Who will this benefit?
Anyone that uses external locations and does non-trivial work

benc-db:
Resolves #812 

### Description

Adding config to have dbt-databricks add catalog and schema information to the path of external tables, to reduce the likelihood of conflicts arising from tables with the same name.  This is a huge pain in testing, but not sure if others hit this in the wild, hence why default is to not do this behavior.  We probably should have always been doing this, but it could potentially break existing workflows, another reason to default to False.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -112,6 +112,7 @@ class DatabricksConfig(AdapterConfig):
     file_format: str = "delta"
     table_format: TableFormat = TableFormat.DEFAULT
     location_root: Optional[str] = None
+    include_full_name_in_path: bool = False
     partition_by: Optional[Union[List[str], str]] = None
     clustered_by: Optional[Union[List[str], str]] = None
     liquid_clustered_by: Optional[Union[List[str], str]] = None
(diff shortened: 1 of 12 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -112,6 +112,7 @@ class DatabricksConfig(AdapterConfig):
     file_format: str = "delta"
     table_format: TableFormat = TableFormat.DEFAULT
     location_root: Optional[str] = None
+    include_full_name_in_path: bool = False
```
jackyhu-db, line 115: can it be moved to the last argument to be backward compatible if someone calls w/o named arguments

jackyhu-db (approved):

dbt/adapters/databricks/impl.py
benc-db: Normally I would agree with you, however...

Configs are never invoked with named arguments, as they are not actually exposed to users.  Instead, users provide config via yaml, which dbt bundles together into a generic config object, which gets cast for type-checking purposes.  This process is generic, i.e. dbt is also not invoking DatabricksConfig with named arguments.  The primary reason to add fields here is to get code completion support, since basically all config provided by user (through yaml file) ends up available on the config object, and if you add the field here, then your IDE knows it will be on the object.  Given that, it is more valuable to keep config fields that are related close together than to worry about ordering.

PR #827, opened 2024-10-14
Bump python sdk version

eric-wang-1990:
Upgrade python sdk version from 0.17.0 to 0.36.0
Create DatabricksCredentialManager to managed credentials, under the hood use Config from databricks.sdk.core to manage the auth part which comes from the Python SDK.
For dbt we support 4 different auth mode:

- token : When user explicitly passed in a **token=xxxx**
- external_browser: When user pass in **auth_type=oauth** with **clientid=xxx** and no client secret
- oauth-m2m: When user pass in **clientid=xxx** and **clientsecret=yyy**
- azure-client-secret: When user pass in **azure_client_id=xxx** and **azure_client_secret=yyy** and your host is an azure host

For oauth-m2m and azure-client-secret we set a preferred auth sequence based on if the clientSecret starts with "dose", so we can reduce the number of false trying as much as possible.

Remove the set/get sharedPassword part, since the python sdk already handles that, but with one defect which I already raise a PR against: https://github.com/databricks/databricks-sdk-py/pull/823

With this change one thing that will no longer work is if customer are using non-databricks OAuth client for U2M case it will not work, since the python sdk does not support modify [redirectUrl](https://github.com/databricks/databricks-sdk-py/blob/ee6e70a1e3a38b465405d41f39f4447ad7dd3090/databricks/sdk/credentials_provider.py#L204C5-L204C17)/[scopes](https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/oauth.py#L467).

### Checklist


dbt/adapters/databricks/credentials.py
```diff
@@ -241,172 +240,150 @@ def extract_cluster_id(cls, http_path: str) -> Optional[str]:
...
-            if isinstance(in_provider, m2m_auth) or isinstance(in_provider, token_auth):
-                self._credentials_provider = in_provider.as_dict()
-            return in_provider
-
-        provider: TCredentialProvider
-        # dbt will spin up multiple threads. This has to be sync. So lock here
-        self._lock.acquire()
-        try:
+        assert self._credentials_manager is not None, "Credentials manager is not set."
+        return self._credentials_manager
```
jackyhu-db, line 246: is lock still required?

dbt/adapters/databricks/credentials.py
```diff
@@ -241,172 +240,150 @@ def extract_cluster_id(cls, http_path: str) -> Optional[str]:
...
+                auth_methods = {
+                    "oauth-m2m": self.authenticate_with_oauth_m2m,
+                    "azure-client-secret": self.authenticate_with_azure_client_secret,
+                    "external-browser": self.authenticate_with_external_browser
+                }
+
+                auth_type = (
+                    "external-browser" if not self.client_secret
+                    # if the client_secret starts with "dose" then it's likely using oauth-m2m
+                    else "oauth-m2m" if self.client_secret.startswith("dose")
```
jackyhu-db, line 337: can it wrapped in a method like `is_databricks_m2m`? And this checks still a little bit tricky, do we really want to infer it from secret format? Can it be explicitly specified by the config to decide whether azure or databricks m2m should be used?

jackyhu-db:
Thanks @eric-wang-1990 for updating this. Can your rename the PR title to make it more readable?

dbt/adapters/databricks/credentials.py
benc-db: +1 to what Jacky said.  In design review we decided to not depend on secret format, but instead to a.) reference if a config flag is set to tell us to use AAD, and b.) if we fail with Databricks OAuth but succeed with AAD, emit a warning telling them to set the config flag.

benc-db (changes requested):
Please remove dependence on secret format, per internal design meeting.

dbt/adapters/databricks/credentials.py
```diff
@@ -241,172 +250,174 @@ def extract_cluster_id(cls, http_path: str) -> Optional[str]:
...
 
-        finally:
-            self._lock.release()
+class BearerAuth(AuthBase):
+    """This mix-in is passed to our requests Session to explicitly
+    use the bearer authentication method.
 
-    def set_sharded_password(self, service_name: str, username: str, password: str) -> None:
-        max_size = MAX_NT_PASSWORD_SIZE
+    Without this, a local .netrc file in the user's home directory
```
eric-wang-1990, line 263: Is it safe to remove it seems databricks cli will not use .netrc anymore? https://docs.databricks.com/en/dev-tools/cli/profiles.html @benc-db @jackyhu-db

dbt/adapters/databricks/credentials.py
eric-wang-1990: I am just honoring the existing behavior. If there are multiple connections using the same DatabricksCredentials instance, there is a shared resource _config which can have concurrent write. @benc-db I think you introduce the original lock, is it still needed?

dbt/adapters/databricks/credentials.py
jackyhu-db: I think `.netrc` is used by python `requests` which might be used by databricks-sdk code as well?

dbt/adapters/databricks/credentials.py
benc-db: I don't think so, but not totally sure.  I think I brought this lock over from somewhere else (from code Ray wrote, which was probably cribbed in part from dbt-spark code).

dbt/adapters/databricks/credentials.py
benc-db: yeah, the .netrc thing bit us in the past.

benc-db:
Looks like something is off in the python path, with those E2E tests failing.

stevenayers-bge:
Hi @benc-db & @eric-wang-1990, can this PR be applied to versions older than 1.10 too? At the moment, I have to have dbt as a separate repo purely to work around the conflict with `databricks-connect==15.4.2`.

I'm using `dbt-databricks<1.80` right now, but I can upgrade to `1.8.x` if you don't want to go that far back.

dbt/adapters/databricks/credentials.py
```diff
@@ -238,172 +245,174 @@ def extract_cluster_id(cls, http_path: str) -> Optional[str]:
...
+                self._config: Optional[Config] = None
+            if self._config is not None:
+                return
+
+            if self.token:
+                self._config = self.authenticate_with_pat()
+            elif self.azure_client_id and self.azure_client_secret:
+                self._config = self.authenticate_with_azure_client_secret()
+            elif not self.client_secret:
+                self._config = self.authenticate_with_external_browser()
```
ghjklw, line 352: I love this, it will make it so much easier to add new authentication types!

I wonder if that could be made even simpler by just having a common generic method to build the Databricks SDK for all authentication types except the legacy ones.

The logic would then be reversed: we first check whether the authentication method used is legacy and otherwise just run something like: `self._config = self.databricks_config`
```python
    @property
    def databricks_config(self) -> Config:
        return Config(
            host=self.host,
            client_id=self.client_id,
            client_secret=self.client_secret,
            auth_type=self.auth_type,
            azure_client_id=self.azure_client_id,
            azure_client_secret=self.azure_client_secret,
            token=self.token,
        )
```

I believe the `databricks.sdk.config.Config` class would handle the `None` values just fine by ignoring them. A major benefit of this approach, is that it would then be straightforward to pass through all of the attributes and thus enabling any possible authentication options.
It would also be easier to document and less confusing as you could just refer to Databricks' Unified Authentication / Python SDK authentication documentation. In addition it could also provide more flexibility by exposing additional configuration options.

For reference here is the full list of these attributes as of now:

- `host`
- `account_id`
- `token`
- `username`
- `password`
- `client_id`
- `client_secret`
- `profile`
- `config_file`
- `google_service_account`
- `google_credentials`
- `azure_workspace_resource_id`
- `azure_use_msi`
- `azure_client_secret`
- `azure_client_id`
- `azure_tenant_id`
- `azure_environment`
- `databricks_cli_path`
- `auth_type`
- `cluster_id`
- `warehouse_id`
- `serverless_compute_id`
- `skip_verify`
- `http_timeout_seconds`
- `debug_truncate_bytes`
- `debug_headers`
- `rate_limit`
- `retry_timeout_seconds`
- `metadata_service_url`
- `max_connection_pools`
- `max_connections_per_pool`
- `databricks_environment`

dbt/adapters/databricks/api_client.py
```diff
@@ -557,8 +556,7 @@ def create(
         http_headers = credentials.get_all_http_headers(
             connection_parameters.pop("http_headers", {})
         )
-        credentials_provider = credentials.authenticate(None)
-        header_factory = credentials_provider(None)  # type: ignore
+        header_factory = credentials.authenticate().credentials_provider()  # type: ignore
```
benc-db, line 559: Do we still need the `# type: ignore`?

dbt/adapters/databricks/credentials.py
benc-db: @ghjklw I think we will get there eventually; when we did internal design review for this migration path, we wanted to make sure we didn't break any users.  At the time, we didn't have behavior flags, so we took this middle ground to achieve that goal.  But it would be great to get to the point of unified client auth.  Part of what has come out of this effort is getting some more of stuff that we had as custom code in dbt-databricks into the SDK auth stack :).  At this point I'm focused on helping Eric finish off this PR, but unified client auth is probably in the future.

benc-db (approved):
Looks good.

benc-db (approved):

PR #830, opened 2024-10-17
Refactor python config handling

benc-db:
### Description

Refactoring python submission components to be a little better factored.  There's still definitely more cleanup to do, but my goal was basically to get the config manipulation and submission into the 3 core ways we actually submit (Workflow, Job/Run, and Command) so that we can maybe one day move to those being the submission helpers, with compute not being named in the submission method.  For now we have to keep the top-level submission helpers so that we don't break any existing users.

Discussed with original author Kyle on renaming 'workflow_job_config' to 'python_job_config' so as to not confuse users when using this config without the workflow submission method.

### Checklist


benc-db:
Note: unit tests need significant reworking, but I wanted to get the core code in front of people, while I work on that.

docs/workflow-job-submission.md
```diff
@@ -109,57 +99,61 @@ dbt will generate a name based on the catalog, schema, and model identifier.
...
+                },
+            },
+          ]
 ```
 
 #### Grants
 
 You might want to give certain users or teams access to run your workflows outside of
-dbt in an ad hoc way. You can define those permissions in the `workflow_job_config.grants`.
+dbt in an ad hoc way. You can define those permissions in the `python_job_config.grants`.
```
benc-db, line 153: I changed this with find/replace, and VS decided it needed to reformat the whole file :/

benc-db:
Moved to using pydantic for managing config verification/extraction, added factory methods (`create`) to simplify init methods.  I don't like having complex logic in inits, because it makes unit testing with mocks waaay harder, but adding the factory methods make it much cleaner to build the object graphs...really wish I could pass the dependencies into the helpers, but we don't own the instantiation (happens inside dbt).  Will move onto adding/updating unit tests now.

@jackyhu-db @eric-wang-1990 @kdazzle apologies for complex refactor, but python_submission.py was already a mess, and Kyle's feature added a bunch of feature support that pushed that module beyond the limits of what I can mentally process.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -1,70 +1,175 @@
...
 
-    def _update_with_acls(self, cluster_dict: dict) -> dict:
-        acl = self.parsed_model["config"].get("access_control_list", None)
-        if acl:
-            cluster_dict.update({"access_control_list": acl})
-        return cluster_dict
+    def __init__(
+        self,
+        api_client: DatabricksApiClient,
+        job_grants: Dict[str, List[Dict[str, Any]]],
```
jackyhu-db, line 140: If it is builder, should `job_grants` and `acls` be passed to build method rather than constructor

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -97,39 +202,20 @@ def _build_job_permissions(self) -> List[Dict[str, Any]]:
...
-            "task_key": "inner_notebook",
-            "notebook_task": {
-                "notebook_path": path,
-            },
-        }
-        job_spec.update(cluster_spec)  # updates 'new_cluster' config
 
-        # PYPI packages
-        packages = self.parsed_model["config"].get("packages", [])
+class PythonLibraryConfigurer:
```
jackyhu-db, line 208: nit: is this abstract class? Should it inherit from `ABC`

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -138,189 +224,318 @@ def _submit_job(self, path: str, cluster_spec: dict) -> str:
...
+        cluster_spec: Dict[str, Any],
+    ) -> "PythonJobConfigCompiler":
+        permission_builder = PythonPermissionBuilder.create(api_client, parsed_model)
+        packages = parsed_model.config.packages
+        index_url = parsed_model.config.index_url
+        additional_libraries = parsed_model.config.additional_libs
+        library_config = PythonLibraryConfigurer.get_library_config(
+            packages, index_url, additional_libraries
+        )
+        cluster_spec.update(library_config)
```
jackyhu-db, line 263: input argument `cluster_spec` should be immutable?

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -138,189 +224,318 @@ def _submit_job(self, path: str, cluster_spec: dict) -> str:
...
+            )
+        else:
+            return PythonCommandSubmitter(self.api_client, self.tracker, self.cluster_id or "")
 
-    def check_credentials(self) -> None:
-        super().check_credentials()
+    @override
+    def validate_config(self) -> None:
+        config = self.parsed_model.config
+        self.cluster_id = config.cluster_id or self.credentials.extract_cluster_id(
```
jackyhu-db, line 382: It changed `cluster_id`, `validate` should be immutable?

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -138,189 +224,318 @@ def _submit_job(self, path: str, cluster_spec: dict) -> str:
...
         }
-        notebook_task.update(cluster_settings)
-        notebook_task.update(self.additional_task_settings)
+        notebook_task.update(self.task_settings)
+
+        self.workflow_spec["tasks"] = [notebook_task] + self.post_hook_tasks
+        return self.workflow_spec, self.existing_job_id
+
+
+class PythonWorkflowCreater:
```
jackyhu-db, line 466: PythonWorkflowCreator?

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -138,189 +224,318 @@ def _submit_job(self, path: str, cluster_spec: dict) -> str:
...
+            if len(response_jobs) > 1:
+                raise DbtRuntimeError(
+                    f"Multiple jobs found with name {workflow_name}. Use a"
+                    " unique job name or specify the `existing_job_id` in the python_job_config."
+                )
 
-    def _submit_through_workflow(self, compiled_code: str, workflow_spec: Dict[str, Any]) -> None:
-        self.api_client.workspace.create_python_model_dir(self.catalog, self.schema)
-        self.api_client.workspace.upload_notebook(self.notebook_path, compiled_code)
+            if len(response_jobs) == 1:
```
jackyhu-db, line 490: nit: `elif`

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -1,70 +1,175 @@
...
+    ) -> None:
+        self.api_client = api_client
+        self.catalog = catalog
+        self.schema = schema
+        self.identifier = identifier
+
+    @staticmethod
+    def create(
+        api_client: DatabricksApiClient, parsed_model: ParsedPythonModel
+    ) -> "PythonNotebookUploader":
```
jackyhu-db, line 109: why not pass `parsed_model` as param of the constructor

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -1,70 +1,175 @@
...
+            parsed_model.schema_,
+            parsed_model.identifier,
         )
 
-        self.job_grants: Dict[str, List[Dict[str, Any]]] = self.workflow_spec.pop("grants", {})
+    def upload(self, compiled_code: str) -> str:
+        """Upload the compiled code to the Databricks workspace."""
+        workdir = self.api_client.workspace.create_python_model_dir(self.catalog, self.schema)
+        file_path = f"{workdir}{self.identifier}"
+        self.api_client.workspace.upload_notebook(file_path, compiled_code)
```
jackyhu-db, line 121: do you need to rollback (delete pytho_model_dir) when `upload_notebook` fails?

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -138,189 +224,318 @@ def _submit_job(self, path: str, cluster_spec: dict) -> str:
...
+    def submit(self, compiled_code: str) -> None:
+        file_path = self.uploader.upload(compiled_code)
+        job_config = self.config_compiler.compile(file_path)
 
         # submit job
-        run_id = self._submit_job(file_path, cluster_spec)
+        run_id = self.api_client.job_runs.submit(
+            job_config.run_name, job_config.job_spec, **job_config.additional_job_config
+        )
+        self.tracker.insert_run_id(run_id)
```
jackyhu-db, line 334: should it delete the file if submit_job fails?

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: Arguably yes.  In this case the builder is one time use, so functionally it doesn't matter, it just moves the process of extracting the job grants into the create method.  However, stylistically, I can see your point.

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: Not abstract, just a static method holder.  The real question is whether to move the function out to the module layer, since it doesn't really need a class to hold it (since it has no state).  I'm kind of on the fence about bare functions, but maybe I should use it more often in cases like this.

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: yes, that is definitely better.  When I first looked, I didn't see a single line dict merge that didn't mutate, but I just found {**dict1, **dict2}, so I will switch to that.

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: I'm on the fence about this; users use the uploaded notebook as part of debugging in case of failure.  They can find the same information in the dbt.log, but along with a lot of other information that may be distracting.

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: I was very much on the fence about the proper spelling, happy to change :)

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: so, elif is not needed, but I guess as a matter of style, it makes sense since it's mutually exclusive with the condition above.

jackyhu-db (dismissed):

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: this case is simple enough that you're probably right.  I was trying to move construction logic out into create methods, but this is so simple that it isn't justified, since there is no conditional logic in it.

dbt/adapters/databricks/python_models/python_submissions.py
benc-db: I don't think we want to delete the directory because there could be other python model notebooks in the same directory

PR #836, opened 2024-10-24
Propose fix for connection testing issues in Cloud.

VersusFacit:
### Description

dbt cloud has a feature for testing connections. It and other features are currently failing due to adapter behavior flags not being registered for those runs. 

If we look at [these lines](https://github.com/databricks/dbt-databricks/blob/84fc024fb8e881ef5dc59d09a56bfcba99842091/dbt/adapters/databricks/impl.py#L183-L188), we see the use of a flag in the adapter class init.  This was par t of #808 

```
    def __init__(self, config: Any, mp_context: SpawnContext) -> None:
        super().__init__(config, mp_context)
        if self.behavior.use_info_schema_for_columns.no_warn:  # type: ignore[attr-defined]
            self.get_column_behavior = GetColumnsByInformationSchema()
        else:
            self.get_column_behavior = GetColumnsByDescribe()
```

This is not guaranteed to succeed. So we can use exception handling to ensure the control flow for other subtasks where the flags won't be registered are handled gracefully.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -85,6 +85,7 @@ from dbt.adapters.spark.impl import SparkAdapter
 from dbt_common.behavior_flags import BehaviorFlag
 from dbt_common.utils import executor
 from dbt_common.utils.dict import AttrDict
+from dbt_common.exceptions import CompilationError
 from dbt_common.exceptions import DbtConfigError
 from dbt_common.exceptions import DbtInternalError
 from dbt_common.contracts.config.base import BaseConfig
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -182,10 +182,15 @@ class DatabricksAdapter(SparkAdapter):
 
     def __init__(self, config: Any, mp_context: SpawnContext) -> None:
         super().__init__(config, mp_context)
-        if self.behavior.use_info_schema_for_columns.no_warn:  # type: ignore[attr-defined]
-            self.get_column_behavior = GetColumnsByInformationSchema()
-        else:
-            self.get_column_behavior = GetColumnsByDescribe()
+
+        # dbt doesn't propogate flags for certain workflows like dbt debug so this requires
+        # an additional guard
+        self.get_column_behavior = GetColumnsByDescribe()
+        try:
+            if self.behavior.use_info_schema_for_columns.no_warn:  # type: ignore[attr-defined]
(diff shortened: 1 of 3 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -182,10 +182,15 @@ class DatabricksAdapter(SparkAdapter):
...
-        else:
-            self.get_column_behavior = GetColumnsByDescribe()
+
+        # dbt doesn't propogate flags for certain workflows like dbt debug so this requires
+        # an additional guard
+        self.get_column_behavior = GetColumnsByDescribe()
+        try:
+            if self.behavior.use_info_schema_for_columns.no_warn:  # type: ignore[attr-defined]
+                self.get_column_behavior = GetColumnsByInformationSchema()
+        except CompilationError:
```
VersusFacit, line 192: we didn't expect behavior flags to be used this early in the runtime, so we need an extra check for the use here that protects against situations where flags don't get registered

mikealfare: I would move this condition into `get_columns_in_relation`. It looks like that's the only thing that references `self.get_column"_behavior` (at least in the adapter itself). And if we're at that point, then we definitely have `flags` from `project.yml`. This also aligns with pushing behavior flags as close to their relevant code as possible. While column metadata is a part of the vast majority of runs, we should still strive to keep the flag close to the code that it impacts, that way when we remove the flag we have a better idea of what we're affecting.

benc-db: Why is this not defined by init time?  In general I try to follow the pattern of https://blog.ploeh.dk/2011/07/28/CompositionRoot/, but man does dbt make that difficult.  I would strongly prefer not to delay selecting behavior until get_columns_in_relation.  I am fine with @VersusFacit's suggestion, but don't really understand why the initialization behavior is not the same for all paths.

benc-db (dismissed):

dbt/adapters/databricks/impl.py
benc-db, line 192: Need to import CompilationError

VersusFacit: We took this thread to a more rapid response forum but for posterity project flags come from the dbt_project.yml which cloud doesn't always have in scope for certain workflows. So for that situation, we pass a dummy dbt_project.yml (empty of flags) and use profiles.yml / the cloud equivalent to, say, connection test.

PR #839, opened 2024-10-29
Update split part to use safe operation

benc-db:
### Description

Hit issue with internal testing; turns out that the split_part macro in dbt-spark uses unsafe indexing (which is a totally reasonable choice), but the adapter tests expect it to be safe to index beyond the number of matched parts.  Switching to 'get' fixes the issue.

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/utils/split_part.sql
+++ b/dbt/include/databricks/macros/utils/split_part.sql
@@ -1,0 +1,43 @@
+{% macro databricks__split_part(string_text, delimiter_text, part_number) %}
+
+    {% set delimiter_expr %}
+
+        -- escape if starts with a special character
+        case when regexp_extract({{ delimiter_text }}, '([^A-Za-z0-9])(.*)', 1) != '_'
+            then concat('\\', {{ delimiter_text }})
+            else {{ delimiter_text }} end
+
+    {% endset %}
+
+    {% if part_number >= 0 %}
+
+        {% set split_part_expr %}
+
+        get(split(
+            {{ string_text }},
+            {{ delimiter_expr }}
+            ), {{ part_number - 1 if part_number > 0 else part_number }})
+
+        {% endset %}
+
+    {% else %}
+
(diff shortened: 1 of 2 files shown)


dbt/include/databricks/macros/utils/split_part.sql
```diff
@@ -0,0 +1,43 @@
...
+        {% set split_part_expr %}
+
+        get(split(
+            {{ string_text }},
+            {{ delimiter_expr }}
+            ), 
+                length({{ string_text }})
+                - length(
+                    replace({{ string_text }},  {{ delimiter_text }}, '')
+                ) + 1 + {{ part_number }}
```
jackyhu-db, line 34: will it be negative?

jackyhu-db:
is this a new file? If yes, where is it used?

benc-db:
> is this a new file? If yes, where is it used?

Hi Jacky.  This is basically a copy-paste from the spark macro so that I can replace array indexing with the 'get' function.  As such, I haven't actually validated the regex behavior, which apparently works since it passes the dbt-core tests, and no one is complaining about that.  I'm just making it so that if you ask for index 3, and there are only 2 matches, you get NULL instead of raising an exception, since that is what is specified in the adapter tests from dbt.

benc-db:
> > is this a new file? If yes, where is it used?
> 
> Hi Jacky. This is basically a copy-paste from the spark macro so that I can replace array indexing with the 'get' function. As such, I haven't actually validated the regex behavior, which apparently works since it passes the dbt-core tests, and no one is complaining about that. I'm just making it so that if you ask for index 3, and there are only 2 matches, you get NULL instead of raising an exception, since that is what is specified in the adapter tests from dbt.

I spoke with @mikealfare about bringing the change into dbt-spark, but he said that since this function is only in recent versions of spark, and unlike us, he can't presume a particular spark version, that for now the fix should be in dbt-databricks.

jackyhu-db (approved):

PR #840, opened 2024-10-30
Change merge alias defaults

### LINKED ISSUE #841: 1.9 renamed internal aliases `DBT_INTERNAL_DEST / DBT_INTERNAL_SOURCE` to `tgt / src` which is causing errors in some runs
### Describe the bug
We renamed "well-known" internal aliases `DBT_INTERNAL_X` here: https://github.com/databricks/dbt-databricks/blob/fec8b30985a2f7007c8ff04866623d7a37238236/dbt/include/databricks/macros/materializations/incremental/strategies.sql#L75-L76

Some users have macros that work off the assumption that those internal aliases would be kept to `DBT_INTERNAL_X` - this generates an invalid SQL like:

```sql
    merge
    into
        `db`.`schema`.`tbl` as tgt
    using
        `tbl__dbt_tmp` as src
    on
        
            DBT_INTERNAL_DEST.updated_at >= '2024-10-29'
            AND DBT_INTERNAL_DEST.updated_at <= '2024-10-29'
        
    and 
                  src.updated_at <=> tgt.updated_at
              
    and 
                  src.id <=> tgt.id

...
              
    when matched
...
```

as `DBT_INTERNAL_X` alias does not exist.

### Steps To Reproduce

### Expected behavior
Worth renaming it back to `DBT_INTERNAL_X`.

### System information
```
<output goes here>
```

benc-db:
Resolves #841 

### Description

During canary usage of the 1.9.0b2, we discovered that some users had overwritten some existing macros to rely on the previous aliases (since they are used through dbt to interact with predicates).  This PR changes the defaults to the previous aliases.

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -72,8 +72,8 @@ select {{source_cols_csv}} from {{ source_relation }}
 {% macro databricks__get_merge_sql(target, source, unique_key, dest_columns, incremental_predicates) %}
   {# need dest_columns for merge_exclude_columns, default to use "*" #}
 
-  {%- set target_alias = config.get('target_alias', 'tgt') -%}
-  {%- set source_alias = config.get('source_alias', 'src') -%}
+  {%- set target_alias = config.get('target_alias', 'DBT_INTERNAL_TARGET') -%}
+  {%- set source_alias = config.get('source_alias', 'DBT_INTERNAL_SOURCE') -%}
 
   {%- set predicates = [] if incremental_predicates is none else [] + incremental_predicates -%}
   {%- set dest_columns = adapter.get_columns_in_relation(target) -%}
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -146,7 +146,7 @@ select {{source_cols_csv}} from {{ source_relation }}
     {%- endif %}
 {% endmacro %}
 
-{% macro get_merge_update_set(update_columns, on_schema_change, source_columns, source_alias='src') %}
+{% macro get_merge_update_set(update_columns, on_schema_change, source_columns, source_alias='DBT_INTERNAL_SOURCE') %}
   {%- if update_columns -%}
     {%- for column_name in update_columns -%}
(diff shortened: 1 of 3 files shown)


benc-db:
Reminder to Ben, need to update the docs on the dbt site.

@mi-volodin for viz.

mi-volodin:
@benc-db thanks for fixing this. If there are any further issues with these merge capabilities - you can ping me, I can take over.

However, I didn't get the justification of a "bug" statement in #841. 
Some users assumed the aliases to be `DBT_INTERNAL_X`, but to the best of my knowledge, these aliases are not fixed in any design document for dbt.

If that is correct, then making an assumption and binding with the internal implementation - is an abstraction leak. With exactly such consequences. So it looks not like the bug for me, because internal implementation should not be constrained by user expectations. Thus it is but more as the feature request to tolerate specific user-side design desicions. 

Just out of curiosity - have you considered the opposite option? I mean to actually motivate such users to migrate to specifically defined aliases as part of the upgrade?

benc-db:
> @benc-db thanks for fixing this. If there are any further issues with these merge capabilities - you can ping me, I can take over.
> 
> However, I didn't get the justification of a "bug" statement in #841. Some users assumed the aliases to be `DBT_INTERNAL_X`, but to the best of my knowledge, these aliases are not fixed in any design document for dbt.
> 
> If that is correct, then making an assumption and binding with the internal implementation - is an abstraction leak. With exactly such consequences. So it looks not like the bug for me, because internal implementation should not be constrained by user expectations. It sounds more like the feature request to tolerate specific user-side design decisions.
> 
> Just out of curiosity - have you considered the opposite option? I mean to actually motivate such users to migrate to specifically defined aliases as part of the upgrade?
> 
> **UPD.** I re-read and it felt a little like an objection 😬 It's not, I am just "tuning" my vision. Will be really helpful in case of probable further contributions from my side. I think I could have avoided this extra work if I had had a better understanding of these expectations before: this "default" aliases topic was surfaced during the tests. So the decision to keep other aliases was not an accident.

Hi Dmitry.  There are some external constraints that inform rolling back as opposed to treating this like users just depending on internal details:
1.) dbt has made a promise to users that when they select 'always latest' they will not be broken by updates
2.) the nature of dbt is that overriding macros is quite common, which means there's kind of no such thing as 'internal' when it comes to Jinja macros
3.) I think the previous mechanism for specifying incremental predicates pretty much required users to reference these 'internal' alias names.

Number 2 bites me from time to time, and not being aware of number 3 was a big miss on my part that I should have made you aware of during the PR.

PR #842, opened 2024-10-31
Downgrade dbt-core dependency to concrete required version (i.e. first one with behavior flags)

benc-db:
### Description

We would like to release our 1.9.0 soon, but dbt-core 1.9.0 has been postponed to December.  In this PR I'm lowering our dependency to released versions, and hiding microbatch as a valid strategy if your installed version is too low (because it will super break otherwise).

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/materializations/incremental/validate.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/validate.sql
@@ -35,7 +35,7 @@
     Use the 'merge' or 'replace_where' strategy instead
   {%- endset %}
 
-  {% if raw_strategy not in adapter.valid_incremental_strategies %}
+  {% if raw_strategy not in adapter.valid_incremental_strategies() %}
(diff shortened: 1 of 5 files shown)


dbt/include/databricks/macros/materializations/incremental/validate.sql
```diff
@@ -35,7 +35,7 @@
     Use the 'merge' or 'replace_where' strategy instead
   {%- endset %}
 
-  {% if raw_strategy not in ['append', 'merge', 'insert_overwrite', 'replace_where', 'microbatch'] %}
+  {% if raw_strategy not in adapter.valid_strategies %}
```
jackyhu-db, line 38: I saw above method is `valid_incremental_strategies`?

benc-db: good call out, probably about to fail the E2E tests lol.

jackyhu-db (dismissed):

PR #843, opened 2024-11-13
Cleanup warnings and test issues

benc-db:
### Description

Dbt asked me to not require pydantic > 2 (as it was causing issues).  So I wrote all of my pydantic in v1; however, if pydantic v2 got installed, a bunch of warnings would be emitted.  So now restricting to < 2.

Marked failing iceberg tests as skip for now (since there is an issue in prod).

### Checklist

### DIFF
--- a/tests/functional/adapter/iceberg/test_iceberg_support.py
+++ b/tests/functional/adapter/iceberg/test_iceberg_support.py
@@ -5,7 +5,8 @@ from dbt.tests import util
 from dbt.artifacts.schemas.results import RunStatus
 
 
-@pytest.mark.skip_profile("databricks_cluster")
+# @pytest.mark.skip_profile("databricks_cluster")
+@pytest.mark.skip("Skip for now as it is broken in prod")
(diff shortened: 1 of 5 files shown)


tests/functional/adapter/iceberg/test_iceberg_support.py
```diff
@@ -5,7 +5,8 @@
 from dbt.artifacts.schemas.results import RunStatus
 
 
-@pytest.mark.skip_profile("databricks_cluster")
+# @pytest.mark.skip_profile("databricks_cluster")
```
jackyhu-db, line 8: why do you comment out instead of deleting it?

jackyhu-db (approved):

tests/functional/adapter/iceberg/test_iceberg_support.py
benc-db: It's only temporary until next week, when the fix will get released to prod.

PR #844, opened 2024-11-15
V2 Materialization flag introduced

benc-db:
### Description

Introduces the V2 Materialization flag (which is primarily going to be used with create table / incremental), and handles the first materialization to be pivoted based on the flag: Seeds.  Will be going through each materialization in a separate PR to keep the changes scoped.

Go here to compare the difference in flows: https://github.com/databricks/dbt-databricks/blob/726e6e16d5aec864a37aadb443991d0da265e4b6/docs/seed_flow.md

Seeds are pretty close to ideal already, but had some no op statements that I removed (since we don't have transactions or indexes).

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/materializations/seeds/seeds.sql
+++ b/dbt/include/databricks/macros/materializations/seeds/seeds.sql
@@ -1,5 +1,58 @@
 {% materialization seed, adapter='databricks' %}
+  {% set target_relation = this.incorporate(type='table') %}
 
+  {% if adapter.behavior.use_table_creation_v2 %}
+    {{ create_seed_v2(target_relation) }}
+  {% else %}
+    {{ create_seed_v1(target_relation) }}
+  {% endif %}
+
+  {{ return({'relations': [target_relation]}) }}
+{% endmaterialization %}
+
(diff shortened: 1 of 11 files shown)


dbt/include/databricks/macros/materializations/seeds/seeds.sql
```diff
@@ -40,8 +93,6 @@
     {{ get_csv_sql(create_table_sql, sql) }};
   {% endcall %}
 
-  {% set target_relation = this.incorporate(type='table') %}
```
benc-db, line 43: There's more shared code that I can pull out, but I want to leave the existing materialization largely intact, since the whole reason to use the flag is to lower risk of significant changes.  Also, eventually V1 will be deleted.  I pulled out target_relation so that I could do the return statement at the top level.

tests/functional/adapter/fixtures.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+
+
+class MaterializationV2Mixin:
```
benc-db, line 4: Mixin that flips the flag on for test classes that include it.

jackyhu-db (approved):

PR #846, opened 2024-11-19
Adding new constraint logic that will be used with V2 flag

benc-db:
### Description

Adds new constraint logic for use with V2 materialization flag.  Not currently hooked up to anything (as that will come when I send the PR that splits table materialization into V1 and V2).

Core ideas:
1.) Use constraint config from dbt (as opposed to our homebrew from the 'meta' dictionary)
2.) Process constraints almost entirely in python since they are so logic heavy (this matches what dbt-core is doing anyway)
3.) Mirror but don't depend on dbt built-in functions for accomplishing this (as they expand the surface of adapter, and this logic has no dependency on the Adapter class).
4.) For processing constraint logic, use a functional style since it is pretty much entirely data transformation.
5.) Store processed constraints on Columns or Relations (since Databricks has timing requirements, namely that a.) not null is a column property (not a proper constraint) and b.) that check constraints can only be added via an alter.
6.) Following a functional style, when we 'enrich' an object, return a new copy.  This may need to switch to mutable in the future, depending on whether we need these changes cached or not (the relation cache does not have an easy to use update mechanism).

While writing this description, I'm realizing I'm missing unit tests for model constraints, so I will add those prior to merge.

### Checklist


dbt/adapters/databricks/constraints.py
```diff
@@ -0,0 +1,233 @@
...
+from functools import partial
+
+
+# Support constants
+CONSTRAINT_SUPPORT = {
+    ConstraintType.check: ConstraintSupport.ENFORCED,
+    ConstraintType.not_null: ConstraintSupport.ENFORCED,
+    ConstraintType.unique: ConstraintSupport.NOT_SUPPORTED,
+    ConstraintType.primary_key: ConstraintSupport.NOT_ENFORCED,
+    ConstraintType.foreign_key: ConstraintSupport.NOT_ENFORCED,
```
eric-wang-1990, line 21: It seems ConstraintType.Custom is also supported?

dbt/adapters/databricks/relation.py
```diff
@@ -152,6 +155,23 @@ def information_schema(self, view_name: Optional[str] = None) -> InformationSche
     def StreamingTable(cls) -> str:
         return str(DatabricksRelationType.StreamingTable)
 
+    def add_constraint(self, constraint: ModelLevelConstraint) -> None:
```
eric-wang-1990, line 158: Should we check constraint name uniqueness?

kmarq:
Is there any ability that this will be able to apply constraints without requiring the full contract enforcement? We've been looking to be able to define specific column constraints and PK/FK without having to define the full contract of columns. You'd have to define any column that you want constraints on, but any other columns would be merged in at runtime.
I've looked into trying to do this via a posthook, but from what I'm seeing if contract enforced=false then the constraints are not even visible by the time you are inside a model.

benc-db:
> Is there any ability that this will be able to apply constraints without requiring the full contract enforcement? We've been looking to be able to define specific column constraints and PK/FK without having to define the full contract of columns. You'd have to define any column that you want constraints on, but any other columns would be merged in at runtime.
I've looked into trying to do this via a posthook, but from what I'm seeing if contract enforced=false then the constraints are not even visible by the time you are inside a model.

It's not in this PR, but the V2 materialization approach basically makes a temp view and a final table that the temp view gets merged into.  This gives us a number of benefits, including that you don't need to specify contract enforced/column types in your model in order for us to have the capability of adding constraints to individual columns.

dbt/adapters/databricks/constraints.py
```diff
@@ -0,0 +1,233 @@
...
+    return supported
+
+
+def render_constraint(
+    constraint: T, render_funcs: dict[ConstraintType, RenderFunc]
+) -> Optional[str]:
+    rendered_constraint = ""
+
+    if constraint.type in render_funcs:
+        if constraint.name:
```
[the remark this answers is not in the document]
benc-db: No, if a constraint doesn't have a name then a.) you don't need to specify CONSTRAINT before the constraint definition (that is basically a keyword that tells DBSQL that the next token is going to be a name, rather than a constraint type) and b.) Databricks will generate one for you.

dbt/adapters/databricks/constraints.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def process_column_constraint(constraint: ColumnLevelConstraint) -> Optional[str]:
+    return process_constraint(constraint, supported_for_columns, column_constraint_map())
+
+
+def render_primary_key_for_column(constraint: ColumnLevelConstraint) -> str:
+    assert constraint.type == ConstraintType.primary_key
+    rendered = "PRIMARY KEY"
+    if constraint.expression:
```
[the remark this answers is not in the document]
benc-db: yes, because this is a column level constraint, so it is perfectly valid to just state that the column is a PRIMARY KEY.  In constrast, a model-level constraint needs to know which columns to treat as the primary key.  The support for expression is in case users want to add a constraint option such as DEFERRABLE.

dbt/adapters/databricks/column.py
```diff
@@ -26,5 +31,44 @@ def create(cls, name: str, label_or_dtype: str) -> "DatabricksColumn":
...
+        """Renders the column for building a create statement."""
+        column_str = f"{self.name} {self.dtype}"
+        if self.not_null:
+            column_str += " NOT NULL"
+        if self.comment:
+            comment = self.comment.replace("'", "\\'")
+            column_str += f" COMMENT '{comment}'"
+        for constraint in self.constraints or []:
+            c = constraints.process_column_constraint(constraint)
+            if c != "":
```
alexguo-db, line 69: nit: This check seems redundant, `process_constraint` only returns None or `render_constraint` which is None or not ""

benc-db: Thanks.  I probably missed this because I kept swapping back and forth on whether to use None or empty.

PR #854, opened 2024-11-23
Set unique temp table suffix to allow parallel merge execution

huangxingyi-git:
Resolves #

### Description
Set unique table suffix to allow parallel incremental execution

Our daily ETL need to execute parallel multiple dbt run of specific incremental model(`incremental_strategy=merge`), in which we pass the date and country as var argument.

If we want to process by batch of N countries and dates in parallel using Airflow concurrency, we need the temp views created by each of the dbt run to be unique. Else, you are going to end up with N insert attempting to run with the same __dbt_tmp name, creating conflict and ultimately creating failure.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Features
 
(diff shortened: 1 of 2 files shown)


benc-db:
Running tests

PR #857, opened 2024-11-26
Fix Python regressions in 1.9.0beta

benc-db:
### Description

Fixing two regressions related to Command submission of python models.

1.) Exceptions were not properly getting propagated to the user
2.) I had accidentally brought back the race condition that the community had already fixed for us.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/api_client.py
+++ b/dbt/adapters/databricks/api_client.py
@@ -88,8 +88,10 @@ class ClusterApi(DatabricksApi):
 
         response = self.session.post("/start", json={"cluster_id": cluster_id})
         if response.status_code != 200:
-            raise DbtRuntimeError(f"Error starting terminated cluster.\n {response.content!r}")
-        logger.debug(f"Cluster start response={response}")
+            if self.status(cluster_id) not in ["RUNNING", "PENDING"]:
+                raise DbtRuntimeError(f"Error starting terminated cluster.\n {response.content!r}")
+            else:
+                logger.debug("Presuming race condition, waiting for cluster to start")
 
         self.wait_for_cluster(cluster_id)
 
@@ -289,7 +291,7 @@ class CommandApi(PollableApi):
             raise DbtRuntimeError(f"Cancel command {command} failed.\n {response.content!r}")
 
     def poll_for_completion(self, command: CommandExecution) -> None:
-        self._poll_api(
+        response = self._poll_api(
             url="/status",
             params={
                 "clusterId": command.cluster_id,
@@ -300,7 +302,13 @@ class CommandApi(PollableApi):
             terminal_states={"Finished", "Error", "Cancelled"},
             expected_end_state="Finished",
             unexpected_end_state_func=self._get_exception,
-        )
+        ).json()
+
+        if response["results"]["resultType"] == "error":
+            raise DbtRuntimeError(
+                f"Python model failed with traceback as:\n"
+                f"{utils.remove_ansi(response['results']['cause'])}"
+            )
 
     def _get_exception(self, response: Response) -> None:
         response_json = response.json()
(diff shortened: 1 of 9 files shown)


dbt/adapters/databricks/api_client.py
```diff
@@ -88,8 +88,10 @@ def start(self, cluster_id: str) -> None:
 
         response = self.session.post("/start", json={"cluster_id": cluster_id})
         if response.status_code != 200:
-            raise DbtRuntimeError(f"Error starting terminated cluster.\n {response.content!r}")
-        logger.debug(f"Cluster start response={response}")
+            if self.status(cluster_id) not in ["RUNNING", "PENDING"]:
+                raise DbtRuntimeError(f"Error starting terminated cluster.\n {response.content!r}")
+            else:
+                logger.debug("Presuming race condition, waiting for cluster to start")
```
benc-db, line 94: When we query state after error, if it's pending or running it means the start failed due to race condition as another thread got the cluster started.

dbt/adapters/databricks/api_client.py
```diff
@@ -300,7 +302,13 @@ def poll_for_completion(self, command: CommandExecution) -> None:
             terminal_states={"Finished", "Error", "Cancelled"},
             expected_end_state="Finished",
             unexpected_end_state_func=self._get_exception,
-        )
+        ).json()
+
+        if response["results"]["resultType"] == "error":
```
benc-db, line 307: Lost this in the refactor: for some reason Command exec will give a state of 'Finished' rather then 'Error' some times, and then stuff the error in the results.

dbt/adapters/databricks/impl.py
```diff
@@ -85,7 +85,7 @@
 SHOW_TABLE_EXTENDED_MACRO_NAME = "show_table_extended"
 SHOW_TABLES_MACRO_NAME = "show_tables"
 SHOW_VIEWS_MACRO_NAME = "show_views"
-GET_COLUMNS_COMMENTS_MACRO_NAME = "get_columns_comments"
+
```
benc-db, line 88: Noticed this wasn't referenced anywhere when doing my debugging.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -70,6 +71,8 @@ def __init__(
 
     @override
     def submit(self, compiled_code: str) -> None:
+        logger.debug("Submitting Python model using the Command API.")
```
benc-db, line 74: Adding debug statement so that we can quickly determine submission method from logs.

alexguo-db (approved):
LGTM

dbt/adapters/databricks/api_client.py
jackyhu-db: should it be `info` rather than `debug`

jackyhu-db (approved):

dbt/adapters/databricks/api_client.py
jackyhu-db: what if the `result` or `resultType` does not exist

dbt/adapters/databricks/api_client.py
benc-db: Anything that is info will show in the normal dbt output, so we are very conservative about what we log with info. 
 This is normal operation; if they are using multiple python models and the command api they will hit this, so I don't think this is worth bringing to the users' attention if it's running or pending.  For whatever reason, the cluster start API errors if you ask to start a cluster that is already in the process of starting.  If it's not running or pending, they'll get the full output from raising the error.

dbt/adapters/databricks/api_client.py
benc-db: these are published parts of the Databricks REST API.  If these don't exist, the entire API becomes untrustworthy.

Issue #726, opened 2024-07-05
Add support for Pandas >2.0


alexeyegorov:
### Describe the bug
Since Numpy released its latest version 2.0, it is not compatible with an older version of Pandas. However, dbt-databricks in version 1.8.3 only supports pandas up to version 2.0.

Workaround: fix numpy version to 1.26.4 (latest before 2.0).

### Steps To Reproduce

1. For my devcontainer setup, I use `requirements.txt` with only few entries:
```
dbt-databricks==1.8.3
sqlfluff
sqlfluff-templater-dbt
```
2. Install the above dependencies.
3. Run `dbt deps`
4. Try to run any dbt command like `dbt compile`

### Expected behavior

Successfull dbt run.

### Screenshots and log output

The outcome of the commans:

![Bildschirmfoto 2024-07-05 um 15 07 16](https://github.com/databricks/dbt-databricks/assets/5102843/b5a0917d-f738-49b4-92f6-f582088a9da1)

DBT now has installed the packages.

But it fails in any other execution (in this case, it is `dbt compile`):

![Bildschirmfoto 2024-07-05 um 15 07 32](https://github.com/databricks/dbt-databricks/assets/5102843/a80ef518-3964-4223-b967-115fa36e9899)

Quote from the logs:

> 13:07:21  Running with dbt=1.8.3
> 
> A module that was compiled using NumPy 1.x cannot be run in
> NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
> versions of NumPy, modules must be compiled with NumPy 2.0.
> Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.
> 
> If you are a user of the module, the easiest solution will be to
> downgrade to 'numpy<2' or try to upgrade the affected module.
> We expect that some modules will need time to support NumPy 2.
> 
> Traceback (most recent call last):  File "/usr/local/bin/dbt", line 8, in <module>
>     sys.exit(cli())
>   File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1157, in __call__
>     return self.main(*args, **kwargs)
>   File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1078, in main
>     rv = self.invoke(ctx)
>   File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1688, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1434, in invoke
[...]

PR #874, opened 2024-12-11
Remove library restrictions if possible

### LINKED ISSUE #872: Pydantic restricted to v1
### Describe the bug

@benc-db, you mentioned that dbt asked you to not require Pydantic > 2, so you decided to restrict Pydantic to v1: https://github.com/databricks/dbt-databricks/pull/843

I am a bit confused as to why it was decided to constraint Pydantic to `>=1.10.0, <2` instead of just `>=1.10.0` which would have met dbt request. I am also confused about the reason for dbt request, as dbt itself does not publish any dependency on Pydantic.  

Based on the date and other related issues/PR, I am guessing that it might be related to issues encountered with the release of Pydantic 2.10 and maybe its impact on Airflow:
- https://github.com/pydantic/pydantic/issues/10910
- https://github.com/apache/airflow/pull/44284
- https://github.com/pydantic/pydantic/issues/10924

However, these were promptly resolved. For example, Airflow itself requires Pydantic v2 and decided to only blacklist the version 2.10.0 of Pydantic.

Making the version 1.9.0 of dbt-databricks effectively incompatible with any project requiring Pydantic v2 seems to be fairly extreme, would it be possible to get any context to the request form dbt?

As a general rule, every project should publish only its constraints. This package has no incompatibility with Pydantic v2, and it should therefore not publish any such constraint ;) If another project has an issue with Pydantic v2, they should publish this constraint themselves to not impact more users than necessary.

This has in turned caused further problems with other packages that were forced to constrain the version of `dbt-databricks`: https://github.com/astronomer/astronomer-cosmos/pull/1376

benc-db:
Resolves #872 , #726 

### Description

Removing dependency constraints that were originally put in place to stop warnings and breaks, but are preventing customer use cases.  If this passes all of the tests, will become 1.9.1

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-## dbt-databricks 1.9.0 (Dec 9, 2024)
+## dbt-databricks 1.9.1 (TBD)
+
+### Under the Hood
+
+- Removed pins for pandas and pydantic to ease user burdens ([874](https://github.com/databricks/dbt-databricks/pull/874))
+
+## dbt-databricks 1.9.0 (December 9, 2024)
 
 ### Features
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,8 +29,7 @@ dependencies = [
     "dbt-core>=1.8.7, <2.0",
     "dbt-spark>=1.8.0, <2.0",
     "keyring>=23.13.0",
-    "pandas<2.2.0",
-    "pydantic>=1.10.0, <2",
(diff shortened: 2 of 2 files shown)


pankajastro:
Hi @benc-db, Would you be able to share when you’re planning to release this change?

benc-db:
> Hi @benc-db, Would you be able to share when you’re planning to release this change?

Monday

PR #876, opened 2024-12-12
Redo constraints (since columns can only take one constraint at create time)

benc-db:
### Description

When I went to apply the constraint logic from the previous PR, I discovered that Databricks does not like receiving multiple constraints for a single column in the table create syntax; however, you can add arbitrarily many table level constraints, so in this PR I aim to do the following:

* if a column has a not_null constraint, directly set not_null on the column
* any other column constraint gets converted to a TypedConstraint, a subclass of ModelLevelConstraint that enforces the requirements of the particular constraint, and holds the rendering logic.  The conversion required setting 'columns=[name]'. Because python makes operator overloading a pain in the butt, I'm going back to OOP for the render logic.
* If a model level constraint is not_null, convert to setting not_null on each affected column, since Databricks does not support not_null constraint definition as a table constraint
* any other model constraint gets converted to a TypedConstraint

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -769,21 +769,28 @@ class DatabricksAdapter(SparkAdapter):
 
     @available
     @staticmethod
-    def get_enriched_columns(
-        existing_columns: list[DatabricksColumn], model_columns: dict[str, dict[str, Any]]
-    ) -> list[DatabricksColumn]:
+    def parse_columns_and_constraints(
+        existing_columns: list[DatabricksColumn],
+        model_columns: dict[str, dict[str, Any]],
+        model_constraints: list[dict[str, Any]],
+    ) -> tuple[list[DatabricksColumn], list[constraints.TypedConstraint]]:
         """Returns a list of columns that have been updated with features for table create."""
         enriched_columns = []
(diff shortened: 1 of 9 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -769,21 +769,28 @@ def generate_unique_temporary_table_suffix(self, suffix_initial: str = "__dbt_tm
 
     @available
     @staticmethod
-    def get_enriched_columns(
-        existing_columns: list[DatabricksColumn], model_columns: dict[str, dict[str, Any]]
-    ) -> list[DatabricksColumn]:
+    def parse_columns_and_constraints(
```
alexguo-db, line 772: Is this used anywhere?

alexguo-db (approved):
LGTM

dbt/adapters/databricks/impl.py
benc-db: Only in my branch where I found out it doesn't really work.  I tried to isolate model and column constraint parsing/processing before, but when I tried to use the constraint code on my local branch, I discovered that Databricks does not like multiple column constraints on a single column; also dbt supports specifying a bunch of null constraints at the model level.  Hence why I have to do this less elegant, 'process them all at the same time' approach :/.

PR #888, opened 2024-12-19
Refactor reading env vars

benc-db:
### Description

Centralize env var reading since it is actually a global state reference, and we should be aware of as such.  We will lazy set once, and return values from then on (since allowing users to change the value during a run, and have that change reflected seems crazy unpredictable).

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -59,6 +59,7 @@ from dbt.adapters.databricks.events.cursor_events import (
     CursorCreate,
 )
 from dbt.adapters.databricks.events.other_events import QueryError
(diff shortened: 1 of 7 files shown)


dbt/adapters/databricks/connections.py
```diff
@@ -475,6 +473,8 @@ def add_query(
         auto_begin: bool = True,
         bindings: Optional[Any] = None,
         abridge_sql_log: bool = False,
+        retryable_exceptions: tuple[type[Exception], ...] = tuple(),
```
benc-db, line 476: This is new interface in latest upstream.  Hopefully this doesn't break anything, otherwise we (and other adapters) will be in for a world of hurt.

PR #896, opened 2025-01-07
Actually refactor to move off info schema

benc-db:
### Description

Step 1. Refactor existing behavior behind a class that I can select at run time to switch behavior based on a behavior flag.  There are no behavior changes in this step, just refactoring.

### Checklist


benc-db:
Note: I'm sure there are things in the existing code that could be better; however, intent is to replace this behavior with a new one, so no attempt has been made to improve the existing code, just moving it behind an interface.

PR #910, opened 2025-01-21
Compress to one connection manager

benc-db:
### Description

For a long time we've had code that suggested that you could either reuse sessions or not by setting an env variable.  This hasn't worked for a long time, and since I need to refactor some of this code to improve cancellation behavior, I want to first compress to a single ConnectionManager, so that I don't need to make changes against both.  I eventually want to get rid of session reuse, but its too large to do as part of 1.9.x, where I want to improve cancellation behavior, so for now, let the code include only the paths that are actually reachable.

Note: no new code here, just collapsing into the parent class, and keeping the union of the code from the child and the parent.  Each method comes either from the parent (if the previous child didn't override) or the child (if the child did override, or added a new method).

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -1,5 +1,4 @@
 import decimal
-import os
 import re
 import sys
 import time
@@ -10,7 +9,6 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from multiprocessing.context import SpawnContext
 from numbers import Number
-from threading import get_ident
 from typing import TYPE_CHECKING, Any, Optional, cast
 
 from dbt_common.events.contextvars import get_node_info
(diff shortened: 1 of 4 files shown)


dbt/adapters/databricks/connections.py
```diff
@@ -431,39 +431,19 @@ def set_connection_name(
...
 
         conn_name: str = "master" if name is None else name
 
         # Get a connection for this thread
-        conn = self.get_if_exists()
-
-        if conn and conn.name == conn_name and conn.state == ConnectionState.OPEN:
-            # Found a connection and nothing to do, so just return it
-            return conn
+        conn = self._get_if_exists_compute_connection(_get_compute_name(query_header_context) or "")
```
benc-db, line 439: Git is representing this as a bunch of changes here, but the code is not net new, just moved up into the parent class.

PR #912, opened 2025-01-24
Clean up cursor management

benc-db:
### Description

This is a step towards improving cancellation behavior by limiting the exposure of cursors, and preventing further operations on connections/cursors that have already been scheduled for close/cancel.  I'm not positive this fixes the issue that kicked off this work, as I have had a hard time reproing the issue (dbt operation gets cancelled, query continues running).  But, at a minimum, this should reduce the likelihood of that happening by ensuring everything is cancelled prior to closing by making sure that once cancellation has started, no new operations start running (hence the lock).  Further, no raw cursors are handed out, and we ensure that every cursor is closed before the next cursor for the connection is opened.  Since the connections have thread affinity, and within a thread we have sequential operation, there should be no reason to keep multiple cursors open from the same connection at the same time.  We also now have much more extensive unit testing of the connection/cursor management behavior than previously.

This PR starts removing the event classes as well.  In the end I think they added more overhead without actually improving readability, since we don't have access to real event classes that they use in Dbt.

Apologies for the random files with formatting changes; not sure why Ruff decided those needed to change.  The import files to review are connections.py, handle.py, and test_handle.py.

### Checklist


dbt/adapters/databricks/handle.py
```diff
@@ -0,0 +1,217 @@
...
+    def __str__(self) -> str:
+        session_id = "Unknown"
+        try:
+            session_id = self._conn.get_session_id_hex()
+        except Exception:
+            pass
+        return f"Connection(session-id={session_id})"
+
+
+def clean_sql(sql: str) -> str:
```
benc-db, line 179: Looking for recommendations on how we want to handle pure functions.  They can live raw like this, or they could be static methods on a class.  That class could be DatabricksHandle, or it could be a separate class that acts as a namespace, since the operations don't actually operate on DatabricksHandles (they operate on data that is private to them).

dbt/adapters/databricks/handle.py
```diff
@@ -0,0 +1,217 @@
...
+    cleaned = sql.strip()
+    if cleaned.endswith(";"):
+        cleaned = cleaned[:-1]
+    return cleaned
+
+
+def get_dbr_version(cursor: Cursor) -> Cursor:
```
jackyhu-db, line 186: why this is not part of the CursorWrapper

dbt/adapters/databricks/handle.py
```diff
@@ -0,0 +1,308 @@
...
+CursorExecOp = Callable[[Cursor], Cursor]
+CursorWrapperOp = Callable[["CursorWrapper"], None]
+ConnectionOp = Callable[[Optional[Connection]], None]
+LogOp = Callable[[], str]
+FailLogOp = Callable[[Exception], str]
+
+
+class CursorWrapper:
+    """
+    Wrap the DBSQL cursor to abstract the details from DatabricksConnectionManager.
```
jackyhu-db, line 31: it would better if it can also implement `__enter__` and `__exit__` so that it can be open with `with` statement which can automatically be closed.

dbt/adapters/databricks/handle.py
```diff
@@ -0,0 +1,308 @@
...
+class DatabricksHandle:
+    """
+    Handle for a Databricks SQL Session.
+    Provides a layer of abstraction over the Databricks SQL client library such that
+    DatabricksConnectionManager does not depend on the details of this library directly.
+    """
+
+    def __init__(
+        self,
+        conn: Optional[Connection],
```
jackyhu-db, line 99: in which case conn will be None? Why it needs to construct with None?

dbt/adapters/databricks/connections.py
```diff
@@ -557,28 +358,24 @@ def _execute_cursor(
...
-                if cursor is not None:
+                if cursor:
                     cursor.close()
 
     def list_schemas(self, database: str, schema: Optional[str] = None) -> "Table":
         database = database.strip("`")
         if schema:
             schema = schema.strip("`").lower()
-        return self._execute_cursor(
+        return self._execute_with_cursor(
```
jackyhu-db, line 368: can we use `with` + code block,  which makes the code more readable?

dbt/adapters/databricks/utils.py
```diff
@@ -77,3 +78,13 @@ def handle_missing_objects(exec: Callable[[], T], default: T) -> T:
...
 def quote(name: str) -> str:
     return f"`{name}`"
+
+
+ExceptionToStrOp = Callable[[Exception], str]
+
+
+def handle_exceptions_as_warning(op: Callable[[], None], log_gen: ExceptionToStrOp) -> None:
+    try:
+        return op()
```
jackyhu-db, line 88: nit: the declaration of op returns `None`, but what if op returns some value instead of `None`, so it does not match the return type of this function.

dbt/adapters/databricks/handle.py
```diff
@@ -0,0 +1,308 @@
...
+
+        connection = DatabricksHandle(conn, is_cluster=is_cluster)
+
+        logger.debug(f"{connection} - Created")
+
+        return connection
+
+    def _cleanup(
+        self,
+        connect_op: ConnectionOp,
```
jackyhu-db, line 205: given this is the private function, can it do the cleanup (close cursor) directly?

benc-db: I'm not certain which of two things you mean:
a.) why do we need to take a function as an argument or 
b.) why do we call an operation against the cursor wrapper rather than the underlying cursor?

If the question is 'a', it's because we have two scenarios with identical structure, but slightly different details (i.e. in one we need to call close and in the other we need to call cancel).  The alternative to taking a function as an argument is just to have the two public methods with identical structure.

If the question is 'b', it's so that I can make sure that if any other parts of the code are still holding the CursorWrapper, if they try to close it again we don't get a warning from PySQL (that users keep thinking is an error :/).

dbt/adapters/databricks/handle.py
benc-db: I'll look into this; I agree that it is better ergonomically, I just know there are a couple of paths in the code today that create the cursor and return it to another function that later gets the results.  I might need to restructure the connection manager to make use of that.  For this PR, I can implement the enter/exit, and then maybe switch to using it in another PR.

dbt/adapters/databricks/handle.py
benc-db: When I was testing, I found cases where PySQL would return None for the connection without raising an error.  In connection manager, the existing code expects these cases.  I will look into it, because this class without a connection is a little strange.

dbt/adapters/databricks/handle.py
```diff
@@ -0,0 +1,308 @@
...
+                # Assuming SQL Warehouse uses the latest version.
+                self._dbr_version = (sys.maxsize, sys.maxsize)
+
+        return self._dbr_version
+
+    @property
+    def session_id(self) -> str:
+        if self._conn:
+            return self._conn.get_session_id_hex()
+        return "N/A"
```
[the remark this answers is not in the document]
benc-db: Reasonable.  This is replicating existing functionality, but since it's only used in logging, it should be fine to switch.

dbt/adapters/databricks/utils.py
benc-db: Good point.  This is something that would be fine in a statically typed world, but wrong in python lol.

dbt/adapters/databricks/handle.py
jackyhu-db: Actually I mean if `_cleanup` is a private function, can we do all the actual cleanup logic inside. e.g.do the cursor.close in the `_cleanup` rather than in the op function.

jackyhu-db (approved):

dbt/adapters/databricks/handle.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+        Common cleanup function for cursor operations, handling either close or cancel.
+        """
+        if self.open:
+            self.open = False
+            logger.debug(startLog())
+            utils.handle_exceptions_as_warning(lambda: cleanup(self._cursor), failLog)
+
+    def fetchall(self) -> Sequence[tuple]:
+        return self._cursor.fetchall()
```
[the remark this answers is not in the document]
benc-db: Probably.  The question then is should we no op or raise an exception...I guess this is where the context manager is a good option, because I'm inclined to no op (since the only case where this _should_ happen is if someone is holding a cursorwrapper after a program cancel has been requested), but if this happens in any scenario other than cancel, it means an assumption in the code execution has been violated and should fail with an assert.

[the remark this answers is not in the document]
benc-db: Since I raised as an exception on connection, also raising here. I think the calling code won't know what to do if we no op, so better to halt execution immediately.  Would prefer we cleanup quietly, but this is the best I think we can do for now.

dbt/adapters/databricks/handle.py
benc-db: Found and fixed, thanks.

PR #919, opened 2025-01-28
Update impl.py to use POSIX standard on location

gsolasab:
### Description

Executing dbt locally in a Windows machine is including front slash when using os.path.join function for separating different elements of the location and so generating issues when creating the tables. 
Using posix join it will always use back slash to separate the elements of the location and so work in any operating system.

### Checklist


benc-db:
Makes sense, just validating against the E2E tests.

benc-db (approved):

PR #935, opened 2025-02-06
Allow auto liquid clustering

### LINKED ISSUE #932: Allow liquid clustering `CLUSTER BY AUTO` option
### Describe the feature
Currently the dbt-databricks liquid clustering feature relies on providing columns and then generating an alter statement that looks like below

Ideally if users don't define columns to cluster by, and a config `auto_liquid_cluster` is true then their relations should be clustered

ShaneMazur:
Resolves #932

### Description

Adds config `auto_liquid_cluster` to allow for tables to be `CLUSTER BY AUTO` on creation or update (i.e. incremental run)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## dbt-databricks 1.9.5 (TBD)
 
+### Features
+
+- Add `auto_liquid_cluster` config to enable Auto Liquid Clustering for Delta-based dbt models. 
+
 ### Under the Hood
 
 - Collapsing to a single connection manager (since the old one no longer works) ([910](https://github.com/databricks/dbt-databricks/pull/910))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Features
 
-- Add `auto_liquid_cluster` config to enable Auto Liquid Clustering for Delta-based dbt models.
+- Add `auto_liquid_cluster` config to enable Auto Liquid Clustering for Delta-based dbt models ([935](https://github.com/databricks/dbt-databricks/pull/935))
 
 ### Fixes
 
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -110,6 +110,7 @@ class DatabricksConfig(AdapterConfig):
     partition_by: Optional[Union[list[str], str]] = None
     clustered_by: Optional[Union[list[str], str]] = None
     liquid_clustered_by: Optional[Union[list[str], str]] = None
+    auto_liquid_cluster: Optional[bool] = None
     buckets: Optional[int] = None
     options: Optional[dict[str, str]] = None
     merge_update_columns: Optional[str] = None
--- a/dbt/include/databricks/macros/relations/liquid_clustering.sql
+++ b/dbt/include/databricks/macros/relations/liquid_clustering.sql
@@ -1,21 +1,29 @@
 {% macro liquid_clustered_cols() -%}
   {%- set cols = config.get('liquid_clustered_by', validator=validation.any[list, basestring]) -%}
+  {%- set auto_cluster = config.get('auto_liquid_cluster', validator=validation.any[boolean]) -%}
   {%- if cols is not none %}
     {%- if cols is string -%}
       {%- set cols = [cols] -%}
     {%- endif -%}
     CLUSTER BY ({{ cols | join(', ') }})
+    {%- elif auto_cluster -%}
+    CLUSTER BY AUTO
   {%- endif %}
 {%- endmacro -%}
 
 {% macro apply_liquid_clustered_cols(target_relation) -%}
   {%- set cols = config.get('liquid_clustered_by', validator=validation.any[list, basestring]) -%}
+  {%- set auto_cluster = config.get('auto_liquid_cluster', validator=validation.any[boolean]) -%}
   {%- if cols is not none %}
     {%- if cols is string -%}
       {%- set cols = [cols] -%}
     {%- endif -%}
     {%- call statement('set_cluster_by_columns') -%}
         ALTER {{ target_relation.type }} {{ target_relation.render() }} CLUSTER BY ({{ cols | join(', ') }})
+    {%- endcall -%}
+  {%- elif auto_cluster -%}
+    {%- call statement('set_cluster_by_auto') -%}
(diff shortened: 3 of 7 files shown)


benc-db:
Can you do me one small favor: do you know if Databricks has any documentation that covers this?  I'm struggling to find it...I'm fairly confident that if the E2E tests pass, this is valid Databricks SQL, but would be nice to have.

benc-db:
> Can you do me one small favor: do you know if Databricks has any documentation that covers this? I'm struggling to find it...I'm fairly confident that if the E2E tests pass, this is valid Databricks SQL, but would be nice to have.

Part of why I'm interested in the documentation is to see if we are using the most discoverable name for the config.

benc-db:
Are you set up with hatch in your dev environment?  If so, please run `hatch run code-quality` to fix the linting issue.  If you don't, I can do it.

ShaneMazur:
> > Can you do me one small favor: do you know if Databricks has any documentation that covers this? I'm struggling to find it...I'm fairly confident that if the E2E tests pass, this is valid Databricks SQL, but would be nice to have.
> 
> Part of why I'm interested in the documentation is to see if we are using the most discoverable name for the config.

There are these but nothing in the Databricks official documentation but it is mentioned in [this video from last year's summit](https://www.youtube.com/watch?v=yZmrpXJg-G8&t=868s) 

There was also a PDF provided to us by the PM but not sure if that has changed and probably not a good idea to share it here anyway. Let me know and I can share it if necessary.

ShaneMazur:
> Are you set up with hatch in your dev environment? If so, please run `hatch run code-quality` to fix the linting issue. If you don't, I can do it.

Done ✅

ShaneMazur:
> No worries, it just means that this will likely be a 'secret' feature of the adapter until the Databricks feature is documented. When you set auto cluster by, do you still need to run optimize manually? Going to pull and push to run the E2E tests.

Yes - `CLUSTER BY AUTO` is just for automatic key selection. Triggering the clustering automatically is still handled by [predictive optimization](https://docs.databricks.com/en/optimizations/predictive-optimization.html#:~:text=Predictive%20optimization%20removes%20the%20need,queues%20these%20operations%20to%20run.).

benc-db:
Do you have an environment where you could write/run an E2E test?  Unit test looks good, but would like to see a test similar to https://github.com/databricks/dbt-databricks/blob/main/tests/functional/adapter/liquid_clustering/test_liquid_clustering.py
to validate that the inserted code doesn't break in DBSQL.

benc-db (changes requested):
Please add a 'functional' aka E2E test similar to existing liquid clustering tests.  Since liquid clustering is an optimization thing, we can't really validate the outcomes, but I just want a regression test to ensure we notice if they change syntax for auto cluster by before final release.

ShaneMazur:
> Please add a 'functional' aka E2E test similar to existing liquid clustering tests. Since liquid clustering is an optimization thing, we can't really validate the outcomes, but I just want a regression test to ensure we notice if they change syntax for auto cluster by before final release.

Added that and included auto liquid clustering in the optimize logic since it is still valid to run optimize on liquid clustering with auto selected keys

benc-db:
> Added that and included auto liquid clustering in the optimize logic since it is still valid to run optimize on liquid clustering with auto selected keys

Looks good, let me just run tests.  If everything is green, I'll merge it in.  Thanks!

benc-db (approved):

Vzzarr:
Hi @ShaneMazur @benc-db , thanks for the change you made here, much appreciated ❤️ 

I don't know if you are the best person to answer my question but starting somewhere:

With the notation introduced here, if I have the tables being recreated on a full refresh every time with Liquid Clustering: will this benefit from the Liquid Clustering or with every full refresh restart from 0 the analytics made on the table usage and predictive optimisation?

Let me know if unclear but I'll probably raise this with Databricks as well.

benc-db:
>With the notation introduced here, if I have the tables being recreated on a full refresh every time with Liquid Clustering: will this benefit from the Liquid Clustering or with every full refresh restart from 0 the analytics made on the table usage and predictive optimisation?

This is a very good question.  I would assume it would start from 0, but it given that there is continuity of history when we use create or replace, it might not?

ShaneMazur:
> > With the notation introduced here, if I have the tables being recreated on a full refresh every time with Liquid Clustering: will this benefit from the Liquid Clustering or with every full refresh restart from 0 the analytics made on the table usage and predictive optimisation?
> 
> This is a very good question. I would assume it would start from 0, but it given that there is continuity of history when we use create or replace, it might not?

I would assume it is persisted with the delta table history - so I think it would have continuity as long as you don't drop your table. It would seem pretty pointless if it starts from scratch every time a table is replaced - excluding table schema changes probably. @benc-db maybe someone internal to Databricks can answer this?

Vzzarr:
Thanks @ShaneMazur @benc-db for your inputs, I raised the concern with Databricks Support and got the following response:

> Engineering team mentions that to reap the benefits of liquid clustering you would have to avoid replacing the table at the end of the day. The CREATE OR REPLACE will wipe the clustering metadata required for liquid clustering.
> 
> Appending data incrementally instead of replacing the table would persist the liquid cluster information. So the best approach would be to use incremental writes (append/merge) and avoiding frequent full refreshes. 

So we would get the benefits only on incremental models and not on the full refresh. Might be worth noting this in the documentation? Let me know if this is something I can do 🙌

benc-db:
>So we would get the benefits only on incremental models and not on the full refresh. Might be worth noting this in the documentation? Let me know if this is something I can do 🙌

You benefit from liquid clustering on reads when the table is sufficiently large, and a traditional partitioning strategy would be hard to manage, due to granularity of the partitions.  So, on a full refresh, you will have write overhead of setting up liquid clustering.  Whether its worth it is going to be a function of how often the table gets queried, and how frequently you replace the table.

ShaneMazur:
@Vzzarr I think Databricks support is giving you incorrect information. I've spoken with the team developing liquid clustering to confirm that auto clustering cols are preserved even during schema evolution as long as the selected clustering columns still exist. I've included a screenshot of one of their responses below. They've said they will add comments here sometime this week

Screenshot 2025-04-04 at 1 40 58 PM

Hopefully this clarifies thing. Fingers crossed the liquid clustering folks can elaborate here on the behaviour and add it to [their docs](https://docs.databricks.com/aws/en/delta/clustering#automatic-liquid-clustering)

PR #938, opened 2025-02-11
Prepare for environments with serverless notebooks

benc-db:
### Description

Learned from working with jobs team today that to use environments, we need the tasks to take an environment key...but also learned this doesn't work yet with serverless notebooks.  This code is to prepare for that feature to come to Databricks, to minimize time without this feature for users.  Normally I hate doing that, but there are user workloads blocked on this.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/api_client.py
+++ b/dbt/adapters/databricks/api_client.py
@@ -327,7 +327,8 @@ class JobRunsApi(PollableApi):
         self, run_name: str, job_spec: dict[str, Any], **additional_job_settings: dict[str, Any]
     ) -> str:
         logger.debug(
-            f"Submitting job with run_name={run_name} and job_spec={job_spec} and additional_job_settings={additional_job_settings}"
+            f"Submitting job with run_name={run_name} and job_spec={job_spec}"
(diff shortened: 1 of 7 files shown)


dbt/adapters/databricks/api_client.py
```diff
@@ -326,6 +326,10 @@ def __init__(self, session: Session, host: str, polling_interval: int, timeout:
     def submit(
         self, run_name: str, job_spec: dict[str, Any], **additional_job_settings: dict[str, Any]
     ) -> str:
+        logger.debug(
+            f"Submitting job with run_name={run_name} and job_spec={job_spec}"
+            " and additional_job_settings={additional_job_settings}"
```
jackyhu-db, line 331: will `additional_job_settings` include some credential data? If yes, we may have to redact it

jackyhu-db (approved):

dbt/adapters/databricks/api_client.py
benc-db: there is no setting that requires a credential.  In that sense it is only as risky as any other user-provided setting, i.e., they can do something like name a table after their SSN.

PR #946, opened 2025-02-26
Cleanup logging

benc-db:
### Description

Removing low information/redundant lines from debug log.  This makes debug logs easier to read, since there is less noise between the core actions.  Removing based on which events have not been useful for investigation in the last year.  Also starting to clean up the formatting of macros, though limited how much of that is in this PR, since there are many places that need attention and it would make this PR too big.  Will clean up macro renderings as I touch them going forward.

FYI, generally macros are gross because the way we want to write them as code is not aligned with how we need to write them so that they get rendered out reasonably.

### Checklist


dbt/include/databricks/macros/relations/view/create.sql
```diff
@@ -1,20 +1,39 @@
...
   {{ tblproperties_clause() }}
-  as
+  as (
     {{ sql }}
+  )
 {% endmacro %}
+
+
+{% macro get_column_comment_sql(column_name, column_dict) -%}
+  {%- if column_name in column_dict and column_dict[column_name]["description"] -%}
```
benc-db, line 26: This is a stop gap where I write over the dbt-spark macro versions.  In my next PR, where I do Views V2, I will probably replace this with python code, like what I did with Table

PR #950, opened 2025-03-03
Dealing with changes to the info schema

benc-db:
### Description

Colin brought to my attention that there is a full_data_type column now in the info schema that is more apt, since it gives the details of struct.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## dbt-databricks 1.9.8 (TBD)
+
+### Fixes
+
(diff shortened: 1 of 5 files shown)


alexguo-db (dismissed):

benc-db:
I tried flipping the flag on and running all tests.  I discovered that we were failing tests due to not being able to repair table for non-Delta tables, even when on UC.  This limits the applicability of this even more.  To be able to use the default behavior for that case, I need to establish if the object is Delta when I create the relation that represents it.

PR #954, opened 2025-03-04
Views V2

benc-db:
### Description

Next step in the V2 materialization saga, this time Views.  The big deal with views is adding the ability to update previously deployed views with ALTER rather than replacement, so that they keep the same grants, and UC id (which is how create or replace already works for tables).  The other change is that now replace.sql is doing more heavy lifting, simplifying what would otherwise be a fairly complicated materialization.

Probably need more tests, but want to get the ideas out in front of reviewers

### Checklist


dbt/adapters/databricks/impl.py
```diff
@@ -103,7 +104,7 @@
 
 USE_MATERIALIZATION_V2 = BehaviorFlag(
     name="use_materialization_v2",
-    default=False,
+    default=True,
```
benc-db, line 107: Need to revert this.

dbt/adapters/databricks/impl.py
```diff
@@ -138,7 +139,9 @@ class DatabricksConfig(AdapterConfig):
     target_alias: Optional[str] = None
     source_alias: Optional[str] = None
     merge_with_schema_evolution: Optional[bool] = None
-    safe_table_create: Optional[bool] = None
+    use_safer_relation_operations: Optional[bool] = None
+    incremental_apply_config_changes: Optional[bool] = None
```
benc-db, line 143: This was previously missing

dbt/adapters/databricks/impl.py
```diff
@@ -744,23 +747,11 @@ def get_relation_config(self, relation: DatabricksRelation) -> DatabricksRelatio
             return StreamingTableAPI.get_from_relation(self, relation)
         elif relation.type == DatabricksRelationType.Table:
             return IncrementalTableAPI.get_from_relation(self, relation)
+        elif relation.type == DatabricksRelationType.View:
```
benc-db, line 750: Probably need to move this down like the other method, to get it closer to the API classes

dbt/adapters/databricks/impl.py
```diff
@@ -918,3 +923,30 @@ def _describe_relation(
             results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
         results["show_tblproperties"] = adapter.execute_macro("fetch_tbl_properties", kwargs=kwargs)
         return results
+
+
+class ViewAPI(RelationAPIBase[ViewConfig]):
```
benc-db, line 928: We need to use something similar to MV/ST and Incremental to check for changes so that we can update views via ALTER.

dbt/adapters/databricks/relation.py
```diff
@@ -91,6 +91,10 @@ def is_materialized_view(self) -> bool:
     def is_streaming_table(self) -> bool:
         return self.type == DatabricksRelationType.StreamingTable
 
+    @property
```
benc-db, line 94: this was missing from a previous PR

dbt/adapters/databricks/relation_configs/base.py
```diff
@@ -119,14 +119,23 @@ def from_results(cls, results: RelationResults) -> Self:
 
         return cls(config=config_dict)
 
-    @abstractmethod
     def get_changeset(self, existing: Self) -> Optional[DatabricksRelationChangeSet]:
         """Get the changeset that must be applied to the existing relation to make it match the
         current state of the dbt project. If no changes are required, this method should return
         None.
         """
+        changes: dict[str, DatabricksComponentConfig] = {}
```
benc-db, line 127: Defining a base since the incremental and view versions of this method are very similar

dbt/adapters/databricks/relation_configs/view.py
```diff
@@ -0,0 +1,22 @@
...
+from dbt.adapters.databricks.relation_configs.tags import TagsProcessor
+from dbt.adapters.databricks.relation_configs.tblproperties import TblPropertiesProcessor
+
+
+class ViewConfig(DatabricksRelationConfigBase):
+    config_components = [TagsProcessor, TblPropertiesProcessor, QueryProcessor, CommentProcessor]
+
+    def get_changeset(self, existing: Self) -> Optional[DatabricksRelationChangeSet]:
+        changeset = super().get_changeset(existing)
+        if changeset and "comment" in changeset.changes:
```
benc-db, line 20: For some reason there is no mechanism in Databricks to alter the comment on a view ;(.  Therefore, even if you want me to update the view via alter, if you update the description, we'll need to replace the view.

dbt/include/databricks/macros/adapters/columns.sql
```diff
@@ -56,7 +56,7 @@
 
 {% macro databricks__get_columns_in_query(select_sql) %}
   {{ log("Getting column information via empty query")}}
-  {% call statement('get_columns_in_query', fetch_result=True, auto_begin=False) -%}
+  {% call statement('get_columns_in_query', fetch_result=True) -%}
```
benc-db, line 59: auto_begin doesn't do anything in Databricks

dbt/include/databricks/macros/adapters/persist_docs.sql
```diff
@@ -4,7 +4,7 @@
       {% set comment = column['description'] %}
       {% set escaped_comment = comment | replace('\'', '\\\'') %}
       {% set comment_query %}
-        alter table {{ relation.render()|lower }} change column {{ api.Column.get_name(column) }} comment '{{ escaped_comment }}';
+        COMMENT ON COLUMN {{ relation.render()|lower }}.{{ api.Column.get_name(column) }} IS '{{ escaped_comment }}';
```
benc-db, line 7: This form works with tables and views

dbt/include/databricks/macros/etc/statement.sql
```diff
@@ -15,3 +15,17 @@
 
   {%- endif -%}
 {%- endmacro %}
+
+{% macro execute_multiple_statements(statements) %}
```
benc-db, line 19: Refactoring out of MV/ST because now we have a third place that we need this pattern.

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -15,7 +15,7 @@
   {% if adapter.behavior.use_materialization_v2 %}
     {{ log("USING V2 MATERIALIZATION") }}
     {#-- Set vars --#}
-    {% set safe_create = config.get('safe_table_create', True) | as_bool  %}
+    {% set safe_create = config.get('use_safer_relation_operations', True) | as_bool  %}
```
benc-db, line 18: renamed this flag since now I want to use it for views too.

dbt/include/databricks/macros/materializations/materialized_view.sql
```diff
@@ -32,7 +32,7 @@
 
         -- get config options
         {% set on_configuration_change = config.get('on_configuration_change') %}
-        {% set configuration_changes = get_materialized_view_configuration_changes(existing_relation, config) %}
+        {% set configuration_changes = get_configuration_changes(existing_relation) %}
```
benc-db, line 35: Was able to make this macro more generic too.

dbt/include/databricks/macros/relations/create_backup.sql
```diff
@@ -1,10 +1,9 @@
 {%- macro create_backup(relation) -%}
-    -- get the standard backup name
-    {% set backup_relation = make_backup_relation(relation, relation.type) %}
+  -- get the standard backup name
```
benc-db, line 2: working on consistent tab size of 2

dbt/include/databricks/macros/relations/replace.sql
```diff
@@ -1,52 +1,71 @@
 {% macro get_replace_sql(existing_relation, target_relation, sql) %}
-    {{- log('Applying REPLACE to: ' ~ existing_relation) -}}
-    {% do return(adapter.dispatch('get_replace_sql', 'dbt')(existing_relation, target_relation, sql)) %}
+  {{- log('Applying REPLACE to: ' ~ existing_relation) -}}
+  {% do return(adapter.dispatch('get_replace_sql', 'dbt')(existing_relation, target_relation, sql)) %}
 {% endmacro %}
 
 {% macro databricks__get_replace_sql(existing_relation, target_relation, sql) %}
-    {# /* use a create or replace statement if possible */ #}
+  {# /* if safe_relation_replace, prefer renaming */ #}
```
benc-db, line 7: Updated this macro per replace_flow.md.  Now works for views as well as MV/ST

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 - Support databricks OAuth M2M auth type. Updated OAuth readme doc with instructions.([827](https://github.com/databricks/dbt-databricks/pull/827))
 - Introduced use_materialization_v2 flag for gating materialization revamps. ([844](https://github.com/databricks/dbt-databricks/pull/844))
-- Introduce Tables V2, including safe_table_create which will not change the production table unless new data can safely be ingested ([927](https://github.com/databricks/dbt-databricks/pull/927))
+- Introduce Tables V2, including use_safer_relation_operations which will not change the production table unless new data can safely be ingested ([927](https://github.com/databricks/dbt-databricks/pull/927))
```
jackyhu-db, line 7: is this PR 954?

dbt/adapters/databricks/relation_configs/base.py
```diff
@@ -119,14 +119,23 @@ def from_results(cls, results: RelationResults) -> Self:
...
+        for component in self.config_components:
+            key = component.name
+            value = self.config[key]
+            diff = value.get_diff(existing.config[key])
+            if diff:
+                changes[key] = diff
+
+        if len(changes) > 0:
+            return DatabricksRelationChangeSet(changes=changes, requires_full_refresh=False)
+        return None
```
jackyhu-db, line 138: nit: `return DatabricksRelationChangeSet(changes=changes, requires_full_refresh=False) if len(changes) > 0 else None`

jackyhu-db (approved):

dbt/adapters/databricks/relation_configs/view.py
jackyhu-db: nit: can you add this comment into the code rather than the PR so people have more context when reading the code?

dbt/adapters/databricks/impl.py
```diff
@@ -918,3 +923,30 @@ def _describe_relation(
...
+        return ViewConfig
+
+    @classmethod
+    def _describe_relation(
+        cls, adapter: DatabricksAdapter, relation: DatabricksRelation
+    ) -> RelationResults:
+        results = {}
+        kwargs = {"relation": relation}
+
+        results["information_schema.views"] = get_first_row(
```
jackyhu-db, line 942: is `information_schema` also dict or namespace/prefix of the property/key name in the result dict?

dbt/adapters/databricks/impl.py
```diff
@@ -138,7 +139,9 @@ class DatabricksConfig(AdapterConfig):
     target_alias: Optional[str] = None
     source_alias: Optional[str] = None
     merge_with_schema_evolution: Optional[bool] = None
-    safe_table_create: Optional[bool] = None
+    use_safer_relation_operations: Optional[bool] = None
+    incremental_apply_config_changes: Optional[bool] = None
+    view_update_via_alter: Optional[bool] = None
```
eric-wang-1990, line 144: Do we want to mention these incremental_apply_config_changes,  view_update_via_alter usage in ChangeLog as well? Are these customer configured?

dbt/adapters/databricks/impl.py
```diff
@@ -724,7 +727,7 @@ def get_persist_doc_columns(
             if name in columns:
                 config_column = columns[name]
                 if isinstance(config_column, dict):
-                    comment = columns[name].get("description", "")
```
eric-wang-1990, line 727: Is this change expected? Not clear to me if we wanna return none or empty string.

eric-wang-1990 (approved):

dbt/adapters/databricks/relation_configs/view.py
benc-db: I'm thinking I might add a log statement that will include the information.

CHANGELOG.md
benc-db: this was an auto find-replace; I guess I should roll this bit back and have a line stating it was renamed in 954.  Thanks.

dbt/include/databricks/macros/relations/view/alter.sql
```diff
@@ -0,0 +1,22 @@
...
+  {{ log("Updating view via ALTER VIEW") }}
+  {% if configuration_changes %}
+    {% set tags = configuration_changes.changes.get("tags") %}
+    {% set tblproperties = configuration_changes.changes.get("tblproperties") %}
+    {% set query = configuration_changes.changes.get("query") %}
+    {% if tags %}
+      {% do apply_tags(target_relation, tags.set_tags, tags.unset_tags) %}
+    {%- endif -%}
+    {% if tblproperties %}
+      {% do apply_tblproperties(target_relation, tblproperties.tblproperties) %}
```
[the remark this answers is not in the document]
benc-db: tblproperties is what I named the change object when I read it out, then tblproperties is a field on that change object.  Some change objects have multiple fields, but it just so happens that the tblproperties change doesn't.  I could rename the change object after I read it out, or I could rename the field.  Any preference on which?

dbt/adapters/databricks/impl.py
benc-db: I could name these keys anything; for information_schema content, I have just been using the pattern of naming it essentially 'schema.table'.  So information_schema is not a separate dict, it's just a prefix to indicate this comes from the information_schema.

dbt/adapters/databricks/impl.py
benc-db: Yes.  I haven't added the changelog statement for this PR yet, but it is a good callout that I need to mention these.

dbt/adapters/databricks/impl.py
benc-db: I haven't validated it yet, but I have bug report that empty descriptions get persisted as "", because of equality mismatch.  I should validate that this actually fixes that bug, but yes, the change was intentional, to get a default of None instead of ""

benc-db: Will validate.

dbt/include/databricks/macros/relations/view/alter.sql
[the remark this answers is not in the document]
jackyhu-db: I see. it is OK, you can keep it as-is.

dbt/adapters/databricks/relation_configs/base.py
benc-db: When I rewrite as that, the width limits turn it into 5 lines, lol.  Going to keep this as is

PR #959, opened 2025-03-12
View V2 Testing

benc-db:
### Description

Makes small modifications to Views V2 to make it more testable and added unit tests of the macros.  Interestingly, I tried to update column comments with the same methodology as the other changes and backed it out because altering the query blows comments away.  This means that if I modify the query, I need to full set of comments, not just the changes. As such, I'm sticking with persist_docs and running after the query has potentially been updated.

### Checklist


dbt/include/databricks/macros/materializations/view.sql
```diff
@@ -14,11 +14,14 @@
           {{ exceptions.raise_compiler_error("Cannot update a view in the Hive metastore via ALTER VIEW. Please set `view_update_via_alter: false` in your model configuration.") }}
         {% endif %}
         {% set configuration_changes = get_configuration_changes(existing_relation) %}
-        {% if configuration_changes and not configuration_changes.requires_full_refresh %}
-          {{ alter_view(target_relation, configuration_changes) }}
-        {% else %}
-          {{ replace_with_view(existing_relation, target_relation) }}
+        {% if configuration_changes and configuration_changes.changes %}
+          {% if configuration_changes.requires_full_refresh %}
```
benc-db, line 18: Reordered this because because I realized I left out a condition -> alter view is requested, but configuration changes is empty.  The correct thing to do in that case is to check if we need to persist docs, but otherwise leave the view as is.

dbt/include/databricks/macros/relations/components/query.sql
```diff
@@ -0,0 +1,14 @@
+{% macro alter_query(target_relation, query) %}
```
benc-db, line 1: Moved here to match the other relation config components

tests/functional/adapter/views/test_views.py
```diff
@@ -14,6 +14,8 @@ def seeds(self):
     def models(self):
         return {"initial_view.sql": fixtures.view_sql, "schema.yml": fixtures.schema_yml}
 
+
```
benc-db, line 17: Had to split out into multiple classes because when I write out a file update, that file can get in the way of other tests.  By making them separate test classes, I ensure all files are isolated between tests.

dbt/adapters/databricks/relation_configs/query.py
```diff
@@ -15,14 +15,21 @@ class QueryConfig(DatabricksComponentConfig):
 
     query: str
 
+    def get_diff(self, other: "QueryConfig") -> Optional["QueryConfig"]:
```
benc-db, line 18: Realized this was broken when I started testing more deeply

dbt/adapters/databricks/relation_configs/query.py
```diff
@@ -15,14 +15,21 @@ class QueryConfig(DatabricksComponentConfig):
...
 
 class QueryProcessor(DatabricksComponentProcessor[QueryConfig]):
     name: ClassVar[str] = "query"
 
     @classmethod
     def from_relation_results(cls, result: RelationResults) -> QueryConfig:
-        row = result["information_schema.views"]
-        return QueryConfig(query=row["view_definition"])
+        view_definition = result["information_schema.views"]["view_definition"].strip()
+        if view_definition[0] == "(" and view_definition[-1] == ")":
```
benc-db, line 30: When we get the view definition from information_schema, it seems to include the optional parentheses being added by macro that formats the statement.

PR #968, opened 2025-03-19
Moving liquid clustering to the check for difference approach

benc-db:
### Description

For incremental models, liquid clustering column setting was running on every incremental run.  We can use the check for difference approach basically for free, since it reuses data we were already gathering, so moving to that.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/credentials.py
+++ b/dbt/adapters/databricks/credentials.py
@@ -1,7 +1,6 @@
 import itertools
 import json
 import re
-import threading
 from collections.abc import Iterable
 from dataclasses import dataclass, field
 from typing import Any, Callable, Optional, cast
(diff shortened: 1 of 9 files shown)


dbt/adapters/databricks/credentials.py
```diff
@@ -329,60 +327,54 @@ def authenticate_with_azure_client_secret(self) -> Config:
         )
 
     def __post_init__(self) -> None:
-        self._lock = threading.Lock()
```
benc-db, line 332: Removed the lock because it was rarely, but sometimes, breaking tests.  Essentially, dbt serializes the entire project, and if that happens after a lock has been set, it will error during the deepcopy.  Need to validate whether its safe to remove, but my thinking is that each thread will have its own copy of credentials, so we don't need to lock.

benc-db: Here in particular we shouldn't need the lock, since __post_init__ is called exactly once for creating the object.

PR #971, opened 2025-03-20
Adding more macro unit tests, cleaning up some macros

benc-db:
### Description

Using copilot to help me troubleshoot the very annoying to write macro tests, but I did discover some bugs while doing it.  Will need to chip away at this over time.  A few other minor fixes that I will comment on in the PR.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/credentials.py
+++ b/dbt/adapters/databricks/credentials.py
@@ -110,6 +110,7 @@ class DatabricksCredentials(Credentials):
             "catalog",
             "schema",
             "_user_agent_entry",
+            "user_agent_entry",
         ):
             if key in connection_parameters:
                 raise DbtValidationError(f"The connection parameter `{key}` is reserved.")
--- a/dbt/adapters/databricks/relation.py
+++ b/dbt/adapters/databricks/relation.py
@@ -183,6 +183,9 @@ class DatabricksRelation(BaseRelation):
         processed = map(process_constraint, self.create_constraints)
         return ", ".join(c for c in processed if c is not None)
 
+    def render(self) -> str:
+        return super().render().lower()
+
 
 def is_hive_metastore(database: Optional[str]) -> bool:
     return database is None or database.lower() == "hive_metastore"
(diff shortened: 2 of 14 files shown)


dbt/adapters/databricks/credentials.py
```diff
@@ -110,6 +110,7 @@ def __post_init__(self) -> None:
             "catalog",
             "schema",
             "_user_agent_entry",
+            "user_agent_entry",
```
benc-db, line 113: with PySQL 4.0.1, _user_agent_entry is being replaced with user_agent_entry.  Specifying both gets the warning to silence.

dbt/adapters/databricks/relation.py
```diff
@@ -183,6 +183,9 @@ def render_constraints_for_create(self) -> str:
         processed = map(process_constraint, self.create_constraints)
         return ", ".join(c for c in processed if c is not None)
 
+    def render(self) -> str:
```
benc-db, line 186: every place we render, I want to lower, because any time we compare against a name stored in metadata, it will be stored lower case.  Will remove the jinja calls to lower as I touch relevant files.

dbt/include/databricks/macros/adapters/catalog.sql
```diff
@@ -1,99 +1,95 @@
 {% macro databricks__get_catalog(information_schema, schemas) -%}
+  {% set query %}
```
benc-db, line 2: This file has some formatting improvements, but also a legitimate bug fix that I think just hasn't come up because I disabled querying by relations.  If I turn that back on (i.e. if information_schema perf improves), I want it to work, so fixed it after adding unit tests.

dbt/include/databricks/macros/adapters/columns.sql
```diff
@@ -1,29 +1,33 @@
 
 {% macro get_columns_comments(relation) -%}
```
benc-db, line 2: started structuring to SQL rendering separate from execution, since its easier to unit test the SQL rendering pieces on their own.  If I can figure out a good way to stub run_query and statement, I can start testing these outer methods too.

PR #992, opened 2025-04-09
Fix regression related to insert_overwrite

benc-db:
### Description

In 1.10.0, we starting failing users that previously were passing for INSERT OVERWRITE with SQL Warehouses.  They should have always failed, since without dynamic partition setting, INSERT OVERWRITE is just truncate + insert, which is no better than the table materialization.  Hence, we guard every call to this setting with a check for compute type, and if not cluster, raise an information error message.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 - Fix: DBT Python Model Canceled Notebook Job Treated as Successful ([985](https://github.com/databricks/dbt-databricks/pull/985))
 - Fix failures when models end with ';' ([990](https://github.com/databricks/dbt-databricks/pull/990))
 - Added internal table property delta.columnMapping.maxColumnId to ignore_list to allow enabling delta.columnMapping.mode without breaking concurrent job runs (thanks @
-iamatharkhan!) ([991](https://github.com/databricks/dbt-databricks/pull/991))
+  iamatharkhan!) ([991](https://github.com/databricks/dbt-databricks/pull/991))
+- For insert_overwrite, raise exception when using SQL Warehouse and inform users that such use causes truncate + insert ([992](https://github.com/databricks/dbt-databricks/pull/992))
 
 ## dbt-databricks 1.10.0 (Apr 08, 2025)
 
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -204,6 +204,9 @@ class DatabricksConnectionManager(SparkConnectionManager):
             )
         return self._api_client
 
+    def is_cluster(self) -> bool:
+        return self.get_thread_connection().credentials.cluster_id is not None
+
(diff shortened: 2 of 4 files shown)


benc-db:
> @benc-db does this mean warehouses can't do dynamic partitioning or just that insert_overwrite does not enforce dynamic partitioning?

> I'm currently working on a custom insert_overwrite that'll enforce partitioning. Just want to make sure it'll work.

With the current syntax, warehouses cannot do dynamic partitioning as trying to set the required spark setting errors.  With current strategy implementation, insert_overwrite needs dynamic partitioning to be practical, since as you hint at, we don't have a mechanism to express the static partitioning.  I think what you have in mind may work, but will also mention that there is something in the works that should supersede this issues sometime in the next several months.  So, if you have a more urgent need, I do recommend exploring what you can accomplish today; it may be possible.

kyle-kent:
@benc-db thanks for your quick response. I see the problem now. `spark.sql.sources.partitionOverwriteMode` is a spark config, not an arg for the insert command.

We do need a solution now which will require overwriting partitions with new data. The `replace_where` strat's incremental predicate can't be dynamic correct? So I'm leaning towards a custom incremental that does a `delete` then an `insert`/`merge` to append data. Any knowledge about performance you can throw my way?

Thanks!

NortySpock:
> INSERT OVERWRITE is just truncate + insert, which is no better than the table materialization.

We are trying to integrate dbt with Immuta, and currently we are trying to find a solution to a problem where the default table materialization (`CREATE OR REPLACE TABLE AS`) (_CORTAS_) causes the UDFs that are tied to a table to be dropped (because as I understand it technically you get a new table id when you _CORTAS_), which means data on the table is incorrectly secured until Immuta can regenerate the UDFs on the table. (It's still currently in the testing and proof-of-concept phase, so it's not disastrous, but it's really hindering implementation of Immuta at our business.)

`INSERT OVERWRITE` appeared to be atomic and didn't drop the table (or the UDFs), so it seemed like it would be a workable solution for us...

I guess I'll write a custom incremental materialization that bypasses this native dbt-databricks implementation for now, but is there a method we could use that does not drop the table when reloading the contents? (and can it be atomic?) Specifically in some cases I do not have a Primary Key on a table that I am reloading, so I cannot use the MERGE statement. (I might be able to put a primary key together based on a hash, but I was hoping not to have to write (and test, and verify) that...)

benc-db:
@NortySpock you have replace-where and append as options

https://docs.getdbt.com/reference/resource-configs/databricks-configs#the-replace_where-strategy
https://docs.databricks.com/aws/en/delta/selective-overwrite#arbitrary-selective-overwrite-with-replacewhere

NortySpock:
@benc-db  Thanks for the quick response. We will re-examine replace-where.

PR #995, opened 2025-04-15
Replace illegal characters in unique temporary table names

### LINKED ISSUE #994: Unique temp tables persist and throw warnings on session close
### Describe the bug

When using the `unique_tmp_table_suffix` config with incremental models, a temporary table is generated with a name like `dbt_task__dbt_tmp_a232cf9a-252c-40b4-8d10-f45e23011df4`. This table name contains the `-` illegal character that must be quoted for successful executions.

However, despite quoting of table identifiers set to `true`, this table name is not quoted when attempting to drop the table when the session closes leading to warnings and the temporary table ultimately not being dropped. This leaves temporary tables behind after every model execution which causes clutter in the catalog. 

### Steps To Reproduce

1. Create any model with an incremental materialization and the config `unique_tmp_table_suffix` enabled e.g.
  ```sql
  {{ config(
      materialized='incremental',
      unique_tmp_table_suffix=True,
  ) }}

  SELECT * FROM VALUES (1,2,3)
  ```

2. Run the model to create it initially.
3. Run the model again to incrementally add data to it. The model will be updated, but the temporary table storing the intermediary data will remain with warnings thrown in the log.

### Expected behavior

The temporary table should be dropped without warnings.

```
20:41:32  databricks-sql-connector adapter: Attempt to close session raised an exception at the server: 
[INVALID_IDENTIFIER] The unquoted identifier schema_changes__dbt_tmp_ce5713c1-86db-449b-b96a-1d2b12f56b30 is invalid and must be back quoted as: `schema_changes__dbt_tmp_ce5713c1-86db-449b-b96a-1d2b12f56b30`.
Unquoted identifiers can only contain ASCII letters ('a' - 'z', 'A' - 'Z'), digits ('0' - '9'), and underbar ('_').
Unquoted identifiers must also not start with a digit.
Different data sources and meta stores may impose additional restrictions on valid identifiers. SQLSTATE: 42602 (line 1, pos 32)

== SQL ==
schema_changes__dbt_tmp_ce5713c1-86db-449b-b96a-1d2b12f56b30
--------------------------------^^^
```

### System information
```
Core:
  - installed: 1.9.4
  - latest:    1.9.4 - Up to date!

Plugins:
  - databricks: 1.10.0 - Up to date!
```

MacOS 15.3.2

Python 3.12.9

geo-martino:
Resolves #994

### Description

Fixes a bug where temporary tables with unique identifiers would not be dropped on session close due to the table name containing illegal characters generated by the `uuid`.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Added internal table property delta.columnMapping.maxColumnId to ignore_list to allow enabling delta.columnMapping.mode without breaking concurrent job runs (thanks @
   iamatharkhan!) ([991](https://github.com/databricks/dbt-databricks/pull/991))
 - For insert_overwrite, raise exception when using SQL Warehouse and inform users that such use causes truncate + insert ([992](https://github.com/databricks/dbt-databricks/pull/992))
-- Remove illegal characters in unique temporary table names which prevented dropping these tables on session close ([995](https://github.com/databricks/dbt-databricks/pull/995))
+- Remove illegal characters in unique temporary table names which prevented dropping these tables on session close (thanks @geo-martino!) ([995](https://github.com/databricks/dbt-databricks/pull/995))
 
 ## dbt-databricks 1.10.0 (Apr 08, 2025)
 
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -753,8 +753,9 @@ class DatabricksAdapter(SparkAdapter):
         return return_columns
 
     @available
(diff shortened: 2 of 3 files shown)


benc-db:
Hmm...this probably means we're missing a test.  If you could add a test, that would be nice.  I'm going to run the functional tests to validate that nothing else is disrupted by this change.  Thank you :)

geo-martino:
> Hmm...this probably means we're missing a test. If you could add a test, that would be nice. I'm going to run the functional tests to validate that nothing else is disrupted by this change. Thank you :)

Sure thing, added a couple of tests for this method

benc-db (approved):

PR #999, opened 2025-04-16
Drop pre-existing backup name from the adapter's cache

geo-martino:
Resolves #

### Description

When creating a backup relation, an error is thrown if the backup relation already exists. While the pre-existing relation is safely dropped, it's name is not removed from the cache. This causes an error to be thrown when attempting rename the source relation to the backup relation name.

Suggest using the `drop_relation_if_exists` macro instead of `drop_relation` which utilises the `drop_relation` adapter method hence dropping the table name from the cache.

### Checklist


dbt/include/databricks/macros/relations/create_backup.sql
```diff
@@ -3,7 +3,7 @@
   {% set backup_relation = make_backup_relation(relation, relation.type) %}
 
   -- drop any pre-existing backup
-  {{ drop_relation(backup_relation) }}
+  {{ drop_relation_if_exists(backup_relation) }}
```
benc-db, line 6: the previous drop doesn't interact with the cache?  I agree that this is more correct, just surprised.

benc-db (dismissed):

benc-db:
Will run E2E tests after PR 1000

PR #1012, opened 2025-05-07
Migrate to the dbt catalog framework

mikealfare:
Resolves #

### Description

dbt will support catalogs and catalog integrations as stand-alone objects, just like relations.
Since `dbt-databricks` already had the concept of catalogs, there is some work to migrate the existing concept into the generic catalog concept. This PR attempts to migrate to the new framework while maintaining backwards compatibility. Points of discussion:

- We need to support existing user behavior for all macros and all methods made available (`@available`) on the adapter in the jinja context.
- We make the assumption that every relation has a corresponding catalog integration, and a catalog relation within this catalog integration as a result. This means we need to supply a "default" catalog integration for all existing materializations that use tables (e.g. `table`, `incremental`, `snapshot`, `seed`).
- We are not supporting python models yet.
- dbt combines the concepts of a metadata catalog (e.g. Iceberg), a file format (e.g. Parquet), and an external volume (e.g. DBX's `location`). Models can override this if needed, but the idea is that the catalog integration can carry defaults for these settings.

### Checklist


dbt/adapters/databricks/catalogs/_delta.py
```diff
@@ -0,0 +1,42 @@
...
+    allows_writes = True
+
+    def __init__(self, config: CatalogIntegrationConfig) -> None:
+        super().__init__(config)
+        if location_root := config.adapter_properties.get("location_root"):
+            self.external_volume: Optional[str] = location_root
+        self.table_format: str = config.adapter_properties.get("table_format")
+        self.file_format: str = config.adapter_properties.get("file_format")
+
+    @property
```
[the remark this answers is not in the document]
mikealfare: Could you elaborate?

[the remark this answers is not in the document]
mikealfare: Oh, I think I see. The comment showed up next to `table_format` and `file_format` on the thread. Yeah, I'm not sure how to handle the `location_root`/`external_volume` bit. We elected to go with `external_volume` as the canonical identifier for the location for the data. In DBX, that looks like the base for the `location` config. This is an attempt to allow folks to speak in DBX parlance while still meeting the generic pattern.

dbt/adapters/databricks/catalogs/__init__.py
```diff
@@ -0,0 +1,9 @@
+from dbt.adapters.databricks.catalogs._hive_metastore import HiveMetastoreCatalogIntegration
+from dbt.adapters.databricks.catalogs._relation import DatabricksCatalogRelation
+from dbt.adapters.databricks.catalogs._unity import UnityCatalogIntegration
+
+__all__ = [
```
benc-db, line 5: So, this might be a really naive python question, but what does this do?

dbt/adapters/databricks/catalogs/_hive_metastore.py
```diff
@@ -0,0 +1,42 @@
+from typing import Optional
```
benc-db, line 1: Why are these files named with underscore?  What are we trying to hide :)?

dbt/adapters/databricks/constants.py
```diff
@@ -0,0 +1,38 @@
...
+DELTA_FILE_FORMAT = "delta"
+PARQUET_FILE_FORMAT = "parquet"
+HUDI_FILE_FORMAT = "hudi"
+
+
+UNITY_CATALOG_TYPE = "unity"
+HIVE_METASTORE_CATALOG_TYPE = "hive_metastore"
+
+
+DEFAULT_UNITY_CATALOG = SimpleNamespace(
```
benc-db, line 16: What is this class?

dbt/adapters/databricks/parse_model.py
```diff
@@ -0,0 +1,48 @@
+import posixpath
+from typing import Optional
+
+from dbt.adapters.contracts.relation import RelationConfig
+from dbt.adapters.databricks import constants
+
+
+def catalog_name(model: RelationConfig) -> Optional[str]:
```
benc-db, line 8: I think the standard that we are trying to move towards with functions is to encapsulate in a class to act as the namespace, e.g.: https://github.com/databricks/dbt-databricks/blob/a6d02ceb72f4af30279143b7500fd052d6b069aa/dbt/adapters/databricks/handle.py#L264C1-L264C16

In this case though, since you make several of these calls in succession on the same object, maybe it should look more like this: https://github.com/databricks/dbt-databricks/blob/a6d02ceb72f4af30279143b7500fd052d6b069aa/dbt/adapters/databricks/connections.py#L94

benc-db (approved):
@mikealfare take at look at the feedback.  Most of it is just me asking about python features I don't know yet :).  Generally approve.

dbt/adapters/databricks/catalogs/__init__.py
mikealfare: It's the list of what will be imported when you use `from dbt.adapters.databricks.catalogs import *`. But the reason I'm doing it is because it's a standard way of passing `mypy` checks when "raising" objects in an `__init__.py` like this. `mypy` doesn't like it because it's being imported and not used. You could also add an ignore of this type of violation, but that's a project level thing. I would do that if you're alright with it. Here's an [example](https://github.com/dbt-labs/dbt-adapters/blob/4dc395b42dae78e895adf9c66ad6811534e879a6/.pre-commit-config.yaml#L40).

dbt/adapters/databricks/catalogs/_hive_metastore.py
mikealfare: This gives us some room to move things around or rename if needed. The pattern I lean towards is to define all of the modules as private and then raise up the public objects in `__init__.py`. Then whatever is not exposed (directly or indirectly) in `__init__.py` is free game. As a trivial example, if we decide we need to call this file `_hive_metastore_v1.py` for some reason, we can do that, and then handle the update in the `__init__.py`. In particular, if we wanted to be able to support `_hive_metastore_v1.py` and `_hive_metastore_v2.py` based on some conditional, we could do that here. This is a very constrived example, but that's the idea.

dbt/adapters/databricks/constants.py
mikealfare: There are a few "out of the box" catalogs in each adapter. For example, many databases have the database itself (generally called info_schema) and then a managed Iceberg catalog. The user should not need to define these catalogs since they are the same for everyone. They are effectively defaults. Then if the user wants to provide their own catalog, they can do so via `catalogs.yml`. `DEFAULT_UNITY_CATALOG` is the catalog that will get used if the user does not specify a catalog and they're not using HMS.

dbt/adapters/databricks/catalogs/_hive_metastore.py
benc-db: hmm, but the vast majority of users do not use this library via a python import, and if they're are going to those lengths, then they have the power to import whatever they want because python is awful :P.

dbt/adapters/databricks/constants.py
benc-db: I meant what is SimpleNamespace/what are we using it for?

dbt/adapters/databricks/parse_model.py
benc-db: The functions themselves.  I guess in the end it doesn't really matter, and at least you don't import them raw.

dbt/adapters/databricks/catalogs/_hive_metastore.py
mikealfare: Agreed on many levels, however this also serves as an indicator to ourselves of what we can and can't change. It serves as a way to encapsulate functionality within the adapter. I'm completely fine removing the leading underscore (since we know that won't affect anything anyway). But that was my rationale. Let me know if you want to rename them and I'll push an update.

dbt/adapters/databricks/constants.py
mikealfare: Oh, that's just a built-in that effectively turns a dictionary into a class with attributes that can be accessed via dot notation. This config object needs to meet the `CatalogIntegrationConfig` [protocol](https://github.com/dbt-labs/dbt-adapters/blob/4dc395b42dae78e895adf9c66ad6811534e879a6/dbt-adapters/src/dbt/adapters/catalogs/_integration.py#L8-L43). The easiest way to do that is with a `SimpleNamespace`. I think of it as a step between a `TypedDict` and a `dataclass`.

dbt/adapters/databricks/parse_model.py
mikealfare: Oh I see. The issue would be that something like `catalog_name` could be a variable in the importing module because of how common a name it is. And then you need to deal with shadowing variables and whatnot. Yeah, that's part of why I put it in a separate module and import the entire module. I'm using the module in the same way that you are using the class. Either is fine; I just avoid classes when possible. This particular functionality is specific to each adapter; we're not inheriting anything from the base adapter. If you feel like we should update this to be in a class we could do that. Would it be something like `ParsedModel.catalog_name(cls, model)`, `ParsedModel.file_format(cls, model)`, etc.?

dbt/adapters/databricks/catalogs/_hive_metastore.py
benc-db: I'm fine with this.  Maybe if dbt Labs starts contributing more, our styles will converge :)

dbt/adapters/databricks/parse_model.py
benc-db: I'm happy to sign off as is, because like you say, you're importing the module.  It's way worse when raw functions are imported.

benc-db:
@mikealfare please update changelog and then we can merge.  I'm fine with the minor stylistic things.

Issue #982, opened 2025-04-04
Contract Enforcement and Constraints in incremental models


meretri:
### Describe the bug
I have defined a table with contracts enforced and a PK Constraint + a not_null constraint on one column. the materialization is incremental. this works well and I can see the constraint are being added to the table. if I then remove a column from the model.yml (but not the SQL), I can still run the model even though the contract is not valid anymore? Also if I add a constraint to the model.yaml after the first time I have run the model, the constraint won't be created on the table.

### Expected behavior
Contracts should be enforced also on incremental runs and there should be an option to add a constraint even if the table already exists.

PR #1013, opened 2025-05-09
Constraints for incremental materialization

ericj-db:
Resolves constraint part of #982. Does not resolve the contract enforcement 

### Description

Support constraints for incremental materializations. Databricks supports the following 4 constraints based on https://docs.databricks.com/aws/en/tables/constraints
- Non-null (enforced, requires UC)
- Check (enforced, does not require UC)
- Primary key (not enforced, requires UC)
- Foreign key (not enforced, requires UC)

### Checklist

### DIFF
--- a/dbt/adapters/databricks/relation_configs/constraints.py
+++ b/dbt/adapters/databricks/relation_configs/constraints.py
@@ -1,5 +1,5 @@
 from dataclasses import asdict
-from typing import ClassVar, Optional, cast
+from typing import ClassVar, Optional
 
 from agate import Table
 
@@ -22,29 +22,23 @@ from dbt.adapters.relation_configs.config_base import RelationResults
 class ConstraintsConfig(DatabricksComponentConfig):
     """Component encapsulating the constraints of a relation."""
 
-    set_non_nulls: list[str]
-    unset_non_nulls: list[str] = []
-    set_constraints: list[TypedConstraint]
-    unset_constraints: list[TypedConstraint] = []
-
     class Config:
         arbitrary_types_allowed = True
 
+    set_non_nulls: set[str]
+    unset_non_nulls: set[str] = set()
+    set_constraints: set[TypedConstraint]
+    unset_constraints: set[TypedConstraint] = set()
+
     def get_diff(self, other: "ConstraintsConfig") -> Optional["ConstraintsConfig"]:
-        constraints_to_unset = []
-        non_nulls_to_unset = []
         # Find constraints that need to be unset
-        for constraint in other.set_constraints:
-            if constraint not in self.set_constraints:
-                constraints_to_unset.append(constraint)
+        constraints_to_unset = other.set_constraints - self.set_constraints
         # Find non-nulls that need to be unset
-        for non_null in other.set_non_nulls:
-            if non_null not in self.set_non_nulls:
-                non_nulls_to_unset.append(non_null)
+        non_nulls_to_unset = other.set_non_nulls - self.set_non_nulls
 
         # Set constraints that exist in self but not in other
-        set_constraints = [c for c in self.set_constraints if c not in other.set_constraints]
-        set_non_nulls = [n for n in self.set_non_nulls if n not in other.set_non_nulls]
+        set_constraints = self.set_constraints - other.set_constraints
+        set_non_nulls = self.set_non_nulls - other.set_non_nulls
 
         if set_constraints or set_non_nulls or constraints_to_unset or non_nulls_to_unset:
             return ConstraintsConfig(
@@ -62,12 +56,12 @@ class ConstraintsProcessor(DatabricksComponentProcessor[ConstraintsConfig]):
     @classmethod
     def _process_check_constraints(
         cls, check_constraints_table: Optional[Table]
-    ) -> list[CheckConstraint]:
-        check_constraints = []
+    ) -> set[CheckConstraint]:
+        check_constraints = set()
         if check_constraints_table:
             for row in check_constraints_table.rows:
                 if row[0].startswith("delta.constraints."):
-                    check_constraints.append(
+                    check_constraints.add(
                         CheckConstraint(
                             type=ConstraintType.check,
                             name=row[0].replace("delta.constraints.", ""),
@@ -79,8 +73,8 @@ class ConstraintsProcessor(DatabricksComponentProcessor[ConstraintsConfig]):
     @classmethod
     def _process_primary_key_constraints(
         cls, pk_constraints_table: Optional[Table]
-    ) -> list[PrimaryKeyConstraint]:
-        pk_constraints = []
+    ) -> set[PrimaryKeyConstraint]:
+        pk_constraints = set()
         if pk_constraints_table:
             # Group columns by constraint name
             constraint_columns: dict[str, list[str]] = {}
@@ -93,7 +87,7 @@ class ConstraintsProcessor(DatabricksComponentProcessor[ConstraintsConfig]):
 
             # Create one PrimaryKeyConstraint per unique constraint name
             for constraint_name, columns in constraint_columns.items():
-                pk_constraints.append(
+                pk_constraints.add(
                     PrimaryKeyConstraint(
                         type=ConstraintType.primary_key, name=constraint_name, columns=columns
                     )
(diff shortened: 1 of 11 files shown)


dbt/adapters/databricks/relation_configs/constraints.py
```diff
@@ -0,0 +1,74 @@
...
+            {**asdict(col), "constraints": [c.to_dict() for c in col.constraints]}
+            for col in columns.values()
+        ]
+
+        non_nulls, other_constraints = parse_constraints(columns, constraints)
+
+        return ConstraintsConfig(
+            set_non_nulls=list(non_nulls),
+            set_constraints=other_constraints,
+        )
```
ericj-db, line 74: @benc-db does this look reasonable? When I run the functional test on debug mode and set a breakpoint here, the objects look correct. The RelationConfig contains `ModelLevelConstraint` and `ColumnLevelConstraint` objects which are defined in dbt-common. Here, we are transforming them to Databricks internal representations defined in https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/constraints.py

dbt/adapters/databricks/relation_configs/constraints.py
```diff
@@ -0,0 +1,74 @@
...
+    DatabricksComponentProcessor,
+)
+from dbt.adapters.relation_configs.config_base import RelationResults
+
+
+class ConstraintsConfig(DatabricksComponentConfig):
+    """Component encapsulating the constraints of a relation."""
+
+    set_non_nulls: list[str]
+    unset_non_nulls: list[str] = []
```
ericj-db, line 20: It looks like we don't have a dedicated class for non-null constraints. I see it's represented as a list of column names in https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/constraints.py. This is just following that

benc-db: That's fine.  It's probably because they get applied differently than the other constraints.  You can either try to unify by adding a class, or handling it like this is also fine.

dbt/adapters/databricks/relation_configs/constraints.py
```diff
@@ -0,0 +1,74 @@
...
+    set_constraints: list[TypedConstraint]
+    unset_constraints: list[TypedConstraint] = []
+
+    class Config:
+        arbitrary_types_allowed = True
+
+    def get_diff(self, other: "ConstraintsConfig") -> Optional["ConstraintsConfig"]:
+        constraints_to_unset = []
+        non_nulls_to_unset = []
+        for constraint in other.set_constraints:
```
benc-db, line 30: It looks like you're only checking the diff in one direction, but we want to check it in both - i.e. if the constraint is in both, it does not need to be in the config we return, because it already exists in Databricks.

dbt/adapters/databricks/relation_configs/constraints.py
benc-db: Without digging in deep, this looks correct on first glance.  Presumably if there is something off, it will show up during unit testing.

dbt/adapters/databricks/relation_configs/constraints.py
```diff
@@ -0,0 +1,184 @@
...
+            {"name": name, **(col if isinstance(col, dict) else asdict(col))}
+            for name, col in columns.items()
+        ]
+
+        non_nulls, other_constraints = parse_constraints(columns, constraints)
+
+        return ConstraintsConfig(
+            set_non_nulls=list(non_nulls),
+            set_constraints=other_constraints,
+        )
```
ericj-db, line 184: @benc-db when I debugged functional tests, the nested objects we were dealing with were always ModelNode, ColumnInfo, etc. But I saw in some places, we also try to handle dicts. That's what I tried to do here. Is this actually necessary?

benc-db: I wish I could be confident about that.  This is one of those places that python sucks.  If you don't check for the dict case and it works both in tests and in a real dbt project, then you're good.  I have it in the other places because I hit cases where the type wasn't what I expected at run time.

tests/functional/adapter/incremental/test_incremental_constraints.py
```diff
@@ -0,0 +1,246 @@
+import pytest
```
benc-db, line 1: Really well done test suite for this feature!

dbt/adapters/databricks/relation_configs/constraints.py
```diff
@@ -0,0 +1,184 @@
...
+class ConstraintsConfig(DatabricksComponentConfig):
+    """Component encapsulating the constraints of a relation."""
+
+    set_non_nulls: list[str]
+    unset_non_nulls: list[str] = []
+    set_constraints: list[TypedConstraint]
+    unset_constraints: list[TypedConstraint] = []
+
+    class Config:
+        arbitrary_types_allowed = True
```
benc-db, line 31: What is this needed for?

dbt/adapters/databricks/relation_configs/constraints.py
```diff
@@ -0,0 +1,184 @@
...
+)
+from dbt.adapters.relation_configs.config_base import RelationResults
+
+
+class ConstraintsConfig(DatabricksComponentConfig):
+    """Component encapsulating the constraints of a relation."""
+
+    set_non_nulls: list[str]
+    unset_non_nulls: list[str] = []
+    set_constraints: list[TypedConstraint]
```
benc-db, line 27: should these not be set, rather than list?

dbt/include/databricks/macros/relations/components/constraints.sql
```diff
@@ -0,0 +1,135 @@
...
+        AND table_name = '{{ relation.identifier|lower }}'
+        AND constraint_type = 'FOREIGN KEY'
+    )
+  ORDER BY kcu.ordinal_position;
+{%- endmacro -%}
+
+{% macro apply_constraints(relation, constraints) -%}
+  {{ log("Applying constraints to relation " ~ constraints) }}
+  {%- if constraints and relation.is_hive_metastore() -%}
+    {{ exceptions.raise_compiler_error("Constraints are only supported for Unity Catalog") }}
```
benc-db, line 88: This compiler error is not quite right, though its one of those errors that should never be hit, right?  Because if the table is HMS, we would have never build the change list.  So, I think we actually want the error to indicate 2 things: 1.) incremental application of constraints is not supported for HMS, and 2.) if we reach this line the user should report a bug because the code should be unreachable.  Now, if it should be unreachable, why even check here?  The one use case I could see is if someone tries to call this macro from a custom materialization.  So for that reason, I think we can keep this check, but let's be more precise about the nature of the issue.

benc-db (changes requested):
Just a few minor changes, otherwise really good.  Make sure to add a changelog entry.

dbt/adapters/databricks/relation_configs/constraints.py
ericj-db: yeah that makes more sense, updated in ebcfe334cd7f7148c514c358b8883abab84f4127

dbt/adapters/databricks/relation_configs/constraints.py
ericj-db: I was getting this error without it. I don't actually really understand what it means tbh
```
RuntimeError: no validator found for <class 'dbt.adapters.databricks.constraints.TypedConstraint'>, see `arbitrary_types_allowed` in Config
```

dbt/adapters/databricks/relation_configs/constraints.py
ericj-db: Makes sense, addressed in 556de3fcd869a2d8299a3ee2c082a8c7f8ec61b7

dbt/include/databricks/macros/relations/components/constraints.sql
ericj-db: Got it, updated in 7faee05f8bffd18a37160210f23896e36db1ab7c. I didn't specify that this should be reported as an issue because you mentioned that it's possible to be called by custom materializations

dbt/adapters/databricks/relation_configs/constraints.py
benc-db: ah.  Is this config supported in Pydantic v1?  For compat reasons, dbt has asked us to not enforce using v2.

That error looks like it is complaining that TypedConstraint is not a pydantic model (I think it's a dataclass?)  We could probably convert to a pydantic model, since it's basically a dataclass with extras.

ericj-db: Actually it seems like TypedConstraint wasn't annotated as a dataclass to begin with. I annotated it and was able to get rid of this code. Ref: e7099a2c75a92fa5240989b19d246a0ca8a313ef

benc-db (approved):

PR #1026, opened 2025-05-16
Fix: drop temp tables in python model materialization v2

### LINKED ISSUE #1010: `dbt_tmp` tables not dropped for Python models
### Describe the bug
When running Python models, temporary tables with the `dbt_tmp` suffix are created as part of the replace flow. The temporary tables should be dropped after adding their data to the main table but I'm finding these tables are left behind causing a massive build up of these tables in our catalog. I don't see this behaviour for SQL models at all.

I haven't had much time to look at some edge cases and see if any certain config is affecting this (e.g. using `unique_tmp_table_suffix`) but I should have time to investigate further next week if needed.

### Steps To Reproduce
I have the following relevant config set in my project file:
```yaml
flags:
  use_materialization_v2: yes
  use_user_folder_for_python: yes

models:
  +unique_tmp_table_suffix: yes  # needed to run updates to tables concurrently
  +view_update_via_alter: yes
```

This bug should happen upon generating any python model for a relation which needs to be replaced.

### Expected behavior
The temporary tables should be dropped after adding their data to the main table

Image

### System information
```
Core:
  - installed: 1.9.4
  - latest:    1.9.4 - Up to date!

Plugins:
  - databricks: 1.10.1 - Up to date!
```

MacOS 15.4.1

Python 3.12.10

### Additional context
We are executing the dbt command from an all-purpose cluster which also handles the execution of of the Python models too.

ericj-db:
Resolves #1010

### Description

I was able to repro in materialization v2 for table and incremental
- For incremental, the cleanup logic is nested in a way that only temp tables created on subsequent runs are cleaned up. The one created by the first run lives forever
- For tables, the cleanup logic does not exist at all

This PR fixes both with a call to `drop_relation_if_exists` when language is python

### Checklist


dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -73,6 +73,10 @@
     {% do apply_grants(target_relation, grant_config, should_revoke) %}
     {% do optimize(target_relation) %}
 
+    {%- if language == 'python' -%}
+      {{ drop_relation_if_exists(staging_relation) }}
+    {%- endif -%}
+
```
ericj-db, line 79: @benc-db I think I am hitting the problem you mentioned in
https://github.com/databricks/dbt-databricks/blob/48624cff40cedcf6392a06a9aa8fe23acabb59dc/dbt/include/databricks/macros/materializations/incremental/incremental.sql#L139-L148

I tried `adapter.drop_relation` as well and it doesn't work. Did you ever figure out why this doesn't work? So confused because under the hood, it seems like `statement_with_staging_table` ultimately just calls the same thing https://github.com/databricks/dbt-databricks/blob/48624cff40cedcf6392a06a9aa8fe23acabb59dc/dbt/adapters/databricks/impl.py#L307-L309

ericj-db: Actually it seems to work with `drop_relation_if_exists` if I remove the python language check.. I wonder why it's not matching

benc-db: That comment is from before my time lol.  It might even exist in dbt-spark and was just copied over.

benc-db:
> databricks__make_intermediate_relation macro (which was not being used anywhere)

This would not be in use only if 'make_intermediate_relation' is not in use, due to the macro delegation pattern that dbt uses.

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -21,7 +21,7 @@
     {% set should_replace = existing_relation.is_dlt or existing_relation.is_view or full_refresh %}
     {% set is_replaceable = existing_relation.can_be_replaced and is_delta and config.get("location_root") %}
 
-    {% set intermediate_relation = make_intermediate_relation(target_relation) %}
```
benc-db, line 24: I'm surprised this wasn't already delegating to databricks__make_intermediate_relation.  Do we know why?  Did I do something dumb like overwrite make_intermediate_relation directly?

benc-db (commented):
Probably approved, but I would like to understand why my mental model is being broken here, since the change you introduced should be a no-op in my model

dbt/include/databricks/macros/materializations/incremental/incremental.sql
ericj-db: I'm not sure why, I actually didn't even know that the delegation should be automatic. Just for my own learning, where does that delegation logic live?

benc-db (dismissed):
Add Changelog, otherwise approve

PR #1033, opened 2025-05-28
Add column mask support for tables and incremental

### LINKED ISSUE #670: Feature Request: Support Column Masking
### Describe the feature
Support Column Masking as part of the YAML definition:

```sql
{% macro create_hash_function() %}
    CREATE OR REPLACE FUNCTION my_catalog.my_schema.hash_mask(raw_value STRING)
    RETURN sha2(raw_value, 256);
{% endmacro %}
```

```yaml

on-run-start: "{{ create_hash_function() }}"

models:
  - name: foo
    description: I'm a teapot
    columns:
      - name: id
      - name: bank_details
        column_mask: my_catalog.my_schema.hash_mask
```

Doing this in a post-hook. It works fine but it's a bit messy,

### Additional context
Please include any other relevant context here.

### Who will this benefit?
Anyone who wants to define masking for their models

Yes

ericj-db:
Resolves #670

### Description

- Introduce a new column-level config: `column_mask`
- Supported for tables and incremental. Throws compiler error for views and materialized views

#### New config example

```
models:
  - name: column_mask_sql
    columns:
        - name: id
        - name: password
          column_mask:
            function: mask_function
            using_columns: "id, 'hello', 5"
```

Executed SQL: 
```
ALTER TABLE column_mask_sql ALTER COLUMN password SET MASK mask_function USING COLUMNS(id, 'hello', 5)
```

Note that `id` is a reference to the column `id` of the same table. `'hello'` is a string literal

#### Limitations

When executing the following steps for incremental materializations, I ran into issues
1. Create a function: `mask_function(value STRING)`
2. Create a model using this function to mask a column
3. Run the model
4. Update the function: `mask_function(value STRING, another_value STRING)`
5. Update the model, now specifying `using_columns` to pass in the extra argument
6. Run the model

The above will fail because applying the changeset (i.e. executing the `ALTER TABLE` statement to update the mask) is done later than the first `SELECT` statement that runs against the table. This results in an error like
```
[WRONG_NUM_ARGS.WITHOUT_SUGGESTION] The `catalog`.`schema`.`mask_function` requires 2 parameters but the actual number is 1.
```

Looking at the macros, handling this edge case seems like it will need significant refactors. Not sure if it is worth the effort, so I have instead added a runtime error that is thrown when detecting that an existing function's signature has changed

UPDATE: we will not try to do anything fancy to handle this edge case. We consider this a user error as changing the function signature puts the table in a broken state. We will document this clearly in the doc update

### Checklist


dbt/include/databricks/macros/relations/table/create.sql
```diff
@@ -12,6 +12,7 @@
 
   {{ apply_alter_constraints(target_relation) }}
   {{ apply_tags(target_relation, tags) }}
+  {{ apply_column_masks_from_model_columns(target_relation) }}
```
ericj-db, line 15: Since this is not performed as part of the same transaction as the table creation, I believe there could be a period of time where the masks are not applied. Based on https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-column-mask, it is technically possible to apply masks at table creation

If this is preferred, I could make that change

benc-db: Yes, on create we should apply at create time.  That's the reason we are restricting the feature to Mat v2, because in v1, the presence of 'select' prevents us from doing column stuff.

dbt/adapters/databricks/column.py
```diff
@@ -48,6 +51,10 @@ def render_for_create(self) -> str:
         if self.comment:
             comment = self.comment.replace("'", "\\'")
             column_str += f" COMMENT '{comment}'"
+        if self.mask:
+            column_str += f" MASK {self.mask['function']}"
+            if "using_columns" in self.mask:
+                column_str += f" USING COLUMNS ({self.mask['using_columns']})"
```
ericj-db, line 57: The changes in this file are sufficient to get the masks correctly applied as part of the create statement. This is because `parse_columns_and_constraints` calls this function

@benc-db Actually, on that note: if we go with the current approach in https://github.com/databricks/dbt-databricks/pull/1046 of using `parse_columns_and_constraints` for MV/STs, then those would also automatically support column masks without having to write custom macros (which I have not done in this PR)

dbt/include/databricks/macros/relations/table/create.sql
ericj-db: Addressed this in d62217fb185f07715a265539f219551380023f61

dbt/adapters/databricks/column.py
benc-db: The only issue with that is that alter column (for column mask changes) is only in PuPr for MV/ST.  If we don't already have a task for this in 1.11.0 epic, let's add one.

dbt/include/databricks/macros/relations/components/column_mask.sql
```diff
@@ -0,0 +1,80 @@
...
+  {% endcall %}
+  {% do return(load_result('list_column_masks').table) %}
+{%- endmacro -%}
+
+{% macro fetch_column_masks_sql(relation) -%}
+  SELECT 
+    column_name,
+    mask_name,
+    using_columns
+  FROM `{{ relation.database|lower }}`.information_schema.column_masks
```
benc-db: should probably just be `system` for consistency

dbt/include/databricks/macros/relations/components/column_mask.sql
```diff
@@ -0,0 +1,80 @@
...
+    {%- for column, mask in column_masks.set_column_masks.items() -%}
+      {%- call statement('main') -%}
+        {{ alter_set_column_mask(relation, column, mask) }}
+      {%- endcall -%}
+    {%- endfor -%}
+  {%- endif %}
+{%- endmacro -%}
+
+{% macro alter_drop_column_mask(relation, column) -%}
+  ALTER TABLE {{ relation.render() }}
```
benc-db: Since column masks are supported on MV/ST in Databricks (even though we're not going to support them quite yet), we should not hard code the relation type here.

dbt/include/databricks/macros/relations/materialized_view/create.sql
```diff
@@ -1,4 +1,8 @@
 {% macro databricks__get_create_materialized_view_as_sql(relation, sql) -%}
+  {# Column masks result in silent failure for materialized views, so throw custom compiler error #}
+  {% if column_mask_exists() %}
+    {% do exceptions.raise_compiler_error("Column masks are not supported for materialized views. Column '" ~ column_name ~ "' has a mask defined.") %}
```
benc-db: Why is the failure silent?  It looks like column mask is supported per our docs.  If this is a silent failure, we should ask DLT team what's going on.

dbt/include/databricks/macros/relations/view/create.sql
```diff
@@ -1,4 +1,8 @@
 {% macro databricks__create_view_as(relation, sql) %}
+  {# Column masks result in silent failure for views, so throw custom compiler error #}
+  {% if column_mask_exists() %}
```
benc-db, line 2: Speaking of things we should raise, it doesn't look like column masks are supported for views, but we should raise an Aha ticket asking why, since conceptually if we support them on MV, they should also make sense for views.

ericj-db: Screenshot 2025-06-03 at 10 21 56 AM

Actually looks like it is supported lol. I guess the documentation is just outdated?

dbt/include/databricks/macros/relations/materialized_view/create.sql
ericj-db: My bad, I confused myself. It actually is supported and works as expected. Updated the comment for accuracy

dbt/include/databricks/macros/relations/components/column_mask.sql
ericj-db: Thanks for catching this, just updated

dbt/adapters/databricks/column.py
ericj-db: Sounds good, just filed https://databricks.atlassian.net/browse/PECO-2500

dbt/include/databricks/macros/relations/components/column_mask.sql
```diff
@@ -0,0 +1,65 @@
...
+
+{% macro alter_drop_column_mask(relation, column) -%}
+  ALTER {{ relation.type }} {{ relation.render() }}
+  ALTER COLUMN {{ column }}
+  DROP MASK;
+{%- endmacro -%}
+
+{% macro alter_set_column_mask(relation, column, mask) -%}
+  ALTER {{ relation.type }} {{ relation.render() }}
+  ALTER COLUMN {{ column }}
```
benc-db: Does this backtick escape?  If not, we should, at least if quote is set on the column.

dbt/include/databricks/macros/relations/components/column_mask.sql
```diff
@@ -0,0 +1,65 @@
...
+  SET MASK {{ mask.function }}
+  {%- if mask.using_columns %}
+  USING COLUMNS ({{ mask.using_columns }})
```
benc-db, line 54: Same as above, are these backticked?

dbt/include/databricks/macros/relations/materialized_view/create.sql
```diff
@@ -1,4 +1,8 @@
 {% macro databricks__get_create_materialized_view_as_sql(relation, sql) -%}
+  {# Column masks are supported in DBSQL, but not yet wired up to the adapter. Return a helpful error until supported. #}
+  {% if column_mask_exists() %}
+    {% do exceptions.raise_compiler_error("Column masks are not supported for materialized views.") %}
```
benc-db: maybe say 'not yet', to indicate that yes, we know, it is in PuPr

dbt/include/databricks/macros/relations/view/create.sql
benc-db: If it is supported for views, can we add this to the view relation_config and get rid of this blocker?

benc-db (changes requested):
A few more minor changes requested

dbt/include/databricks/macros/relations/view/create.sql
ericj-db: Oh wait I just realized my experiment was wrong. I was applying the mask on the source (after the `AS` clause) instead of the on the view. I get the following error when trying to apply it to the view definition
```
[SYNTAX_DISCONTINUED.INVALID_VIEW_COLUMN_SPEC] Support of the clause or keyword: MASK has been discontinued in this context. CREATE VIEW supports column lists with column names and column comments.
Do not use column type-definition, NOT NULL, or DEFAULT.
```

dbt/include/databricks/macros/relations/components/column_mask.sql
ericj-db: Thanks for catching, it wasn't backticked. Addressed in cd9ef7521e59b794b2433d2d43f9aeab1a9b9d6e

dbt/include/databricks/macros/relations/components/column_mask.sql
ericj-db: Also addressed in cd9ef7521e59b794b2433d2d43f9aeab1a9b9d6e

benc-db (dismissed):

PR #1035, opened 2025-05-28
Fix insert overwrite compile check when using model level compute override

### LINKED ISSUE #1032: Invalid exception raised on insert overwrite
### Describe the bug
A recent change in v1.10.1 started raising an exception for insert overwrite materialization even though compute is overwritten to use interactive compute for that model

### Steps To Reproduce
1. Set sqlwarehouse as compute at project level
2. Create a model with insert overwrite as materialization startegy
3. add [compute per model config](https://docs.getdbt.com/reference/resource-configs/databricks-configs#selecting-compute-per-model) to use an interactive compute for the model

### Expected behavior
Any model with compute setup as interactive compute should be able to run insert overwrite

Image

Image
The interactive_compute maps to an all purpose compute 

### System information
```
On model.bted_edw.fct_org_vehicles_stats: /* {"app": "dbt", "dbt_version": "2025.5.21+b7eefcc", "dbt_databricks_version": "1.10.1post2+50c7b9d8e0668268123dc3666bdcd12d105c051e", "databricks_sql_connector_version": "4.0.3", "profile_name": "user", "target_name": "prod", "node_id": "model.bted_edw.fct_org_vehicles_stats"} */
```

dbt-cloud

ericj-db:
Resolves #1032

### Description

Looks like the bug is coming from the fact that the connection credentials always point to the project level compute details. When overriding compute at the model level, we need to check elsewhere to verify whether it is a cluster or SQL warehouse

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Fix bug where python model run starts before all libraries are installed on the cluster ([1028](https://github.com/databricks/dbt-databricks/issues/1028))
+- Fix bug where incorrect insert overwrite exception was being thrown when overriding compute at model level ([1032](https://github.com/databricks/dbt-databricks/issues/1032))
 
 ## dbt-databricks 1.10.2 (May 21, 2025)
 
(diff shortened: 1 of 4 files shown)


benc-db (approved):
LGTM

PR #1050, opened 2025-06-03
Fix to conform to catalog integration config standardization

ericj-db:
Resolves #

### Description

Unit tests were broken by https://github.com/dbt-labs/dbt-adapters/commit/2cfb72caa9cb200e9af6befaaa563f47749ae3cb. Minor fix to align with the standardization in this commit

### Checklist


benc-db (approved):

pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
 dependencies = [
     "databricks-sdk>=0.41, <0.48.0",
     "databricks-sql-connector[pyarrow]>=4.0.0, <5.0.0",
-    "dbt-adapters>=1.14.3, <2.0",
+    "dbt-adapters>=1.16.0, <2.0",
```
ericj-db: @benc-db does this look okay based on [this thread](https://databricks.slack.com/archives/C02JN376A5R/p1748991923990449?thread_ts=1748991716.114139&cid=C02JN376A5R)? Should we call this out in the changelog?

ericj-db: Actually I guess this would fail in CI until 1.16.0 is available in pypi. Should we leave it for now, then bump it before we cut our next release?

benc-db: Yeah...that is annoying.  Maybe file a Github Issue on us that we can't ship 1.10.3 until we get 1.16.0 of dbt-adapters?  Because I think we need these changes to get tests passing again, right?

ericj-db: Yeah our tests will fail in CI without this. Just created https://github.com/databricks/dbt-databricks/issues/1051 and reverted the version pin

benc-db (approved):

PR #1058, opened 2025-06-05
Fix insert_overwrite handling of schema update

### LINKED ISSUE #1057: Incremental insert_overwrite strategy fails on schema change
### Describe the bug
Hi team — Customer is using DBT to incrementally populate aggregated tables on a daily basis, with each table partitioned by a "day" column. Occasionally, they need to delete and reprocess data for past days due to technical issues or logic fixes. In those cases, they use DBT’s insert_overwrite incremental strategy to overwrite the relevant partitions.
One of the great features of DBT is how it handles schema changes: when we remove a column from a DBT model, DBT doesn’t physically drop the column from the table, but simply stops including it in new records. For append-style inserts, this works smoothly—DBT outputs DEFAULT (i.e. NULL) for any missing columns that still exist in the destination table.
However, they recently noticed this behavior doesn't carry over when using the insert_overwrite strategy. In those cases, the DBT-generated SQL omits the step that lists all columns in the destination table and substitutes DEFAULT for missing ones. This leads to errors or unexpected behavior when deprecated columns are still present in the table schema but missing from the DBT model.
```
11:07:39    Database Error in model requests_imps_per_adid_bots (models/gold/requests_imps_per_adid_bots.sql)
  [UNRESOLVED_COLUMN.WITH_SUGGESTION] A column, variable, or function parameter with name `sivt_bots_fire_sting` cannot be resolved. Did you mean one of the following? [`sivt_bots_fireball`, `sivt_bots_parentus`, `sivt_bots_alertus`, `sivt_bots_diceroll`, `sivt_bots_fmscam`]. SQLSTATE: 42703; line 4 pos 951
  compiled code at target/run/ivta/models/gold/requests_imps_per_adid_bots.sql
```

### Steps To Reproduce
Using insert_overwrite, remove an existing column and re-run the model

### Expected behavior
Target table still has the removed column, but filled with default values for the new records

ericj-db:
Resolves #1057

### Description

Update insert_overwrite SQL to do the same thing as append strategy in handling schema updates. Removed column is set to default values for new records

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/materializations/incremental/strategies.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/strategies.sql
@@ -25,15 +25,28 @@
     {{ return(get_insert_overwrite_sql(source, target)) }}
 {% endmacro %}
 
+{% macro get_common_columns(dest_columns, source_columns) %}
+    {%- set common_columns = [] -%}
+    {%- for dest_col in dest_columns -%}
+      {%- if dest_col in source_columns -%}
+        {%- do common_columns.append(dest_col) -%}
+      {%- else -%}
+        {%- do common_columns.append('DEFAULT') -%}
+      {%- endif -%}
+    {%- endfor -%}
+    {% do log("Common columns: " ~ common_columns, info=True) %}
+    {{ return(common_columns) }}
+{% endmacro %}
 
 {% macro get_insert_overwrite_sql(source_relation, target_relation) %}
-
-    {%- set dest_columns = adapter.get_columns_in_relation(target_relation) -%}
-    {%- set dest_cols_csv = dest_columns | map(attribute='quoted') | join(', ') -%}
+    {%- set dest_columns = adapter.get_columns_in_relation(target_relation) | map(attribute='quoted') | list -%}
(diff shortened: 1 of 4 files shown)


dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -25,15 +25,28 @@
...
+    {%- set dest_cols_csv = dest_columns | join(', ') -%}
+    {%- set source_cols_csv = common_columns | join(', ') -%}
     insert overwrite table {{ target_relation }}
     {{ partition_cols(label="partition") }}
-    select {{dest_cols_csv}} from {{ source_relation }}
-
+    select {{source_cols_csv}} from {{ source_relation }}
 {% endmacro %}
 
 {% macro get_replace_where_sql(args_dict) -%}
```
ericj-db, line 46: @benc-db should we also update the replace_where implementation?

benc-db: Yes, but we can do that in a separate PR.

benc-db (approved):

PR #1062, opened 2025-06-09
Add column-level tag support for table, MV, ST, incremental

### LINKED ISSUE #649: Feature Request: Support for column-level databricks tags
### Describe the feature
The new support for Databricks tags on tables & views is really useful. Any chance of supporting column tags too? This is useful for tagging columns that contain particularly sensitive data.

In the same way that `tags` sits under `config` on a model, but `tags` is a top-level attribute on columns:
```yaml
models:
  - name: foo
    description: I'm a teapot
    config:
      tags:
        - hourly
      databricks_tags:
        - foo
    columns:
      - name: id
        tags:
          - bla
      - name: bank_details
        tags:
          - bla
        databricks_tags:
          - don't publish this column on the internet please
      - name: date
        data_type: date
```

Sure, let me know if how I've designed the spec in YAML looks good, and I'll raise a PR for the feature.

ericj-db:
Resolves #649

### Description

Introduce a new `databricks_tags` at the column level with same structure as the model-level one (i.e. key-value map)

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## dbt-databricks 1.10.4 (TBD)
 
+### Features
+
+- Support column tags (with Materialization V2) ([649](https://github.com/databricks/dbt-databricks/issues/649))
+
 ### Fixes
 
 - Fix bug where schema update causes insert_overwrite strategy to fail on subsequent runs ([1057](https://github.com/databricks/dbt-databricks/issues/1057))


dbt/include/databricks/macros/relations/streaming_table/create.sql
```diff
@@ -3,6 +3,9 @@
 {%- endmacro %}
 
 {% macro databricks__get_create_streaming_table_as_sql(relation, sql) -%}
+  {% if column_tags_exist() %}
```
benc-db: This is frustrating...why would this work on MV but not ST ;(.  Have you checked if you can set with ALTER?

benc-db (approved):

dbt/include/databricks/macros/relations/streaming_table/create.sql
ericj-db: Actually my bad, looks like it works. Let me update the PR to support STs

Screenshot 2025-06-12 at 1 42 38 PM

ericj-db: Fixed in https://github.com/databricks/dbt-databricks/pull/1062/commits/bc00f04680cf9fab89a96626311520e0b0195e4a. Thanks for the callout! Would've missed it otherwise :)

benc-db (dismissed):

PR #1064, opened 2025-06-10
Add unit test coverage report to pre-merge

ericj-db:
Resolves #

### Description

Add test coverage statistics to PRs by using https://github.com/py-cov-action/python-coverage-comment-action. No minimum coverage % enforced at this point, just for informational purposes

### Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,7 +94,7 @@ jobs:
       # Only run coverage comment once (not for all python versions)
       - name: Post Coverage Comment
         if: matrix.python-version == '3.12' && github.event_name == 'pull_request'
-        uses: py-cov-action/python-coverage-comment-action@v4
+        uses: py-cov-action/python-coverage-comment-action@v3
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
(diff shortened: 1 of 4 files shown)


benc-db (approved):

.github/workflows/main.yml
```diff
@@ -68,8 +68,17 @@ jobs:
   unit:
     name: unit test / python ${{ matrix.python-version }}
 
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
+    runs-on: linux-ubuntu-latest
```
ericj-db, line 71: @benc-db I had to update the runner to this databricks-hosted one. Otherwise, we hit permission denied because the runner tries to comment on the PR from a non-allowlisted IP address. Lmk if you have any concerns with this

.github/workflows/main.yml
```diff
@@ -68,8 +68,17 @@ jobs:
...
+    timeout-minutes: 15
```
ericj-db, line 72: The new `pytest-cov` dependency seems to be pretty big, so it added a significant amount of time to syncing dependencies. Still much faster than integration test, so not too concerned with this increase

.github/workflows/main.yml
benc-db: No concerns so long as it doesn't prevent community PRs

benc-db (approved):

PR #1079, opened 2025-07-01
Add cloning for the external tables

samgans:
### Description

This PR adds the ability to clone the external tables to the package. Without that, the `dbt clone` command fails as Databricks requires providing the location of the target if we are cloning the external table.

To achieve that, I've introduced the `is_external_table` property of the `DatabricksRelation` that provides the information if the table is external. Then, I've added the new macro named `databricks__create_or_replace_clone_external` that handles the logic of the external cloning and uses the `location_clause` macro to determine where to perform the clone based on the target table configs.

The decision which macro to choose is happening in the `clone` macro itself: if the `is_external_table` attribute of the relation returns True, we call the `_external` macro. If not, we perform the call of the already existing cloning macro.

The second macro was added based on the security considerations: I wouldn't want to break the existing cloning procedures in my release, thus I prefer to have completely separate branch for the new functionality which is more or less independent of already existing code.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 - Support column tags (with Materialization V2) ([649](https://github.com/databricks/dbt-databricks/issues/649))
 - Support column masking (with Materialization V2) ([670](https://github.com/databricks/dbt-databricks/issues/670))
-- Add cloning support for the the external tables ([1079](https://github.com/databricks/dbt-databricks/pull/1079))
+- Add cloning support for the the external tables (thanks @samgans!) ([1079](https://github.com/databricks/dbt-databricks/pull/1079))
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Support column tags (with Materialization V2) ([649](https://github.com/databricks/dbt-databricks/issues/649))
 - Support column masking (with Materialization V2) ([670](https://github.com/databricks/dbt-databricks/issues/670))
+- Add cloning support for the the external tables ([1079](https://github.com/databricks/dbt-databricks/pull/1079))
 
 ### Fixes
 
--- a/dbt/adapters/databricks/relation.py
+++ b/dbt/adapters/databricks/relation.py
@@ -94,6 +94,10 @@ class DatabricksRelation(BaseRelation):
         return self.type == DatabricksRelationType.StreamingTable
 
     @property
+    def is_external_table(self) -> bool:
+        return self.type == DatabricksRelationType.External
+
+    @property
     def is_dlt(self) -> bool:
         return self.is_materialized_view or self.is_streaming_table
 
--- a/dbt/include/databricks/macros/materializations/clone.sql
+++ b/dbt/include/databricks/macros/materializations/clone.sql
@@ -8,6 +8,13 @@
     shallow clone {{ defer_relation }}
 {% endmacro %}
 
+{% macro databricks__create_or_replace_clone_external(this_relation, defer_relation) %}
+    create or replace
+    table {{ this_relation }}
+    shallow clone {{ defer_relation }}
+    {{ location_clause(this_relation) }}
+{% endmacro %}
+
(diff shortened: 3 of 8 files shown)


mi-volodin:
Also would add a minor question - on which version (branch) should we rebase the change?

benc-db:
> Also would add a minor question - on which version (branch) should we rebase the change? If the change is feasible and would be welcomed, of course.

Branching from main is correct, and looking at the PR, this seems like something we are happy to take.  My understanding is that basically if a table is asked to be cloned, we have been limited previously for external tables because the syntax for the clone requires the location; this PR detects if the source is external and modifies the syntax to accommodate it, right?

> Should I do something with them? Do you have any suggestions on how can I ensure that this change is robust and works?

The problem with external locations is that a.) Databricks doesn't clean up the files because they are treated as unmanaged (i.e. dropping a table does not delete the files backing the table), and b.) the process of clean up outside of Databricks might be a function of what cloud the files are written in.  This makes it hard to write portable tests that don't leave artifacts lying around.

If possible, can you look at the macro unit tests and write a unit test that validates that the macro output looks exactly like what you're expecting?  You could write an E2E test as well, but it will almost certainly leave files lying around.

samgans:
Understood, will write the macro test. Thanks for the input!

samgans:
Hey @benc-db, I've created the tests for this functionality, but to be able to do this I needed to introduce some more changes:

1. The `clone.sql` file now only the helper function `databricks__can_clone_table` and the definition of the materialization.
2. The clone strategies are moved to separate `strategies.sql` file.
3. Everything related to the clone is moved to separate `clone/` directory inside `materializations/`.

I needed to do this to get rid of the issue that pops ups when trying to test the macro defined in the same file where `materialization` tag present: `tag materialization doesn't exist`.

I've taken the approach which I've found with the `incremental` materialization and replicated this. Hope this is acceptable.

Additionally, I've introduced some small changes to the unit tests:

1. Moved `unity_relation` function to the `utils.py` as it is currently reused by two unit tests.
2. Adapted `test_table_macros` to the usage of the common `unity_relation`.

And to the clone macros themselves:

1. It seems to be expected to pass the CatalogRelation to the `location_clause`, which is now generated in the `databricks__create_or_replace_clone_external`.
2. We call `render()` function of the relations to render their names instead of their implicit passing to be able to test this (I've replicated this from the `insert` macros, so not sure if this is something which is expected to be done).

And finally, I've added the `test_clone_macros.py` code which replicates the logic of `test_table_macros`, but tests the cloning itself.

Open for the feedback.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Support column tags (with Materialization V2) ([649](https://github.com/databricks/dbt-databricks/issues/649))
 - Support column masking (with Materialization V2) ([670](https://github.com/databricks/dbt-databricks/issues/670))
+- Add cloning support for the the external tables ([1079](https://github.com/databricks/dbt-databricks/pull/1079))
```
mi-volodin:
```suggestion
- Add cloning support for the the external tables ([1079](https://github.com/databricks/dbt-databricks/pull/1079))(thanks @samgans)
```

samgans: Done

samgans:
Hm, I see that functional tests related to Streaming Tables are failing. Do you think it is related to any of my changes? I can't really assess, it seems like no, but I'm not too deep into the code base.

samgans:
Hm I'm wondering if this is related to this change somehow:

https://github.com/databricks/dbt-databricks/pull/1082/files

I'm not sure if I need to merge the main branch into mine.

samgans:
I've rebased my branch to be on top of the current main.

mi-volodin:
@benc-db @ericj-db  could you please re-trigger the tests?

ericj-db (approved):

samgans:
Hey @benc-db @ericj-db FYI we have found some bugs in the implementation from this PR and are in the process of fixing them.

We will open the follow-up PR with some more changes in the next two days.

Sorry for that, hope it doesn't break the release cycle.

PR #1080, opened 2025-07-01
Fix inefficient query when getting column schema for MV/STs

### LINKED ISSUE #1074: DBT Creating Materialized Views with Redundant Query Executions
Describe the bug
When creating materialized views in Databricks using dbt, the process is executing the same query twice, which significantly increases processing time. The first query appears to be a validation/schema check, followed by the actual materialized view creation, but this behavior has suddenly started and is doubling our processing time.

Steps To Reproduce
Create a standard dbt model that uses materialized view materialization
Run a command like dbt run --select ft_table
Observe that dbt runs two queries:
First query: SELECT * FROM source_table WHERE conditions
Second query: CREATE MATERIALIZED VIEW target_table (...columns...) AS SELECT * FROM source_table WHERE conditions
Sample model code
-- models/marts/xx/ft_table.sql
SELECT
  *
FROM {{ source('source', 'ft_table') }}

Model configuration
# __models.yml

  materialized: materialized_view

  # other settings...

Expected behavior
dbt should create the materialized view with a single query execution. Previous versions of dbt didn't require this double execution.

it is the dbt version 

Core:
  - installed: 1.10.2

Plugins:

Now I have to roll back to 

Core:
  - installed: 1.10.2

Plugins:
  - databricks: 1.10.1 - Update available!

Impact
This issue is doubling our processing time for materialized views, which has a significant impact on our development and production refresh cycles. For large datasets, this means a build that previously took 7 minutes now takes 15 minutes or more.

ericj-db:
Resolves #1074

### Description

The implementation of `get_column_schema_from_query` (used for MV/STs) runs the entire user query which is inefficient. Quick fix to do a `LIMIT 0` for faster execution

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -406,7 +406,9 @@ class DatabricksAdapter(SparkAdapter):
     @available.parse(lambda *a, **k: [])
     def get_column_schema_from_query(self, sql: str) -> list[DatabricksColumn]:
         """Get a list of the Columns with names and data types from the given sql."""
-        _, cursor = self.connections.add_select_query(sql)
+        # Optimize by wrapping the query with LIMIT 0 to get schema without executing the full query
+        optimized_sql = f"{sql} LIMIT 0"
+        _, cursor = self.connections.add_select_query(optimized_sql)
         try:
             columns: list[DatabricksColumn] = [
                 self.Column.create(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## dbt-databricks 1.10.5 (TBD)
 
+### Fixes
+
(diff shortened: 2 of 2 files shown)


benc-db (dismissed):

dbt/adapters/databricks/impl.py
```diff
@@ -406,7 +406,9 @@ def _get_hive_relations(
     @available.parse(lambda *a, **k: [])
     def get_column_schema_from_query(self, sql: str) -> list[DatabricksColumn]:
         """Get a list of the Columns with names and data types from the given sql."""
-        _, cursor = self.connections.add_select_query(sql)
+        # Optimize by wrapping the query with LIMIT 0 to get schema without executing the full query
+        optimized_sql = f"{sql} LIMIT 0"
```
benc-db: I think we get called upstream with a SQL that already has a limit 0 in it.  We should use that upstream macro, which has the same name.

benc-db (approved):

PR #1084, opened 2025-07-03
Fix TypedConstraint equality check

### LINKED ISSUE #1081: Cannot update models with primary & foreign key constraints since version 1.10.2 with MaterializationV2
### Describe the bug

Primary key constraints fail to be dropped when there are dependent foreign key constraints.

```
Database Error in model .....
  Cannot delete constraint `..._pk` with non empty child constraints. Set the argument CASCADE to true to delete both this constraint and all child constraints.
  compiled code at target/run/.....
```

### Steps To Reproduce
Create two models, give one a primary key, and the other a foreign key, make it depend on the model with the primary key. Now run dbt and watch it fail to drop the primary key constraint.

### Expected behavior
This SQL should have `CASCADE` at the end when a primary key constraint is dropped, so that foreign key constraints depending on it are dropped as well.

https://github.com/databricks/dbt-databricks/blob/df3d0f6b6ecd7976d365878742e74403afb1ca68/dbt/include/databricks/macros/relations/components/constraints.sql#L133-L135

### System information
```
Core:
  - installed: 1.10.2

Plugins:
```

`OS: WSL`

`Python 3.12.10`

### Additional context

With MaterializationV2 enabled, the last working configuration is:

```
dbt-core: 1.10.2
dbt-databricks: 1.10.1
```

If I disable MaterializationV2 I can use the latest version of dbt-databricks (1.10.4).

ericj-db:
Resolves #1081

### Description

It looks like the root cause behind #1081 is because of how we determine equality between TypedConstraint objects. During incremental runs, we rely on the equality check to determine which constraints to drop or apply https://github.com/databricks/dbt-databricks/blob/6981678333b6f1b715b272041b166576df925e37/dbt/adapters/databricks/relation_configs/constraints.py#L30-L32

The current default equality check is considering fields like `warn_unenforced`. This does not work because this is a user supplied config that has no meaning when we are creating a `TypedConstraint` from existing Databricks relations. The Fix is to override the `__eq__` to match the `__hash__` which already ignores such fields

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Fix inefficient query when getting column schema for MV/STs ([1074](https://github.com/databricks/dbt-databricks/issues/1074))
+- Fix bug causing false positives in diffing constraints between existing relation and model config for incremental runs ([1081](https://github.com/databricks/dbt-databricks/issues/1081))
 
 ## dbt-databricks 1.10.4 (June 24, 2025)
 
--- a/dbt/adapters/databricks/constraints.py
+++ b/dbt/adapters/databricks/constraints.py
@@ -76,6 +76,14 @@ class TypedConstraint(ModelLevelConstraint, ABC):
         )
         return hash(fields)
 
+    def __eq__(self, other: Any) -> bool:
+        """Override equality to only compare fields used in hash calculation.
+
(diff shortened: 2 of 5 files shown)


benc-db (approved):

Korijn:
I installed dbt-databricks from this branch and tested locally; the issue seems to be fixed. 🎉

PR #1087, opened 2025-07-07
Execute no-op statement in apply_tags when no tags to set

### LINKED ISSUE #1077: alter view seems not working as expected
### Describe the bug
I enabled the view_update_via_alter flag in dbt_project.yml but it doesn't seem to be applied when recreating a view with dbt run/build command

### Steps To Reproduce

dbt_project.yml has the following attributes:
```
...

flags:
    use_materialization_v2: true
    view_update_via_alter: true

models:
  ssda:
    +post_hook: 
      - "{{ apply_primary_keys() }}"
      - "{{ apply_tags_db() }}"
    +file_format: delta 
    +table_properties:
      delta.enabled: true
    +materialized: view # By default we materialize everything as a view


...
```

I also tested view_update_via_alter in the config of dbt_project.yml but this also didn't help:

```
...

flags:
    use_materialization_v2: true

models:
  ssda:
    +post_hook: 
      - "{{ apply_primary_keys() }}"
      - "{{ apply_tags_db() }}"
    +file_format: delta 
    +table_properties:
      delta.enabled: true
    +materialized: view # By default we materialize everything as a view
    +config:
      +view_update_via_alter: true

...
```

I looked at the logs whether the view was altered after it was already created but this seems not the case.

I do see the following query in the logs to check if the view exists:
```
SELECT
  table_name,
  if(table_type IN ('EXTERNAL', 'MANAGED', 'MANAGED_SHALLOW_CLONE', 'EXTERNAL_SHALLOW_CLONE'), 'table', lower(table_type)) AS table_type,
  lower(data_source_format) AS file_format,
  table_owner
FROM `system`.`information_schema`.`tables`
WHERE table_catalog = 'xxx' 
  AND table_schema = 'xxx'
```

Resulting in the following log (and after the log a create or replace view statement).
```

[0m11:44:56.802605 [debug] [Thread-1 (]: MATERIALIZING VIEW
[0m11:44:56.836092 [debug] [Thread-1 (]: Applying REPLACE to: `xxx`.`xxx`.`xxx`
```

### Expected behavior

I expected to see an alter view statement after dbt build/run command when the view already exists. We need this because otherwise permissions on our prod environment get overwritten.

I had an alternative macro that worked and expected this new flag/config to work kind of the same. Below the macro I wrote myself:

``` 
{% materialization view, adapter="databricks" -%}
    {#  Deze macro overschrijft default view materialisatie en doet een alter view als de view al bestaat om te voorkomen dat rechten vervallen #}
    
    {%- set target_relation = api.Relation.create(
        identifier=this.identifier, schema=this.schema, database=this.database,
        type='view') -%}
    {%- set tags = config.get('databricks_tags') -%}

    {{ run_hooks(pre_hooks) }}

    {% call statement('main') -%}
        {%- set view_exists_query -%}
            select 1
            from {{this.database}}.information_schema.views
            where table_schema = '{{ this.schema }}'
                and table_name = '{{ this.identifier }}'
        {%- endset -%}
        {%- set view_exists_result  = none %}
        {%- set view_exists_result = run_query(view_exists_query) -%}

        {%- if view_exists_result and view_exists_result.rows|length > 0 -%}
            alter view {{ this }} as {{ sql }}
        {%- else -%}
            {{ get_create_view_as_sql(target_relation, sql) }}
        {%- endif -%}
    {%- endcall %}

    {{ run_hooks(post_hooks) }}

    {%- do apply_tags(this, tags) -%}

    {{ return({'relations': [target_relation]}) }}
 
{% endmaterialization -%}
```

### System information
```
Core:
  - installed: 1.10.2

Plugins:
```

Python 3.12.0

### Additional context

When I run the query below in databricks sql editor, I see the view exists (with an unknown file format though?)

Image

```
SELECT
  table_name,
  if(table_type IN ('EXTERNAL', 'MANAGED', 'MANAGED_SHALLOW_CLONE', 'EXTERNAL_SHALLOW_CLONE'), 'table', lower(table_type)) AS table_type,
  lower(data_source_format) AS file_format,
  table_owner
FROM `system`.`information_schema`.`tables`
WHERE table_catalog = 'xxx' 
  AND table_schema = 'xxx'
```

Do I miss any additional flag to be set to true or any other settings?

ericj-db:
Resolves #1077

### Description

`databricks_tags` is an "add only" config meaning that the Databricks adapter will never issue a statement to drop it once it is set (reference: https://docs.getdbt.com/reference/resource-configs/databricks-configs#configuring-tables)

There is currently a bug related to this which can be reproduced when:
- Existing view has a tag
- Model config does not have a matching tag
- No other diff exists between the existing view and the model config

This results in a no-op where no statements are executed, so `main is not being called during running model` error is thrown. This PR fixes this behavior by executing a no-op statement with name `main` for the edge case

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/relations/view/alter.sql
+++ b/dbt/include/databricks/macros/relations/view/alter.sql
@@ -8,7 +8,7 @@
   {% set tblproperties = changes.get("tblproperties") %}
   {% set query = changes.get("query") %}
   {% set column_comments = changes.get("column_comments") %}
-  {% if tags and tags.set_tags %}
+  {% if tags %}
     {{ apply_tags(target_relation, tags.set_tags) }}
   {% endif %}
   {% if tblproperties %}


benc-db (dismissed):
Are we confident that tags is the only path that can end not having main called?

ericj-db:
> Are we confident that tags is the only path that can end not having main called?

For views, I drilled down to all of the macros I could find and it seemed like everything else would either issue a set or unset statement. I think tags is a unique case because of the fact that we never execute unset statements

PR #1093, opened 2025-07-08
Handle complex data types that are truncated by DESCRIBE TABLE

### LINKED ISSUE #1083: Complex data types are truncated when using materialization_v2, causing errors
### Describe the bug
When using materialization_v2, tables with large nested fields cause model refreshes to fail. Specifically, the `create or replace table` statement tries to create a field in the format `column_name struct < field1::type, field2::type, ... field25::type...75 more fields >`. This `75 more fields` part is due to truncation when running the `DESCRIBE` statement to get field types, and setting `use_info_schema_for_columns` doesn't resolve this issue.

I assume that the reason why `use_info_schema_for_columns` doesn't work is because it's running a `DESCRIBE` statement on a temporary table, which isn't in the information schema. However, the truncation is causing model failures. I'm not sure if this is viable, but it seems like `DESCRIBE EXTENDED ... AS JSON` doesn't truncate the data types, so that may be one potential solution?

### Steps To Reproduce
1. Create a model with a large struct, at least 30 fields or so.
2. Set `use_materialization_v2` to true
3. Run model

### Expected behavior
I would expect the model to run successfully and not error, specifically by correctly compiling with untruncated data types.

ericj-db:
Resolves #1083

### Description

Instead of using `DESCRIBE TABLE`, use `DESCRIBE TABLE EXTENDED <relation> AS JSON` to get full type information for columns. I wish there was a better/more sustainable solution than us having to write the JSON -> DDL translation, but looks like nothing better exists at the moment. For example I thought about using `CREATE TABLE .. USING JSON`, but looks like it can only accept a path to a file, and not the literal value which would be too complicated

### Checklist

### DIFF
--- a/dbt/adapters/databricks/behaviors/columns.py
+++ b/dbt/adapters/databricks/behaviors/columns.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+from typing import Any
 
 from dbt_common.utils.dict import AttrDict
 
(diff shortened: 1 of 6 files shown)


dbt/adapters/databricks/behaviors/columns.py
```diff
@@ -33,24 +34,96 @@ class GetColumnsByDescribe(GetColumnsBehavior):
...
-            )
+        for col_info in data.get("columns", []):
+            col_name = col_info.get("name")
+            col_type = cls._parse_type(col_info.get("type"))
+            comment = col_info.get("comment")
+            columns.append(DatabricksColumn(column=col_name, dtype=col_type, comment=comment))
 
         return columns
 
+    @classmethod
```
benc-db: I think somewhere we will need logic like this, but I question if this is where it should go.  Let me do some research.

benc-db:
I suspect they didn't implement the AS JSON for HMS

PR #1095, opened 2025-07-10
[FIX]: Fix issue with cloning of external tables

samgans:
### Description

This PR resolves the bugs related to the external cloning introduced in #1079 (sorry for introducing them :smiling_face_with_tear:).

#### What are these bugs?

The cloning of the external tables still doesn't work, because the `{% if other_existing_relation.is_external %}` branch isn't followed due to multiple issues:

1.  The property of the relation is named `is_external_table` instead of `is_external`. This is the typo that was introduced by me.
2. The property `is_external_table` returns `False` for every external table due to the fact that the `get_uc_tables_sql` macro casts all the tables that are of  `'EXTERNAL', 'MANAGED', 'MANAGED_SHALLOW_CLONE', 'EXTERNAL_SHALLOW_CLONE'` types to the `TABLE` type thus making any external-related `DatabricksRelationType` not available as `type` attribute of the `DatabricksRelation`.

In addition to that, the `databricks__create_or_replace_clone_external` macro was not available for the usage due to the dispatch syntax used in its declaration. Instead, we should have declared it as `create_or_replace_clone_external`.

##### Bug 1
To fix the bug 1, I've simply fixed the call of the function.

[File 1.
](https://github.com/databricks/dbt-databricks/blob/523963507f7574736f3f1db9aeb3d9f10323fa28/dbt/include/databricks/macros/materializations/clone/clone.sql)

##### Bug 2
To fix the bug 2, I've changed the `get_uc_tables_sql` macro to also return an attribute named `databricks_table_type` which additionally classifies the databricks table. To make this work, I've created another enumeration `DatabricksTableType` which means the specific databricks table type and made this as a `databricks_table_type` attribute of the `DatabricksRelation`. The Relation uses this attribute to determine if it is an external table or not.

[File 1.
](https://github.com/databricks/dbt-databricks/blob/523963507f7574736f3f1db9aeb3d9f10323fa28/dbt/include/databricks/macros/adapters/metadata.sql)
Adding `databricks_table_type` attribute instead of simply updating the `type` attribute was done to keep the backwards compatibility of the attribute and not break potential assumptions about this attribute.

[File 2.
](https://github.com/databricks/dbt-databricks/blob/523963507f7574736f3f1db9aeb3d9f10323fa28/dbt/adapters/databricks/relation.py)
Introducing the new attribute returned from the macro has also required me to fix the type definitions of the return rows, thus I've introduced the new named tuple  `DatabricksRelationInfo` so that the return values of the query that describes the relation are better documented, at the same time keeping the backwards-compatible type.

[File 3.
](https://github.com/databricks/dbt-databricks/blob/523963507f7574736f3f1db9aeb3d9f10323fa28/dbt/adapters/databricks/impl.py)

##### Bug 3
The dispatch of the external macro was fixed by removing the dispatch-based syntax in its definition.

[File 1.
](https://github.com/databricks/dbt-databricks/blob/523963507f7574736f3f1db9aeb3d9f10323fa28/dbt/include/databricks/macros/materializations/clone/strategies.sql)

#### Testing

We've found a way to test these changes in our workspace, so we were able to run the clone operation successfully using the external tables:

image

image

 
### Checklist

### DIFF
--- a/tests/unit/test_adapter.py
+++ b/tests/unit/test_adapter.py
@@ -382,9 +382,7 @@ class TestDatabricksAdapter(DatabricksAdapterBase):
     @patch("dbt.adapters.databricks.api_client.DatabricksApiClient.create")
     def test_list_relations_without_caching__hive_relation(self, _):
         with patch.object(DatabricksAdapter, "get_relations_without_caching") as mocked:
-            mocked.return_value = [
-                DatabricksRelationInfo("name", "table", None, None, None)
-            ]
+            mocked.return_value = [DatabricksRelationInfo("name", "table", None, None, None)]
             adapter = DatabricksAdapter(Mock(flags={}), get_context("spawn"))
             relations = adapter.list_relations("database", "schema")
             assert len(relations) == 1


samgans:
image

samgans:
Code quality run:

image

ericj-db (approved):
Thanks for the fix!

PR #1132, opened 2025-08-04
Fix no-op logic when no databricks_tag changes to apply

### LINKED ISSUE #1122: dbt build skipping materializations
### Describe the bug
dbt build skips materializing all models with dbt-databricks 1.10.6. Tests did run and pass

### Steps To Reproduce
We use databricks and had a requirement file like this:
```
dbt-databricks>=1.10,<1.11
dbt-core>=1.10,<1.11

```
Suddenly today, our tables weren't materialized (they were skipped), even though nothing had changed in the repository. The only difference I noticed was an update from dbt-databricks 1.10.4 to 1.10.6. Downgrading resolved the issue, but it seems like unexpected behavior. A quick glance at the changelog didn’t provide any explanation either.

We also observed that running dbt-databricks 1.10.6 locally in VS Code led to the same behavior.

### System information
We had the issue with:
Image
**python 3.12 in vs code**

Last succesful runs with 

Image

ericj-db:
Resolves #1122

### Description

We previously added `execute_no_op(relation)` to `apply_tags` to prevent `main is not being called during running model` errors for views when there are no config changes to apply. This was the wrong layer to address this problem. The real fix is to make sure the tag relation config returns `None` as the diff when there are no net-new tags to apply

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,6 @@
 ### Under the Hood
 - Simplify connection management to align with base adapter. Connections are no longer cached per-thread
 
-### Under the Hood
-- Simplify connection management to align with base adapter. Connections are no longer cached per-thread
-
 ## dbt-databricks 1.10.7 (July 31, 2025)
 
 ### Fixes


benc-db (approved):

ericj-db:
Failed integration tests seem like a transient issue with python jobs. I re-ran locally on the same branch and they passed. Will merge, but keep an eye on the main branch tests

PR #1134, opened 2025-08-05
Add column tag support for views

ericj-db:
Resolves #

### Description

Previously, we did not add column tag support for views because `ALTER VIEW` does not support it. However, it turns out that `ALTER TABLE` works on views for setting column tags. This PR adds the missing column tag support for views and updates column tag functional tests to add coverage for updated column tag configs between runs

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/relations/components/column_tags.sql
+++ b/dbt/include/databricks/macros/relations/components/column_tags.sql
@@ -34,7 +34,12 @@
 {%- endmacro -%}
 
 {% macro alter_set_column_tags(relation, column, tags) -%}
-  ALTER {{ relation.type | replace('_', ' ') }} {{ relation.render() }}
+  {# ALTER VIEW does not support setting column tags, but ALTER TABLE works for views #}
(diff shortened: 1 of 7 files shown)


dbt/include/databricks/macros/relations/components/column_tags.sql
```diff
@@ -34,7 +34,12 @@
 {%- endmacro -%}
 
 {% macro alter_set_column_tags(relation, column, tags) -%}
-  ALTER {{ relation.type | replace('_', ' ') }} {{ relation.render() }}
+  {# ALTER VIEW does not support setting column tags, but ALTER TABLE works for views #}
+  {%- if relation.type == 'view' -%}
+    ALTER TABLE {{ relation.render() }}
```
benc-db, line 39: this is bizarre.

ni-mi:
@ericj-db Thanks for this one - we were just about to ask for it 😄 
Is there a release with this feature planed soon so we can use it?

Thanks!

PR #1127, opened 2025-08-01
Draft: Support insert_overwrite incremental strategy for SQL warehouses

ericj-db:
Resolves #

### Checklist

- [ ] I have run this code in development and it appears to resolve the stated issue
- [ ] This PR includes tests, or tests are not required/relevant for this PR
- [ ] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section.

PR #1135, opened 2025-08-06
feat: Use REPLACE ON for insert overwrite strategy

ericj-db:
Resolves #

### Description

Previous change in https://github.com/databricks/dbt-databricks/pull/1127 is has 2 bugs
- Uses `REPLACE USING` which is not null tolerant, so we need to use `REPLACE ON` with `<=>` instead
- Supports `liquid_clustered_by` columns (handled the same way as `partition_by`

This PR updates the macro to use `REPLACE ON` or fallback to the `INSERT OVERWRITE TABLE` based on the following logic:
1. Clusters with DBR version 17.1+ always use `REPLACE ON`
2. All other clusters always use `INSERT OVERWRITE TABLE`
3. SQL warehouses with behavior flag `use_replace_on_for_insert_overwrite = True` (default) will always use `REPLACE ON`
4. SQL warehouse with behavior flag `use_replace_on_for_insert_overwrite = False` will always use `INSERT OVERWRITE TABLE`

As discussed with @benc-db , we need to support option 4 until we add an explicit `insert + delete` strategy for incremental

FYI: I updated the all purpose DBR versions to 17.1 for our CI environment, so the functional tests are all running the `REPLACE ON` logic for clusters, and `INSERT OVERWRITE TABLE` for SQL warehouses

TODO: 
- Update dbt docs to reflect this change. I can batch with my currently open doc PR: https://github.com/dbt-labs/docs.getdbt.com/pull/7741

### Checklist

### DIFF
--- a/tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
+++ b/tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
@@ -42,6 +42,9 @@ class TestInsertOverwriteMacros(MacroTestBase):
         )
         context["adapter"].behavior = mock_behavior
 
+        # Mock exceptions.warn to return empty string to avoid polluting SQL output
+        context["exceptions"].warn.return_value = ""
+
     @pytest.mark.parametrize(
         "dbr_version_comparison,expected_sql",
         [


dbt/adapters/databricks/connections.py
```diff
@@ -402,7 +407,8 @@ def connect() -> DatabricksHandle:
             try:
                 # TODO: what is the error when a user specifies a catalog they don't have access to
                 conn = DatabricksHandle.from_connection_args(
-                    conn_args, creds.cluster_id is not None
```
ericj-db: @benc-db this is a bug not directly related to this PR, but caught by a failing test. When we override a SQL warehouse compute with all purpose cluster at the model level, `compare_dbr_version` was always returning 1. This is because the connection was falsely getting initialized as not a cluster even though it is

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -39,28 +39,50 @@
...
+    
+    {%- if replace_columns -%}
+        {%- set replace_conditions = [] -%}
+        {%- for col in replace_columns -%}
+            {%- do replace_conditions.append('t.' ~ col ~ ' <=> s.' ~ col) -%}
+        {%- endfor -%}
+        {%- set replace_conditions_csv = replace_conditions | join(' AND ') -%}
+        insert into table {{ target_relation }} AS t
+        replace on ({{ replace_conditions_csv }})
+        (select {{ source_cols_csv }} from {{ source_relation }}) AS s
```
ericj-db: @longvu-db this is the part that is relevant to `REPLACE ON`

tests/functional/adapter/incremental/test_incremental_strategies.py
```diff
@@ -174,6 +200,38 @@ def test_incremental(self, project):
         util.check_relations_equal(project.adapter, ["overwrite_model", "upsert_expected"])
 
 
+# Insert overwrite in SQL warehouse is expected to behave like a table materialization
+# We support this as a short term hack for customers who want the side effect of reusing
+# the same table on subsequent runs
+# TODO: Remove this once 17.1 is current version for SQL warehouses
+@pytest.mark.skip_profile("databricks_uc_cluster", "databricks_cluster")
```
benc-db: I'm realizing that this means we shouldn't release this until we have an alternative for these users to move to.  We currently do not support delete+insert (one of the standard dbt incremental strategies).  If we can support that to offer a similar behavior, then those users would be able to migrate.  I'm thinking we either need to hide behind a behavior flag, or wait until 1.11.0, and ship with delete+insert as an offramp.

ericj-db: @benc-db I've added the behavior flag and updated the PR description to reflect this. When 17.1 becomes the current DBSQL version, customers should be able to opt-in by ramping up the flag. For SQL warehouses by default, it will fallback to the old macro

dbt/adapters/databricks/connections.py
```diff
@@ -128,15 +128,20 @@ def api_client(self) -> DatabricksApiClient:
             )
         return self._api_client
 
-    def is_cluster(self) -> bool:
-        conn = self.get_thread_connection()
+    @staticmethod
+    def _is_cluster_http_path(cluster_id: Optional[str], http_path: str) -> bool:
```
benc-db: Optional arguments should follow required arguments.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,54 @@
...
+    {#-- Use partition columns if defined (takes precedence) --#}
+    {%- if partition_by -%}
+        {%- if partition_by is string -%}
+            {%- set partition_by = [partition_by] -%}
+        {%- endif -%}
+        {%- for partition_col in partition_by -%}
+            {%- do replace_columns.append(partition_col) -%}
+        {%- endfor -%}
+    {#-- Otherwise use liquid clustering columns if defined --#}
+    {%- elif liquid_clustered_by -%}
```
benc-db: Is this the same precedence we use elsewhere?

dbt/adapters/databricks/connections.py
benc-db: If we want this to be a static method, I think maybe move to utils and use a public name.  It's not really a capability that the ConnectionManager should be responsible for knowing the implementation of.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
ericj-db: I manually verified that we don't have a precedence/explicit guard for both `partition_by` and `liquid_clustered_by` coexisting. Databricks will just throw `[SPECIFY_CLUSTER_BY_WITH_PARTITIONED_BY_IS_NOT_ALLOWED] Cannot specify both CLUSTER BY and PARTITIONED BY. SQLSTATE: 42908`

Given that such a config wouldn't be able to reach this point, I removed the special handling in 35d792a581523d6f84fe8b82c0a5f86078da1473

dbt/adapters/databricks/connections.py
ericj-db: Fixed in ac344eb393fc645e08005455815a293d887d9c44

ericj-db:
@benc-db the failed integration tests are transient. Could you take another pass at this PR?

benc-db (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## dbt-databricks 1.10.10 (TBD)
 
+### Features
+- Support insert_overwrite incremental strategy for SQL warehouses using `REPLACE ON` for null tolerance, gated behind behavior flag ([1025](https://github.com/databricks/dbt-databricks/issues/1025))
```
longvu-db:
```suggestion
- Support `insert_overwrite` incremental strategy for SQL warehouses using `REPLACE ON` for null tolerance, gated behind behavior flag ([1025](https://github.com/databricks/dbt-databricks/issues/1025))
```

longvu-db: > using `REPLACE ON` for null tolerance

"REPLACE ON for null tolerance" is an internal implementation detail, should we mention explicitly here? Especially the "null tolerance" part, which can confuse users, as in why is this detail regarding NULL gets mentioned now instead of previously with the classic SQL DPO `INSERT OVERWRITE` with a knob?

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -191,7 +191,7 @@
       set spark.sql.sources.partitionOverwriteMode = {{ value }}
     {%- endcall -%}
   {% else %}
-    {{ exceptions.warn("INSERT OVERWRITE is only properly supported on all-purpose clusters.  On SQL Warehouses, this strategy would be equivalent to using the table materialization.") }}
+    {{ exceptions.warn("INSERT OVERWRITE is supported on SQL warehouses with DBR 17.1+. On older DBR versions, this strategy would be equivalent to using the table materialization.") }}
```
longvu-db: Does "table materialization" here mean you are overwriting everything instead of selectively overwriting the matching partitions?

https://docs.getdbt.com/reference/resource-configs/databricks-configs

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -191,7 +191,7 @@
       set spark.sql.sources.partitionOverwriteMode = {{ value }}
     {%- endcall -%}
   {% else %}
-    {{ exceptions.warn("INSERT OVERWRITE is only properly supported on all-purpose clusters.  On SQL Warehouses, this strategy would be equivalent to using the table materialization.") }}
```
longvu-db, line 194: P/s: Feels a bit weird that INSERT OVERWRITE here is in capital which suggests that it's SQL, shouldn't it be insert_overwrite instead?

dbt/adapters/databricks/utils.py
```diff
@@ -88,3 +88,12 @@ def handle_exceptions_as_warning(op: Callable[[], None], log_gen: ExceptionToStr
         op()
     except Exception as e:
         logger.warning(log_gen(e))
+
+
+def is_cluster_http_path(http_path: str, cluster_id: Optional[str]) -> bool:
```
longvu-db, line 93:
```suggestion
def is_cluster_http_path(http_path: str, cluster_id_opt: Optional[str]) -> bool:
```

Nit: I prefer when a variable is Optional, that is has the suffix "opt", but this might not follow python coding style guide.

dbt/adapters/databricks/impl.py
```diff
@@ -122,6 +122,15 @@
     ),
 )  # type: ignore[typeddict-item]
 
+USE_REPLACE_ON_FOR_INSERT_OVERWRITE = BehaviorFlag(
+    name="use_replace_on_for_insert_overwrite",
+    default=False,
+    description=(
+        "Use the new INSERT REPLACE ON syntax for insert_overwrite incremental strategy."
+        "  When disabled, falls back to traditional INSERT OVERWRITE syntax."
```
longvu-db:
```suggestion
        "  When disabled, falls back to legacy INSERT OVERWRITE syntax."
```

Should we call it legacy now? https://docs.databricks.com/aws/en/delta/selective-overwrite?language=SQL#dynamic-partition-overwrites-with-partitionoverwritemode-legacy

dbt/adapters/databricks/connections.py
longvu-db: Should we put this in a separate PR?

tests/functional/adapter/incremental/test_incremental_strategies.py
```diff
@@ -120,6 +120,29 @@ def test_incremental(self, project):
         util.check_relations_equal(project.adapter, ["overwrite_model", "upsert_expected"])
 
 
+# TODO: Enable for SQL warehouses once 17.1+ is current version
```
longvu-db: I think we can just enable this test now by enabling the necessary flags right, using a python version of `withSQLConf`? I think it's a good idea to have the test run, before the release version?

tests/functional/adapter/incremental/test_incremental_strategies.py
```diff
@@ -181,6 +204,7 @@ def test_incremental(self, project):
 # Insert overwrite in SQL warehouse is expected to behave like a table materialization
 # We support this as a short term hack for customers who want the side effect of reusing
 # the same table on subsequent runs
+# TODO: Remove this once 17.1 is current version for SQL warehouses
```
longvu-db: We can keep this test with the behavior flag disable to ensure that the old behavior is correct?

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,296 @@
...
+        # Mock is_cluster to return True by default (cluster environment)
+        context["adapter"].is_cluster.return_value = True
+
+        # Mock behavior flags
+        mock_behavior = Mock()
+        mock_behavior.use_insert_replace_on = True  # Default to True for behavior flag
+        context["adapter"].behavior = mock_behavior
+
+    def test_get_insert_overwrite_sql__legacy_dbr_version(self, template, context, config):
+        """Test that DBR < 17.1 uses traditional INSERT OVERWRITE syntax"""
```
longvu-db:
```suggestion
        """Test that DBR < 17.1 uses the legacy DPO INSERT OVERWRITE syntax"""
```

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,55 @@
...
+{% endmacro %}
+
+{% macro get_insert_replace_on_sql(source_relation, target_relation, source_cols_csv) %}
+    {%- set partition_by = config.get('partition_by') -%}
+    {%- set liquid_clustered_by = config.get('liquid_clustered_by') -%}
+    {%- set replace_columns = [] -%}
+    
+    {#-- If both partition_by and liquid_clustered_by are defined, it will fail before this point with a SPECIFY_CLUSTER_BY_WITH_PARTITIONED_BY_IS_NOT_ALLOWED error from Databricks --#}
+    {%- if partition_by -%}
+        {%- if partition_by is string -%}
```
longvu-db: Wonder if we have anyway to refactor this `partition_by` and `clustered_by` with a macro?

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,296 @@
...
+        # Run the macro
+        result = self.run_macro_raw(
+            template,
+            "get_insert_overwrite_sql",
+            source_relation,
+            target_relation,
+        )
+
+        # Verify it uses regular INSERT OVERWRITE syntax
+        expected_sql = """insert overwrite table target_table select a, b from source_table"""
```
longvu-db: Could we put the expected_sql in the same formats as other tests? I.e. the select on a separate line, for readability

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,296 @@
...
+    def test_get_insert_overwrite_sql__behavior_flag_disabled_cluster(
+        self, template, context, config
+    ):
+        """Test that behavior flag disabled on cluster still uses REPLACE ON"""
+        # Positive return value means DBR >= 17.1
+        context["adapter"].compare_dbr_version.return_value = 1
+        # Cluster environment (default from setup)
+        context["adapter"].is_cluster.return_value = True
+        # Disable the behavior flag
+        context["adapter"].behavior.use_insert_replace_on = False
```
longvu-db: Not sure if I follow this test, so our behavior switch doesn't work?

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,296 @@
...
+    def test_get_insert_overwrite_sql__behavior_flag_disabled_old_dbr_cluster(
+        self, template, context, config
+    ):
+        """Test that behavior flag disabled on old DBR cluster uses traditional INSERT OVERWRITE"""
+        # Negative return value means DBR < 17.1
+        context["adapter"].compare_dbr_version.return_value = -1
+        # Cluster environment
+        context["adapter"].is_cluster.return_value = True
+        # Disable the behavior flag
+        context["adapter"].behavior.use_insert_replace_on = False
```
longvu-db: So what's the difference between `use_insert_replace_on` and `use_replace_on_for_insert_overwrite`?

longvu-db (commented):
Thanks for waiting for my review, had to take some time to learn about dbt's `insert_overwrite`, here is my first round of feedback/questions

dbt/include/databricks/macros/materializations/incremental/incremental.sql
ericj-db: Yes, that is the behavior when users configure `materialization = 'table'`

dbt/include/databricks/macros/materializations/incremental/strategies.sql
ericj-db: What kind of refactoring do you have in mind? Just extracting the duplicate logic from line 58 to line 74?

tests/functional/adapter/incremental/test_incremental_strategies.py
ericj-db: Could you share more details on this? This test runs against a serverless SQL warehouse in one of our internal workspaces. Are you saying that there is a way to override the DBR version at execution time?

tests/functional/adapter/incremental/test_incremental_strategies.py
ericj-db: Thanks for pointing this out, this is outdated. We'll keep it around as long as the behavior flag exists

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: Sorry I missed this when I was renaming the flag from `use_insert_replace_on` to `use_replace_on_for_insert_overwrite`. There should only be references to `use_replace_on_for_insert_overwrite` now

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: `is_cluster == True` here, so the behavior flag should have no impact. Since we already supported correct insert_overwrite behavior for clusters via DPO, it should be fine to make this the default behavior for clusters

dbt/adapters/databricks/connections.py
ericj-db: Yeah, this was just me being lazy. Separated out to https://github.com/databricks/dbt-databricks/pull/1158 so this PR is properly scoped

dbt/include/databricks/macros/materializations/incremental/strategies.sql
longvu-db: Yes, 58 -> 65, and 67 -> 74

longvu-db: Yes, 58 -> 65, and 67 -> 74

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,55 @@
...
-    insert overwrite table {{ target_relation }}
-    {{ partition_cols(label="partition") }}
-    select {{source_cols_csv}} from {{ source_relation }}
+    
+    {%- if (adapter.is_cluster() and adapter.compare_dbr_version(17, 1) >= 0) or (not adapter.is_cluster() and adapter.behavior.use_replace_on_for_insert_overwrite) -%}
+        {{ get_insert_replace_on_sql(source_relation, target_relation, source_cols_csv) }}
+    {%- else -%}
+        {#-- Use traditional INSERT OVERWRITE for older DBR versions and SQL warehouses with behavior flag disabled --#}
+        insert overwrite table {{ target_relation }}
+        {{ partition_cols(label="partition") }}
```
longvu-db, line 50: I still don't fully get where do we do DPO, which is setting the `partitionOverwriteMode` SQLConf to dynamic then do INSERT OVERWRITE

https://docs.databricks.com/aws/en/delta/selective-overwrite

This looks like we are doing the **static** partition overwriting
https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-dml-insert-into

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,55 @@
...
+        {%- set replace_conditions = [] -%}
+        {%- for col in replace_columns -%}
+            {%- do replace_conditions.append('t.' ~ col ~ ' <=> s.' ~ col) -%}
+        {%- endfor -%}
+        {%- set replace_conditions_csv = replace_conditions | join(' AND ') -%}
+        insert into table {{ target_relation }} AS t
+        replace on ({{ replace_conditions_csv }})
+        (select {{ source_cols_csv }} from {{ source_relation }}) AS s
+    {%- else -%}
+        {#-- Fallback to regular insert if no partitions or liquid clustering defined --#}
```
longvu-db:
```suggestion
        {#-- Fallback to regular insert if no partitions or liquid clustered columns defined --#}
```

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,55 @@
...
+        insert overwrite table {{ target_relation }}
```
longvu-db, line 88: So if no partitions or liquid clustered columns defined, then we are doing a static insert overwrite, which completely replaces `target` with `source` right?

dbt/adapters/databricks/impl.py
```diff
@@ -122,6 +122,15 @@
     ),
 )  # type: ignore[typeddict-item]
 
+USE_REPLACE_ON_FOR_INSERT_OVERWRITE = BehaviorFlag(
+    name="use_replace_on_for_insert_overwrite",
+    default=False,
+    description=(
+        "Use the new INSERT REPLACE ON syntax for insert_overwrite incremental strategy."
+        "  When disabled, falls back to legacy INSERT OVERWRITE syntax."
```
longvu-db: Only the partitionOverwriteMode=dynamic then INSERT OVERWRITE is legacy,

If we are using the static overwriting syntax, i.e. `INSERT OVERWRITE table [PARTITION clause]`, then it is not legacy.

https://docs.databricks.com/aws/en/delta/selective-overwrite

tests/functional/adapter/incremental/test_incremental_strategies.py
longvu-db: IIUC, this comment "Enable for SQL warehouses once 17.1+ is current version" is saying that we only enable this test, what I'm wondering is whether there is already a way to run this test before 17.1 is the current version, for max test coverage.

dbt/adapters/databricks/impl.py
longvu-db: REPLACE ON is only meant to replace the INSERT OVERWRITE with `partitionOverwriteMode=dynamic`, not the static INSERT OVERWRITE syntax.

tests/functional/adapter/incremental/test_incremental_strategies.py
```diff
@@ -120,6 +120,29 @@ def test_incremental(self, project):
...
 
+# TODO: Enable for SQL warehouses once 17.1+ is current version
+@pytest.mark.skip_profile("databricks_uc_sql_endpoint")
+class TestInsertOverwriteWithLiquidClusteringDelta(InsertOverwriteBase):
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "models": {
+                "+incremental_strategy": "insert_overwrite",
+                "+liquid_clustered_by": "id",
```
longvu-db, line 126: And what about a test for `partitioned_by` as well? Old and new behavior

dbt/adapters/databricks/impl.py
ericj-db: It's set here https://github.com/databricks/dbt-databricks/blob/cd652a7e7f2e4d626abf84ef623ad1de6229cbd8/dbt/include/databricks/macros/materializations/incremental/incremental.sql#L120-L123

dbt/include/databricks/macros/materializations/incremental/strategies.sql
ericj-db: It's set here https://github.com/databricks/dbt-databricks/blob/cd652a7e7f2e4d626abf84ef623ad1de6229cbd8/dbt/include/databricks/macros/materializations/incremental/incremental.sql#L120-L123

https://github.com/databricks/dbt-databricks/blob/cd652a7e7f2e4d626abf84ef623ad1de6229cbd8/dbt/include/databricks/macros/materializations/incremental/incremental.sql#L188-L196

dbt/include/databricks/macros/materializations/incremental/strategies.sql
ericj-db: Yes

tests/functional/adapter/incremental/test_incremental_strategies.py
ericj-db: The test right before this?

https://github.com/databricks/dbt-databricks/blob/cd652a7e7f2e4d626abf84ef623ad1de6229cbd8/tests/functional/adapter/incremental/test_incremental_strategies.py#L101-L120

tests/functional/adapter/incremental/test_incremental_strategies.py
ericj-db: I'm not aware of any way to do this and given that it will become current in early September, don't think it's worth investing into. We will enable these tests in 2-3 weeks when that happens. This feature won't be documented or advertised to customers and it is opt-in, so I doubt that we'll have customers attempting to use this in the next few weeks

dbt/adapters/databricks/impl.py
ericj-db: Updated the wording in 085fcded1df41a7819b1db6608690f279a12b6a7

dbt/include/databricks/macros/materializations/incremental/strategies.sql
ericj-db: Addressed in a3aefc148727179370e38de4de18aab56c6bc37d

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+            "get_insert_overwrite_sql",
+            source_relation,
+            target_relation,
+        )
+
+        # Verify it uses REPLACE ON syntax with multiple conditions
+        expected_sql = """
+            insert into table target_table as t
+            replace on (t.country <=> s.country AND t.name <=> s.name)
+            (select a, b from source_table) as s
```
longvu-db: So source_table contains column `a`, `b`, and `country`?

This expected SQL feels a bit weird, because the expected schema of `(select a, b from source_table) as s` is `a, b`, so not sure how is the REPLACE ON boolean condition able to reference the `country` column from `source_table`?

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+        """
+
+        self.assert_sql_equal(result, expected_sql)
+
+    def test_get_insert_overwrite_sql__modern_dbr_version_multiple_partitions(
+        self, template, context, config
+    ):
+        """Test that DBR >= 17.1 uses REPLACE ON syntax with multiple partition columns"""
+        # Positive return value means DBR > 17.1
```
longvu-db:
```suggestion
        # Positive return value means DBR >= 17.1
```

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+            insert overwrite table target_table
+            partition (partition_col)
+            select a, b from source_table
+        """
+
+        self.assert_sql_equal(result, expected_sql)
+
+    def test_get_insert_overwrite_sql__modern_dbr_version(self, template, context, config):
+        """Test that DBR >= 17.1 uses REPLACE ON syntax"""
+        # Positive return value means DBR > 17.1
```
longvu-db:
```suggestion
        # Positive return value means DBR >= 17.1
```

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+        target_relation.__str__ = lambda self: "target_table"
+
+        result = self.run_macro_raw(
+            template,
+            "get_insert_overwrite_sql",
+            source_relation,
+            target_relation,
+        )
+
+        # Verify it uses legacy INSERT OVERWRITE syntax
```
longvu-db:
```suggestion
        # Verify it uses legacy DPO INSERT OVERWRITE syntax
```

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+        expected_sql = """
+            insert overwrite table target_table
+            partition (partition_col)
+            select a, b from source_table
+        """
+
+        self.assert_sql_equal(result, expected_sql)
+
+    def test_get_insert_overwrite_sql__modern_dbr_version(self, template, context, config):
```
longvu-db: Nit: I feel like this two tests on `test_get_insert_overwrite_sql__legacy_dbr_version` and `test_get_insert_overwrite_sql__modern_dbr_version` can be parameterised for readability, cause there is only a one line difference in the `expected_sql`,  also "legacy DBR" and "modern DBR" can be confusing

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+        self.assert_sql_equal(result, expected_sql)
+
+    def test_get_insert_overwrite_sql__behavior_flag_disabled_sql_warehouse(
+        self, template, context, config
+    ):
+        """Test that behavior flag disabled on SQL warehouse uses traditional INSERT OVERWRITE"""
+        # Positive return value means DBR >= 17.1
+        context["adapter"].compare_dbr_version.return_value = 1
+        # SQL warehouse (not cluster)
+        context["adapter"].is_cluster.return_value = False
```
longvu-db: 1. Hmm, in the test below, `test_get_insert_overwrite_sql__behavior_flag_enabled_sql_warehouse`, 

"        # SQL warehouse (not cluster)
        context["adapter"].is_cluster.return_value = False"
        
=> My understanding is that when `is_cluster` is `False`, we are in SQL Warehouse.

=> `test_get_insert_overwrite_sql__behavior_flag_disabled_sql_warehouse`, if SQL warehouse is disabled, should `is_cluster` = True?

The difference between these 2 tests are the behavior flag.

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+        # Disable the behavior flag
+        context["adapter"].behavior.use_replace_on_for_insert_overwrite = False
```
longvu-db: I think we should combine these 2 tests `test_get_insert_overwrite_sql__behavior_flag_enabled_sql_warehouse` and `test_get_insert_overwrite_sql__behavior_flag_disabled_sql_warehouse` by parameterising the `use_replace_on_for_insert_overwrite` for readability

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+    def test_get_insert_overwrite_sql__behavior_flag_disabled_old_dbr_cluster(
+        self, template, context, config
+    ):
+        """Test that behavior flag disabled on old DBR cluster uses traditional INSERT OVERWRITE"""
+        # Negative return value means DBR < 17.1
+        context["adapter"].compare_dbr_version.return_value = -1
+        # Cluster environment
+        context["adapter"].is_cluster.return_value = True
+        # Disable the behavior flag
+        context["adapter"].behavior.use_replace_on_for_insert_overwrite = False
```
longvu-db: I wonder if we should test `context["adapter"].compare_dbr_version.return_value = -1` and `context["adapter"].behavior.use_replace_on_for_insert_overwrite = **True**` as well

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,53 @@
...
     {%- set dest_cols_csv = dest_columns | join(', ') -%}
     {%- set source_cols_csv = common_columns | join(', ') -%}
-    insert overwrite table {{ target_relation }}
-    {{ partition_cols(label="partition") }}
-    select {{source_cols_csv}} from {{ source_relation }}
+    
+    {%- if (adapter.is_cluster() and adapter.compare_dbr_version(17, 1) >= 0) or (not adapter.is_cluster() and adapter.behavior.use_replace_on_for_insert_overwrite) -%}
+        {{ get_insert_replace_on_sql(source_relation, target_relation, source_cols_csv) }}
+    {%- else -%}
+        {#-- Use traditional INSERT OVERWRITE for older DBR versions and SQL warehouses with behavior flag disabled --#}
```
longvu-db:
```suggestion
        {#-- Use legacy DPO INSERT OVERWRITE for older DBR versions and SQL warehouses with behavior flag disabled --#}
```

Nit

tests/functional/adapter/incremental/test_incremental_strategies.py
longvu-db: I see, didn't see this.

So this Liquid Clustering test won't be able till 17.1 per your comment? - https://github.com/databricks/dbt-databricks/pull/1135/files?w=1#r2294607996

longvu-db (commented):
What's the issue that you are resolving in the description? "Resolves #"

dbt/adapters/databricks/connections.py
```diff
@@ -398,8 +402,7 @@ def connect() -> DatabricksHandle:
             try:
                 # TODO: what is the error when a user specifies a catalog they don't have access to
                 conn = DatabricksHandle.from_connection_args(
-                    conn_args,
-                    is_cluster_http_path(databricks_connection.http_path, creds.cluster_id),
+                    conn_args, creds.cluster_id is not None
```
longvu-db: Out of curiosity, what is this case that we have to relax the condition from `is_cluster_http_path` -> `creds.cluster_id is not None`? Perhaps a comment here could be helpful

dbt/include/databricks/macros/materializations/incremental/strategies.sql
longvu-db: Hmm I see, thanks, the links between these 2 files aren't so obvious to me

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,53 @@
...
+        {%- set replace_conditions = [] -%}
+        {%- for col in replace_columns -%}
+            {%- do replace_conditions.append('t.' ~ col ~ ' <=> s.' ~ col) -%}
+        {%- endfor -%}
+        {%- set replace_conditions_csv = replace_conditions | join(' AND ') -%}
+        insert into table {{ target_relation }} AS t
+        replace on ({{ replace_conditions_csv }})
+        (select {{ source_cols_csv }} from {{ source_relation }}) AS s
+    {%- else -%}
+        {#-- Fallback to regular insert if no partitions or liquid clustering defined --#}
```
longvu-db:
```suggestion
        {#-- Fallback to regular insert overwrite if no partitioning nor liquid clustering defined --#}
```

Ultra nit

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+        # Verify it uses REPLACE ON syntax with clustering columns only
+        expected_sql = """
+            insert into table target_table as t
+            replace on (t.msg <=> s.msg)
+            (select a, b from source_table) as s
+        """
+
+        self.assert_sql_equal(result, expected_sql)
+
+    @pytest.mark.parametrize("dbr_version_return", [-1, 0, 1])
```
longvu-db, line 140: I thought -1 means < 17.1 and 1 means >= 17.1 like the first 2 tests way above? What's `0`?

https://github.com/databricks/dbt-databricks/pull/1135/files#diff-d08155e9cbc16dd90051518088de596d89e2d81f9bc24cf82ef711728e8f91b5R43

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+            template,
+            "get_insert_overwrite_sql",
+            source_relation,
+            target_relation,
+        )
+
+        # Verify it uses traditional INSERT OVERWRITE syntax for SQL warehouse without behavior flag
+        expected_sql = """
+            insert overwrite table target_table
+            partition (partition_col)
```
longvu-db:
```suggestion
```

Dumb qq, shouldn't this be just a normal INSERT OVERWRITE (so not DPO)?

https://github.com/databricks/dbt-databricks/pull/1135/files#diff-458a44b3009b62cbcba138e0305237db817902a9991c216f6cfd7590cce06316R194

If not a classic cluster, then we will not set `spark.sql.partitionOverwriteMode`, so we are doing the "table materialization" behavior

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,316 @@
...
+        self.assert_sql_equal(result, expected_sql)
+
+    def test_get_insert_overwrite_sql__modern_dbr_version_only_clustering(
+        self, template, context, config
+    ):
+        """
+        Test that DBR >= 17.1 uses REPLACE ON syntax with liquid clustering columns when no
+        partitioning is defined
+        """
+        # Positive return value means DBR > 17.1
```
longvu-db, line 115:
```suggestion
        # Positive return value means DBR >= 17.1
```

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: Positive value means strictly greater than. 0 means equality

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: negative means < 17.1
positive means > 17.1
0 means == 17.1

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: Even today, I think we still run the same statement regardless of compute type. For SQL warehouses, it just won't do anything. @benc-db is that right? If that's the case, I don't think there is value in making this change

I don't see a fork between SQL warehouse vs cluster in https://github.com/databricks/dbt-databricks/blob/87073fe7f26bede434a3bd783717a6e49d35893f/dbt/include/databricks/macros/materializations/incremental/strategies.sql#L28-L44

tests/functional/adapter/incremental/test_incremental_strategies.py
longvu-db: @ericj-db Does the flag `use_replace_on_for_insert_overwrite` apply for all customers, or we can somehow set it individually?

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
benc-db: The problem historically is that we don't have a fork here.  Then when we tried to block SQL Warehouse entirely, we got customer pushback.  So what happens is that today, SQL Warehouse runs without the DPO flag, and typically without partition information, which causes the equivalent of truncate + insert.

tests/functional/adapter/incremental/test_incremental_strategies.py
ericj-db: Yes, but we've internally decided to release this with dbt-databricks 1.11.0 in October instead of 1.10.x. By then, 17.1 should be current and I can just enable the test a week or two before release

tests/functional/adapter/incremental/test_incremental_strategies.py
ericj-db: It's not a global flag - customers will opt-in for their specific project

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: The behavior flag is completely independent of the compute type. The flag doesn't determine whether SQL warehouse is enabled. It determines whether `REPLACE ON` is enabled. These tests are verifying the different values for the behavior flag while using SQL warehouse. So `is_cluster` should be false for both

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: Yeah sorry AI added weird column names for partition by. Fixed in cca052692ae3f31023c57a75b3ae37993cf9213a to just use `a` and `b`

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: Addressed in 63e2d578b9069970a428e1fc85ab479cf0cc4965

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
longvu-db: @ericj-db I see, but qq, what does "fork" here mean?

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
longvu-db: Ah I see, I read it as SQL warehouse is enabled/disabled, but it should be behavior flag enabled/disabled, my bad

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,292 @@
...
+                1,  # DBR > 17.1
+                """
+                insert into table target_table as t
+                replace on (t.a <=> s.a)
+                (select a, b from source_table) as s
+                """,
+            ),
+        ],
+    )
+    def test_get_insert_overwrite_sql__dbr_version_syntax(
```
longvu-db:
```suggestion
    def test_get_dynamic_insert_overwrite_sql__dbr_version_syntax(
```

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,292 @@
...
+        self, template, context, config, dbr_version_comparison, expected_sql
+    ):
+        """Test that different DBR versions use appropriate INSERT OVERWRITE syntax"""
```
longvu-db:
```suggestion
        """Test that different DBR versions use appropriate dynamic insert overwrite syntax"""
```

Nit: INSERT OVERWRITE with capital refers to the INSERT OVERWRITE syntax, but dynamic insert overwrite refers to both the DPO with Spark Config knob and REPLACE ON/USING syntax.

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,292 @@
...
+        "dbr_version_comparison,expected_sql",
+        [
+            (
+                -1,  # DBR < 17.1
+                """
+                insert overwrite table target_table
+                partition (a)
+                select a, b from source_table
+                """,
+            ),
```
longvu-db, line 58: I think it's fine that `test_get_insert_overwrite_sql__behavior_flag_disabled_cluster`, `test_get_insert_overwrite_sql__modern_dbr_version_only_clustering`, `test_get_insert_overwrite_sql__modern_dbr_version_multiple_partitions` not tested with `compare_dbr_version` 0,

=> but I think we are missing a test that confirms REPLACE ON works on the exact 17.1 version in `test_get_dynamic_insert_overwrite_sql__dbr_version_syntax `

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,292 @@
...
+    def test_get_insert_overwrite_sql__modern_dbr_version_only_clustering(
+        self, template, context, config
+    ):
+        """
+        Test that DBR >= 17.1 uses REPLACE ON syntax with liquid clustering columns when no
+        partitioning is defined
+        """
+        # Positive return value means DBR > 17.1
+        context["adapter"].compare_dbr_version.return_value = 1
+        config["liquid_clustered_by"] = ["a"]
```
longvu-db:
```suggestion
        config["liquid_clustered_by"] = ["a", "b"]
```

Nit: Multiple columns will have slightly stronger coverage than single column.

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,292 @@
...
+        expected_sql = """
+            insert into table target_table as t
+            replace on (t.a <=> s.a)
+            (select a, b from source_table) as s
+        """
+
+        self.assert_sql_equal(result, expected_sql)
+
+    @pytest.mark.parametrize("dbr_version_return", [-1, 0, 1])
+    def test_get_insert_overwrite_sql__no_partitions(
```
longvu-db:
```suggestion
    def test_get_insert_overwrite_sql__no_partitions_and_liquid_clustered(
```

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,292 @@
...
+        self, template, context, config, dbr_version_return
+    ):
+        """Test that empty partition_by falls back to INSERT OVERWRITE regardless of DBR version"""
```
longvu-db:
```suggestion
        """Test that empty partition_by and liquid_clustered_by falls back to INSERT OVERWRITE regardless of DBR version"""
```

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,292 @@
...
+        result = self.run_macro_raw(
+            template,
+            "get_insert_overwrite_sql",
+            source_relation,
+            target_relation,
+        )
+
+        self.assert_sql_equal(result, expected_sql)
+
+    def test_get_insert_overwrite_sql__modern_dbr_version_multiple_partitions(
```
longvu-db: Nit: For readability, `test_get_insert_overwrite_sql__modern_dbr_version_multiple_partitions` and the liquid clustering test below can be parameterised as well, but fine with me either way.

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: 7430cd36b19766f63ad36528a3dd008e0a2c317f

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,282 @@
...
+        )
+
+        self.assert_sql_equal(result, expected_sql)
+
+    @pytest.mark.parametrize(
+        "config_key,config_value,test_description",
+        [
+            ("partition_by", ["a", "b"], "multiple partition columns"),
+            (
+                "liquid_clustered_by",
```
longvu-db, line 106: I think the config_value is `unchanged` here, so we can just hard code it.

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,282 @@
...
+            source_relation,
+            target_relation,
+        )
+
+        self.assert_sql_equal(result, expected_sql)
+
+    @pytest.mark.parametrize(
+        "config_key,config_value,test_description",
+        [
+            ("partition_by", ["a", "b"], "multiple partition columns"),
```
longvu-db: Let's format the `partition_by` and `liquid_clustered_by` the same way for readability (so either all multiple lines or only one line)

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -191,7 +191,7 @@
...
+    {{ exceptions.warn("insert_overwrite is supported on SQL warehouses with DBR 17.1+. On older DBR versions, this strategy would be equivalent to using the table materialization.") }}
```
longvu-db, line 194: Qq: When `adapter.is_cluster() and adapter.compare_dbr_version(17, 1) >= 0`, which means we will use the REPLACE ON syntax, would we still do `set spark.sql.sources.partitionOverwriteMode = {{ value }}` based on the `if` branch above?

Albeit this feels weird, I guess it should be fine.

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: Done 9a2534b6725db7a8b92b299f6093fa0542fa4d51

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
ericj-db: Done 9a2534b6725db7a8b92b299f6093fa0542fa4d51

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Support databricks_tags for MV/STs
 - Add support for scalar SQL functions (SQL UDFs) ([1197](https://github.com/databricks/dbt-databricks/pull/1197))
 - Add liquid clustering config for materialized views and streaming tables (thanks @reflection!) ([1101](https://github.com/databricks/dbt-databricks/pull/1101))
+- Support `insert_overwrite` incremental strategy for SQL warehouses using `REPLACE ON` for null tolerance, gated behind behavior flag ([1025](https://github.com/databricks/dbt-databricks/issues/1025))
```
longvu-db: 1. Should we call out loudly that the new behavior of `insert_overwrite` is using REPLACE ON instead of table truncation?

2. I have a feeling mentioning "null tolerance" here might be confusing, do users need to know about this internal detail? 

Instead of "Support `insert_overwrite` incremental strategy for SQL warehouses using `REPLACE ON` for null tolerance", I wonder if the messaging should be something along the line of "insert_overwrite now officially works on SQL Warehouses", because it gives the same behavior as in non-SQL Warehouses, and it now behaves the intended way.

=> This motivates users to now officially use `insert_overwrite`

dbt/adapters/databricks/impl.py
```diff
@@ -127,6 +127,16 @@
     ),
 )  # type: ignore[typeddict-item]
 
+USE_REPLACE_ON_FOR_INSERT_OVERWRITE = BehaviorFlag(
+    name="use_replace_on_for_insert_overwrite",
+    default=True,
+    description=(
+        "Use the new INSERT REPLACE ON syntax for insert_overwrite incremental strategy."
+        "  When disabled, falls back to legacy INSERT OVERWRITE syntax with dynamic partition"
```
longvu-db, line 135: When we disable, we fall back to the incorrect table materialization behavior right?

dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -191,7 +191,7 @@
       set spark.sql.sources.partitionOverwriteMode = {{ value }}
     {%- endcall -%}
   {% else %}
```
longvu-db, line 193: Since we are changing the default to `true`, should we put a add a WARNING message somewhere saying that we are doing a parition overwrite, and not truncating?

tests/unit/macros/materializations/incremental/test_insert_overwrite_macros.py
```diff
@@ -0,0 +1,286 @@
...
+    def test_get_insert_overwrite_sql__old_dbr_cluster_behavior_flag(
+        self, template, context, config, use_replace_on_flag
+    ):
+        """Old DBR cluster always uses traditional INSERT OVERWRITE regardless of behavior flag"""
+        # Negative return value means DBR < 17.1
+        context["adapter"].compare_dbr_version.return_value = -1
+        # Cluster environment
+        context["adapter"].is_cluster.return_value = True
+        # Set the behavior flag (should not affect old DBR clusters)
+        context["adapter"].behavior.use_replace_on_for_insert_overwrite = use_replace_on_flag
```
longvu-db, line 217: Btw, I thought when `use_replace_on_for_insert_overwrite` is false, we do a complete overwrite, i.e. `INSERT OVERWRITE target SELECT ... FROM source`, why do we have `partition` in the expected answer? Is it because we only have partition? Do we have a test somewhere that actually tests the full overwrite behavior?

longvu-db (commented):
Final round of review on the new behavior of `true`

dbt/adapters/databricks/impl.py
ericj-db: Yes, that's true for SQL warehouses. For other compute types, it will fallback to the old syntax as well but functionally behave the same

CHANGELOG.md
ericj-db: Updated the wording in b58547dcf231ad656555861bfaa57fbb201bbd7f

dbt/adapters/databricks/impl.py
longvu-db: In that case, could we just quickly add this small comment "that's true for SQL warehouses. For other compute types, it will fallback to the old syntax as well but functionally behave the same" here, due to this behavior difference between SQL Warehouse and non-SQL warehouse?

It's to ensure that users will be aware of the expected behavior when they use this flag.

ericj-db: Sure, does the messaging update in 693614945f11f39f273fb95e50e52286feaf2cb9 make sense?

dbt/include/databricks/macros/materializations/incremental/incremental.sql
ericj-db: This specific block is only executed when the compute type is a cluster (i.e. non SQL warehouse). Do we need to warn users of the `REPLACE ON` vs `INSERT OVERWRITE` when they are functionally the same? Or maybe I am misunderstanding what the suggested warning is for

longvu-db: @ericj-db I think the goal is we should warn SQL Warehouse users, who relied on the table truncation behavior (aka static INSERT OVERWRITE), that we are now doing REPLACE ON, which is DYNAMIC INSERT OVERWRITE.

longvu-db: So I was referring to the other branch

ericj-db: Ah okay got it.. Just added a warning in 80527a77428cbf70fb56eb2e9910834ff48bdea3. Unfortunately we don't have a great way to verify log messages in unit tests, but I have manually verified that it gets printed for SQL warehouses

Screenshot 2025-10-07 at 9 49 40 AM

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,56 @@
     {%- endfor -%}
     {%- set dest_cols_csv = dest_columns | join(', ') -%}
     {%- set source_cols_csv = common_columns | join(', ') -%}
-    insert overwrite table {{ target_relation }}
-    {{ partition_cols(label="partition") }}
-    select {{source_cols_csv}} from {{ source_relation }}
+    
+    {%- if (adapter.is_cluster() and adapter.compare_dbr_version(17, 1) >= 0) or (not adapter.is_cluster() and adapter.behavior.use_replace_on_for_insert_overwrite) -%}
+        {%- if not adapter.is_cluster() %}
+            {{ exceptions.warn("insert_overwrite will perform a partition overwrite. If you depended on the legacy truncation behavior, consider disabling the behavior flag use_replace_on_for_insert_overwrite.") }}
```
longvu-db:
```suggestion
            {{ exceptions.warn("insert_overwrite will perform a dynamic insert overwrite. If you depended on the legacy truncation behavior, consider disabling the behavior flag use_replace_on_for_insert_overwrite.") }}
```

One last nit, so that it's clear that we are doing a dynamic insert overwrite instead of a static insert overwrite, and we can overwrite liquid clustereds, not just partitions as well.

ericj-db: Thanks, just applied the suggestion

longvu-db (commented):
LGTM again, after the default behavior is changed. Thanks @ericj-db very much for addressing one more round of comments and taking the time to verify the warning message!

dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -38,9 +38,56 @@
...
+            {%- for col in columns -%}
+                {%- do target_list.append(col) -%}
+            {%- endfor -%}
+        {%- endif -%}
+    {%- endif -%}
+{% endmacro %}
+
+{% macro get_insert_replace_on_sql(source_relation, target_relation, source_cols_csv) %}
+    {%- set partition_by = config.get('partition_by') -%}
+    {%- set liquid_clustered_by = config.get('liquid_clustered_by') -%}
```
tomvanbussel, line 70: If this is a new config then I would just suggest naming it just `clustered_by`.

longvu-db:
Just in case we forgot, could we please update the PR description to reflect the latest state? i.e. support liquid_clustered_by and default = False

benc-db (approved):

PR #1139, opened 2025-08-07
Add CREATE OR REPLACE branch for materialized views

ericj-db:
Resolves #

### Description

`CREATE OR REPLACE MATERIALIZED VIEW` has been supported since DBR 14.1. Integrate with this syntax to provide atomic replace as opposed to the current drop + create strategy

Streaming tables do not support `REPLACE` yet according to https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-ddl-create-streaming-table

### Checklist

### DIFF
--- a/dbt/adapters/databricks/relation.py
+++ b/dbt/adapters/databricks/relation.py
@@ -76,7 +76,11 @@ class DatabricksRelation(BaseRelation):
     alter_constraints: list[TypedConstraint] = field(default_factory=list)
     metadata: Optional[dict[str, Any]] = None
     renameable_relations = (DatabricksRelationType.Table, DatabricksRelationType.View)
-    replaceable_relations = (DatabricksRelationType.Table, DatabricksRelationType.View)
+    replaceable_relations = (
+        DatabricksRelationType.Table,
+        DatabricksRelationType.View,
+        DatabricksRelationType.MaterializedView,
+    )
     databricks_table_type: Optional[DatabricksTableType] = None
     temporary: Optional[bool] = False
 
@@ -129,8 +133,7 @@ class DatabricksRelation(BaseRelation):
         return (
             self.type == DatabricksRelationType.View
             or self.type == DatabricksRelationType.MaterializedView
-            or self.is_delta is True
-            and self.type == DatabricksRelationType.Table
+            or (self.is_delta is True and self.type == DatabricksRelationType.Table)
         )
 
     @property


dbt/include/databricks/macros/relations/replace.sql
```diff
@@ -19,6 +19,8 @@
       {{ return(get_replace_view_sql(target_relation, sql)) }}
     {% elif is_replaceable and existing_relation.is_table %}
       {{ return(get_replace_table_sql(target_relation, sql)) }}
+    {% elif is_replaceable and existing_relation.is_materialized_view %}
```
benc-db, line 22: You need to update relation so that it knows that materialized view can be replaced.  I suspect this branch cannot be hit currently.

ericj-db: Ah yeah good catch.. I initially manually verified by running this test and checking the dbt logs https://github.com/dbt-labs/dbt-adapters/blob/6dc698497b4123813210f332b09f04960b0a04f7/dbt-tests-adapter/src/dbt/tests/adapter/materialized_view/changes.py#L168

I saw `Applying REPLACE to:`, so I thought we were good. But that actually gets logged regardless of what strategy we choose to perform the replace. Further down the log, I do see `Using drop_and_create`

Fixed this in 8ca1df25bd7253761fefc5fe7a494bf1338855a8 and re-verified via logs that the correct path is being used and we do not see the drop and create logs anymore

dbt/adapters/databricks/relation.py
```diff
@@ -124,6 +124,7 @@ def stats(self) -> Optional[str]:
     def can_be_replaced(self) -> bool:
         return (
             self.type == DatabricksRelationType.View
+            or self.type == DatabricksRelationType.MaterializedView
             or self.is_delta is True
             and self.type == DatabricksRelationType.Table
```
benc-db: Can we wrap the last compound condition?  I think python interprets this correctly, but would like to use parentheses to ensure.

benc-db: I think there is another attribute in Relation that lists the replacable types.  I don't think we reference it, but in case anyone else does, can you update that as well?

benc-db (changes requested):
Very minor changes requested

dbt/adapters/databricks/relation.py
ericj-db: Sure, addressed in 4bdad29fd73efbcb0d1b53fa938eab7b82e52c04

PR #1142, opened 2025-08-11
Support databricks_tags for MV/STs

ericj-db:
Resolves #

### Description

- Reuse the existing `apply_tags` macro for MV/STs and add functional tests for validation
- Add a new `DatabricksRelationType.render()` function. We are using `ALTER {{ relation.type }}` in many existing macros. This does not work for types that have underscores like materialized_view and streaming_table. Updated the macros to use `ALTER {{ relation.type.render() }}` instead

### Checklist

### DIFF
--- a/dbt/adapters/databricks/relation.py
+++ b/dbt/adapters/databricks/relation.py
@@ -43,6 +43,10 @@ class DatabricksRelationType(StrEnum):
     MetricView = "metric_view"
     Unknown = "unknown"
 
+    def render(self) -> str:
+        """Return the type formatted for SQL statements (replace underscores with spaces)"""
+        return self.value.replace("_", " ")
+
 
 class DatabricksTableType(StrEnum):
     External = "external"
--- a/dbt/include/databricks/macros/adapters/persist_docs.sql
+++ b/dbt/include/databricks/macros/adapters/persist_docs.sql
@@ -28,7 +28,7 @@ COMMENT ON COLUMN {{ column_path }} IS '{{ escaped_comment }}'
 {% endmacro %}
 
 {% macro alter_relation_comment_sql(relation, description) %}
-COMMENT ON {{ relation.type.upper() }} {{ relation.render() }} IS '{{ description | replace("'", "\\'") }}'
+COMMENT ON {{ relation.type.render().upper() }} {{ relation.render() }} IS '{{ description | replace("'", "\\'") }}'
 {% endmacro %}
 
 {% macro alter_column_comments(relation, column_dict) %}
--- a/dbt/include/databricks/macros/materializations/materialized_view.sql
+++ b/dbt/include/databricks/macros/materializations/materialized_view.sql
@@ -64,8 +64,11 @@
     {{ run_hooks(pre_hooks, inside_transaction=True) }}
 
     {% set grant_config = config.get('grants') %}
+    {% set tags = config.get('databricks_tags') %}
 
     {{ execute_multiple_statements(build_sql) }}
+
+    {%- do apply_tags(target_relation, tags) -%}
 
     {% set column_tags = adapter.get_column_tags_from_model(config.model) %}
     {% if column_tags %}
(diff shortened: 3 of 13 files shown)


dbt/include/databricks/macros/adapters/persist_docs.sql
```diff
@@ -28,7 +28,7 @@ COMMENT ON COLUMN {{ column_path }} IS '{{ escaped_comment }}'
 {% endmacro %}
 
 {% macro alter_relation_comment_sql(relation, description) %}
-COMMENT ON {{ relation.type.upper() }} {{ relation.render() }} IS '{{ description | replace("'", "\\'") }}'
+COMMENT ON {{ relation.type.render().upper() }} {{ relation.render() }} IS '{{ description | replace("'", "\\'") }}'
```
benc-db, line 31: why not do the upper in the render?

dbt/adapters/databricks/relation.py
```diff
@@ -43,6 +43,10 @@ class DatabricksRelationType(StrEnum):
     MetricView = "metric_view"
     Unknown = "unknown"
 
+    def render(self) -> str:
```
benc-db, line 46: are you able to replace the to string method?  Then you wouldn't need to make the calls to render

dbt/include/databricks/macros/adapters/persist_docs.sql
ericj-db: I don't think that's safe because of other usages like https://github.com/databricks/dbt-databricks/blob/92f1442faabe0fce6f0375b95e46ebcbfcea4c67/dbt/include/databricks/macros/relations/components/column_tags.sql#L38-L40

I'm open to updating these usages if you think that's cleaner

dbt/adapters/databricks/relation.py
ericj-db: Same as https://github.com/databricks/dbt-databricks/pull/1142#discussion_r2271159638, it will affect other macro usages

dbt/include/databricks/macros/adapters/persist_docs.sql
benc-db: I meant specifically for relation.type.render, which we have control over.

dbt/adapters/databricks/relation.py
benc-db: just for relation type, not for relation.

dbt/include/databricks/macros/adapters/persist_docs.sql
ericj-db: Oh yeah I mean line 38, we're doing `relation.type == 'view'`. We would need to update it to `relation.type == 'VIEW'`, right? (if we override the `__str__` function)

benc-db: I mean line 31.  You have relation.type.render().upper().  You introduced relation type render in this PR, so I'm wondering why not include upper inside its definition.

ericj-db: Sorry, I'm kind of combining your 2 suggestions in this thread and maybe that's causing some confusion. If we apply both suggestions, we get this
```
DatabricksRelationType(StrEnum):
  def __str__(self):
    return self.value.replace("_", " ").upper()
```

My point is that we are not only using `relation.type` as a drop-in for `ALTER {{ relation.type }} ...` statements. We are also using it in places like https://github.com/databricks/dbt-databricks/blob/92f1442faabe0fce6f0375b95e46ebcbfcea4c67/dbt/include/databricks/macros/relations/components/column_tags.sql#L38-L40 

(I'm referring to line 38 in this snippet). We would need to update that to `{%- if relation.type == 'VIEW' -%} `. Is that what we want? I wasn't sure if this is a risky change if macros outside of this package rely on `relation.type`

benc-db: ah, yeah, this is a reason to not override the to string, but we can move the upper into render.

ericj-db: Addressed in a9bfe9fc4438ed14ee25c8be47be164df54704b5

ericj-db:
> Don't we want to use the relation config machinery to ensure we detect changes?

Yeah my bad, totally missed this. Addressed this and added functional tests for applying changes in 3e7ca842faa87a86e7e7285bf5d9a33ace8e17cf

benc-db (approved):

PR #1143, opened 2025-08-14
Adjust order of dbt dependency installation

ericj-db:
Resolves #

### Description

Protobuf version was updated by our dbt dependencies:
- https://github.com/dbt-labs/dbt-adapters/pull/1259
- https://github.com/dbt-labs/dbt-common/pull/300
- https://github.com/dbt-labs/dbt-core/pull/11916

This broke our CI because pip is installing published pypi versions of transitive dependencies. The fix is to install dependencies in order, starting from leaf nodes

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,12 +64,12 @@ check-sdist = [
 
 [tool.hatch.envs.default]
 pre-install-commands = [
-    "pip install git+https://github.com/dbt-labs/dbt-adapters.git#subdirectory=dbt-adapters",
+    "pip install git+https://github.com/dbt-labs/dbt-common.git",
+    "pip install git+https://github.com/dbt-labs/dbt-adapters.git@main#subdirectory=dbt-adapters",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -64,12 +64,12 @@ check-sdist = [
 
 [tool.hatch.envs.default]
 pre-install-commands = [
-    "pip install git+https://github.com/dbt-labs/dbt-adapters.git#subdirectory=dbt-adapters",
+    "pip install git+https://github.com/dbt-labs/dbt-common.git",
```
benc-db, line 67: Why do these need to be pre-install and not just dependencies?

ericj-db: When they are listed under `dependencies`, hatch does not install them in the order they are listed in. So `dbt-core` can get installed first and it will install a published version of `dbt-common` as a transitive dependency. Then when `dbt-common` gets installed, it detects the latest version is already installed and does not install from github main

PR #1146, opened 2025-08-15
feat: Add managed iceberg support behind behavior flag

ericj-db:
Resolves #

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -122,6 +122,15 @@ USE_MATERIALIZATION_V2 = BehaviorFlag(
     ),
 )  # type: ignore[typeddict-item]
 
+USE_MANAGED_ICEBERG = BehaviorFlag(
+    name="use_managed_iceberg",
+    default=False,
+    description=(
+        "Use managed Iceberg tables when table_format is iceberg. Always uses Parquet as the file"
+        " format. When this flag is disabled, UniForm with Delta is used instead."
+    ),
+)  # type: ignore[typeddict-item]
+
 
 class DatabricksRelationInfo(NamedTuple):
     table_name: str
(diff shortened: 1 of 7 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -122,6 +122,15 @@
     ),
 )  # type: ignore[typeddict-item]
 
+USE_MANAGED_ICEBERG = BehaviorFlag(
+    name="use_managed_iceberg",
+    default=False,
+    description=(
+        "Use managed Iceberg tables when table_format is iceberg. Only works with Parquet as the"
+        " file_format. When this flag is disabled, UniForm with Delta is used instead."
```
benc-db: Should we maybe just ignore file_format when using table format?  Feels like just a way to trip users up since ultimately its all variants of parquet anyway.

dbt/adapters/databricks/impl.py
```diff
@@ -240,7 +254,22 @@ def update_tblproperties_for_iceberg(
                     "When table_format is 'iceberg', materialized must be 'incremental'"
                     ", 'table', or 'snapshot'."
                 )
-            result.update(catalog_relation.iceberg_table_properties)
+            # Only add UniForm properties if not using managed Iceberg tables
+            if not self.behavior.use_managed_iceberg:
+                result.update(catalog_relation.iceberg_table_properties)
+            elif (
```
benc-db: Really weird place for this, since you're not updating tblproperties for iceberg in the changes you've made.  I think we actually should not call this method at all if use_managed_iceberg is true.  The decision about raising an error on HMS should probably also happen before we get to this method, because the failure isn't coming from updating the tblproperties, but from the model being misconfigured.  So, if we want to check the configuration validity in python, let's make that method separate, and just call to update the iceberg table properties if a.) we pass the validity check and b.) the flag is false.

benc-db: Basically, I'm acknowledging that I didn't make an extensible choice when I initially did my config checking here; that was the mistake, but let's not double down on it :)

dbt/adapters/databricks/impl.py
ericj-db: Yeah we could do that, I was mainly worried that there would be some macros relying on the file format. But even if that is the case, we could probably just hide that detail from the users and internally set the file_format correctly depending on the table_format?

dbt/adapters/databricks/impl.py
ericj-db: Ah yeah I thought it was a bit weird but was lacking the context. I'll refactor this

dbt/adapters/databricks/impl.py
ericj-db: Addressed in 6ff6a3245aadf6e3c1e4baa1c3d86f8e420e1a91

dbt/adapters/databricks/impl.py
ericj-db: Addressed in b9806117640701125633d024fb84bfdeff4f5ac2

PR #1147, opened 2025-08-18
chore: Tighten version range for higher risk dependencies

ericj-db:
Resolves #

### Description

Tighten versions for some upstream dependencies that have broken us in the past. For dbt packages, pin patch version as they frequently contain significant changes. The only exception is `dbt-common` which rarely has patch versions (https://pypi.org/project/dbt-common/#history), so just pinning minor version for that

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 ]
 dependencies = [
     "databricks-sdk>=0.41, <0.48.0",
-    "databricks-sql-connector[pyarrow]>=4.0.0, <4.1.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -23,11 +23,11 @@ classifiers = [
 ]
 dependencies = [
     "databricks-sdk>=0.41, <0.48.0",
```
benc-db: While we're making changes, can we test a newer version of the SDK?  Have been asked for this by a customer.

ericj-db: Bumped the version and tests look good. Good to merge?

The failed integration test is unrelated (will look into it shortly)

PR #1161, opened 2025-08-22
fix: Remove external path on intermediate tables for python incremental models

gsolasab:
### Description
This update changes how intermediate tables are handled for Python incremental models (with Materialization V2). Previously, when Python models were configured to save as external tables, their intermediate tables were also stored in the external path. This caused two main issues:

- Concurrency problems: Even when using unique_tmp_table_suffix=True, all runs shared the same external path, leading to conflicts.
- Unnecessary storage: Data from intermediate tables was not deleted when the tables were dropped, leaving unwanted files in the storage account.

With this change, intermediate tables will be stored as managed tables, which will:

- Prevent concurrency issues by ensuring each run uses a unique location.
- Automatically remove intermediate data after dbt execution, keeping storage clean and efficient.

### Checklist


benc-db:
Hmm, this is an interesting idea.  Going to run the test suite and see if this plays nicely with everything.

gsolasab:
Hello,
I have changed the description of the PR to follow your guidelines. Tell me if anything else is needed.

Thanks!

Regards

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Improve ANSI mode error handling for Python models and add debug instrumentation ([1157](https://github.com/databricks/dbt-databricks/pull/1157))
+- Remove external path on intermediate tables for incremental models ([1161](https://github.com/databricks/dbt-databricks/pull/1161))
```
benc-db: Looks like this only applies to Materialization V2 (which makes sense because we do almost all the table stuff in SQL in V2, rather than in Python), so you may want to mention materialization V2 in your change description

benc-db (approved):

CHANGELOG.md
gsolasab: done

PR #1189, opened 2025-09-20
[WIP] Backport external table clone on top of 1.9.latest

mi-volodin:
This is a backport of the changes introduced in #1079 and #1095.

Not everything is applied cleanly as there are some major changes between 1.9 and 1.10 in terms of metadata management and testing.

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -303,7 +303,7 @@ class DatabricksAdapter(SparkAdapter):
 
     def list_relations_without_caching(  # type: ignore[override]
         self, schema_relation: DatabricksRelation
-    ) -> List[DatabricksRelation]:
+    ) -> list[DatabricksRelation]:
         empty: list[DatabricksRelationInfo] = []
 
         results = handle_missing_objects(


dbt/adapters/databricks/impl.py
```diff
@@ -293,24 +303,32 @@ def execute(
 
     def list_relations_without_caching(  # type: ignore[override]
         self, schema_relation: DatabricksRelation
-    ) -> list[DatabricksRelation]:
-        empty: list[tuple[Optional[str], Optional[str], Optional[str], Optional[str]]] = []
+    ) -> List[DatabricksRelation]:
```
samgans: You are not importing any `List` object from `typing`.


```suggestion
    ) -> list[DatabricksRelation]:
```

This should still be fine, because Python 3.9 supports the typing info on the objects themselves.

And the 1.9 version of the package accepts only 3.9 and higher.

tests/unit/test_adapter.py
```diff
@@ -11,10 +11,18 @@
 from dbt.adapters.databricks.column import DatabricksColumn
 from dbt.adapters.databricks.credentials import (
     CATALOG_KEY_IN_SESSION_PROPERTIES,
+    DBT_DATABRICKS_HTTP_SESSION_HEADERS,
+    DBT_DATABRICKS_INVOCATION_ENV,
```
samgans, line 15: Why is it there? I don't see any mentions of those in the code.

tests/unit/utils.py
```diff
@@ -1,5 +1,8 @@
 import os
 from copy import deepcopy
+from typing import Optional
+
+from dbt.adapters.databricks import catalogs, constants
```
mi-volodin, line 5: @samgans I think all the changes in this files is unrelated to your code, is that correct?

PR #1209, opened 2025-10-07
Add support for Databricks query tags

ericj-db:
### Description

Add a new `query_tags` config at two different levels:
- Connection level (e.g. `~/.dbt/profiles.ym`)
- Model level (e.g. `~/.dbt/dbt_project.yml`)

The following default tags are set based on the query context:
- dbt_model_name (only for queries tied to a model)
- dbt_materialized (only for queries tied to a model)
- dbt_core_version (for all queries)
- dbt_databricks_version (for all queries)

For user configured tag overrides, the expected value is a JSON string such as `'{"team": "marketing", "cost_center": "3000"}'`

### Testing

Aside from the unit tests, I have manually verified that query tags are set by querying the query history system table. Unfortunately, doing this systematically would be too flaky as system tables don't have a strict SLO. In my manual test, I set the following overrides:
- Connection level: `query_tags: '{"team": "marketing_project_override", "project_only_tag": "something"}'`
- Model level: `query_tags: '{"team": "marketing_model_override", "model_only_tag": "something_else"}'`

The screenshot below shows query history records for two queries: the first one is not tied to a specific model execution, while the second one is. This explains why the query tags for the first record does not include model-specific tags

Screenshot 2025-10-07 at 3 11 07 PM

### Checklist


tests/unit/test_query_tags.py
```diff
@@ -0,0 +1,210 @@
...
+    def test_parse_query_tags_valid_json(self):
+        """Test parsing valid JSON query tags."""
+        tags_json = '{"team": "marketing", "cost_center": "3000"}'
+        expected = {"team": "marketing", "cost_center": "3000"}
+        assert QueryTagsUtils.parse_query_tags(tags_json) == expected
+
+    def test_parse_query_tags_converts_values_to_strings(self):
+        """Test that all values are converted to strings."""
+        tags_json = '{"team": "marketing", "cost_center": 3000, "active": true}'
+        expected = {"team": "marketing", "cost_center": "3000", "active": "True"}
```
ericj-db: @benc-db I wonder if this will become a source of bugs/confusion for customers. During parsing, the casing of booleans would change to Python style (i.e. first letter upper cased)

dbt/adapters/databricks/connections.py
```diff
@@ -32,6 +33,7 @@
 from databricks.sql import __version__ as dbsql_version
 from databricks.sql.exc import Error
 from dbt.adapters.databricks.__version__ import version as __version__
+from dbt.adapters.databricks.__version__ import version as databricks_version
```
benc-db: why both of these lines?

dbt/adapters/databricks/connections.py
```diff
@@ -480,3 +515,47 @@ def get_http_path(context: QueryContextWrapper, creds: DatabricksCredentials) ->
...
+    @staticmethod
+    def get_merged_query_tags(
+        query_header_context: Any,
+        creds: DatabricksCredentials,
+    ) -> dict[str, str]:
+        """
+        Get merged query tags from connection config, model config, and default tags.
+        Model config takes precedence over connection config.
+        """
+        from dbt.adapters.databricks.utils import QueryTagsUtils
```
benc-db: Move to top...classic AI move :P

dbt/adapters/databricks/handle.py
```diff
@@ -297,8 +297,13 @@ def clean_sql(sql: str) -> str:
 
     @staticmethod
     def prepare_connection_arguments(
-        creds: DatabricksCredentials, creds_manager: DatabricksCredentialManager, http_path: str
+        creds: DatabricksCredentials,
+        creds_manager: DatabricksCredentialManager,
+        http_path: str,
+        query_tags: Optional[dict[str, str]] = None,
     ) -> dict[str, Any]:
+        from dbt.adapters.databricks.utils import QueryTagsUtils
```
benc-db: move to top

dbt/adapters/databricks/utils.py
```diff
@@ -96,3 +97,99 @@ def is_cluster_http_path(http_path: str, cluster_id: Optional[str]) -> bool:
...
+                f"Too many total query tags ({len(merged)}). "
+                f"Maximum allowed is {QueryTagsUtils.MAX_TAGS}"
+            )
+
+        return merged
+
+    @staticmethod
+    def format_query_tags_for_databricks(tags: dict[str, str]) -> str:
+        """Format query tags for Databricks session configuration (without quotes)."""
+        if not tags:
```
benc-db: This is already taken care of without this return.  If tags is in fact a dict, then an empty dict will return "" from the final two lines of this method.

tests/unit/test_query_tags.py
benc-db: should tags ever be treated as anything other than strings?

ericj-db: Probably not.. Should we just validate the input and throw an error if we detect a non-string value in the JSON?

dbt/adapters/databricks/connections.py
ericj-db: Oops did not realize it was already imported. Addressed this comment and a few others in 01001cb487806af6459e665d6c67844f5fc586ee

tests/unit/test_query_tags.py
ericj-db: Added explicit handling in 6a20679cd4cefcc9fed517094261305c9be597c5

PR #1212, opened 2025-10-08
fix: column order mismatch in incremental models with schema changes

### LINKED ISSUE #1211: Bug: Column order mismatch in incremental models with on_schema_change causes data corruption
## Describe the bug

When using `on_schema_change: sync_all_columns` with incremental models, dbt-databricks adds new columns to the target table but subsequent INSERT statements use `SELECT *` without explicit column specifications. This causes column order mismatches where values are inserted into the wrong columns, resulting in data corruption.

## Steps To Reproduce

1. Create an incremental model with `on_schema_change: sync_all_columns`
2. Run the model initially with columns: `id, name, status`
3. Modify the model to add a new column in a different position: `id, new_column, name, status`
4. Run the model again
5. Observe that values intended for `name` are inserted into `new_column`, etc.

## Expected behavior

Values should be inserted into the correct columns regardless of column order in the source query.

## Screenshots and log output

The problematic code is in `/dbt/include/databricks/macros/relations/table/create.sql` line 21:

```sql
{% call statement('merge into target') %}
    insert into {{ target_relation }} select * from {{ intermediate_relation }}
{% endcall %}
```

This `SELECT *` approach relies on positional matching, which breaks when column order changes.

Similar issues exist in:
- `/dbt/include/databricks/macros/materializations/incremental/strategies.sql` in various INSERT macros

## System information

- Multiple versions affected (all current versions)

- All platforms

- All Python versions

## Which database adapter are you using with dbt?

databricks

### Proposed Solution

Use Databricks `INSERT BY NAME` syntax to let the database handle column matching by name instead of position:

```sql
insert into {{ target_relation }} by name
select * from {{ intermediate_relation }}
```

### Benefits
- Fixes data corruption bug
- Simplifies code (no need to explicitly list and match columns)
- Aligns with Databricks best practices

### Requirements
- Requires DBR 12.2 LTS or higher (released ~2022)
- Users on older runtimes would need to pin to previous adapter version

### Affected Code Locations
1. `/dbt/include/databricks/macros/relations/table/create.sql:21` - `create_table_at()` macro
2. `/dbt/include/databricks/macros/materializations/incremental/strategies.sql:67-80` - `insert_into_sql_impl()` macro
3. `/dbt/include/databricks/macros/materializations/incremental/strategies.sql:28-44` - `get_insert_overwrite_sql()` macro
4. `/dbt/include/databricks/macros/materializations/incremental/strategies.sql:46-59` - `get_replace_where_sql()` macro

### Testing Needed
- Functional tests with `on_schema_change: sync_all_columns` adding columns in different positions
- Tests across all three strategies: append, insert_overwrite, replace_where
- Tests across all three compute environments: HMS cluster, UC cluster, SQL Warehouse
- Unit tests for macro SQL generation

### Alternative Approach
Could implement as feature flag (`insert_by_name: true/false`), but since current behavior is objectively broken (data corruption), recommend implementing as direct bug fix.

randypitcherii:
Fixes #1211

## Summary
Uses Databricks `INSERT BY NAME` syntax to fix data corruption bug when `on_schema_change: sync_all_columns` adds columns in different order.

## Problem
When using `on_schema_change: sync_all_columns` with incremental models, dbt adds new columns to the target table but INSERT statements used `SELECT *` with positional matching. This causes values to be inserted into the wrong columns when column order changes, resulting in data corruption.

## Solution  
Modified all INSERT statements in incremental materializations to use Databricks `INSERT BY NAME` syntax, which matches columns by name instead of position.

## Changes
- Modified `insert_into_sql_impl()` to use `INSERT INTO ... BY NAME`
- Modified `get_insert_overwrite_sql()` to use `INSERT OVERWRITE ... BY NAME`
- Modified `get_replace_where_sql()` to use `INSERT INTO ... BY NAME REPLACE WHERE`
- Modified `create_table_at()` to use `INSERT INTO ... BY NAME`
- Updated unit tests to verify BY NAME syntax
- Added functional tests reproducing and validating fix for column order bug

## Breaking Change
**Requires Databricks Runtime 12.2 LTS or higher.** Users on older runtimes should pin to dbt-databricks 1.10.x.

## Testing
- ✅ Unit tests pass demonstrating correct SQL generation
- ✅ Functional tests validate column order changes work correctly
- ✅ All existing tests continue to pass (updated to reflect new behavior)
- ✅ Linter passes
- Tested across all three strategies: `append`, `insert_overwrite`, `replace_where`
- Applies to `merge` strategy via table creation

Co-Authored-By: Claude <noreply@anthropic.com>


.cursor/mcp.json
```diff
@@ -0,0 +1,9 @@
+{
+  "mcpServers": {
```
benc-db: add .cursor to .gitignore please

randypitcherii: smart - my bad. First batch ignored it and it must have snuck back in. 

Want me to add a github action to auto check for this and tell people to fix it as an apology for wasting your time on this?

randypitcherii: or shit, I'll do you one better and just PR an inclusion in the gitignore and that should fix it hahaha

PR #1218, opened 2025-10-15
refactor: Restructure dbr feature gating

benc-db:
### Description

Refactoring to use a comprehensive solution for specifying DBR-gated features.  Should make it easier/more obvious how to add such features in the future.

### Checklist


AGENTS.md
```diff
@@ -160,6 +161,44 @@ DatabricksAdapter (impl.py)
...
+- **Key Features**:
+  - Per-compute caching (different clusters can have different capabilities)
+  - Named capabilities instead of magic version numbers
+  - Automatic detection of DBR version and SQL warehouse environments
+- **Supported Capabilities**:
+  - `TIMESTAMPDIFF` (DBR 10.4+): Advanced date/time functions
+  - `INSERT_BY_NAME` (DBR 12.2+): Name-based column matching in INSERT
+  - `ICEBERG` (DBR 14.3+): Apache Iceberg table format
+  - `COMMENT_ON_COLUMN` (DBR 16.1+): Modern column comment syntax
+  - `JSON_COLUMN_METADATA` (DBR 16.2+): Efficient metadata retrieval
```
ericj-db, line 176: Is this meant to be a comprehensive list? I think it's missing `STREAMING_TABLE_JSON_METADATA` from your implementation

dbt/adapters/databricks/dbr_capabilities.py
```diff
@@ -0,0 +1,169 @@
...
+class DBRCapabilities:
+    """
+    Manages DBR version capabilities for a specific compute resource.
+
+    This class is constructed with compute information and provides
+    capability checks based on that compute's characteristics.
+    """
+
+    # Define all capability specifications
+    CAPABILITY_SPECS = {
```
ericj-db, line 41: `REPLACE ON` for insert overwrite is only available 17.1+. Should we add that here as well?

dbt/adapters/databricks/dbr_capabilities.py
```diff
@@ -0,0 +1,169 @@
...
+            min_version=(12, 2),  # DBR 12.2 LTS
+        ),
+    }
+
+    def __init__(
+        self,
+        dbr_version: Optional[tuple[int, int]] = None,
+        is_sql_warehouse: bool = False,
+        is_unity_catalog: bool = False,
+        capability_overrides: Optional[dict[DBRCapability, bool]] = None,
```
ericj-db: What's the use case for this override? I don't think I see it being used anywhere other than tests

benc-db: good point;  I'm pretty sure it's only test support, but we can implement that a different way.

AGENTS.md
benc-db: good catch.

dbt/adapters/databricks/dbr_capabilities.py
benc-db: yes, will add.

AGENTS.md
benc-db: Eh, not essential to always update this.

PR #1221, opened 2025-10-20
feat: add delete+insert incremental strategy

### LINKED ISSUE #1217: Support for delete+insert incremental strategy
### Describe the feature
Adapter support for an incremental strategy that is able to handle non-unique matching keys when replacing incremental data (as described in the specs for the built-in [delete+insert](https://docs.getdbt.com/docs/build/incremental-strategy#deleteinsert) strategy). A potential solution could be based around Databricks' [REPLACE ON](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-dml-insert-into#replace-on) syntax, that replaces rows based on a boolean expression:

> If table_name is a Delta Lake table, REPLACE ON atomically deletes any rows in the table that match the query rows based on the specified [boolean_expression](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-expression), then inserts the query rows. Unlike REPLACE WHERE’s [boolean_expression](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-expression), REPLACE ON’s [boolean_expression](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-expression) can reference columns from both the table and the query. If the query is empty, no deletions occur because there are no query rows to match.

The compilation code could be similar to `replace_where` strategy, compiling the following model:
```sql
{{ config(
    materialized='incremental',
    file_format='delta',
    incremental_strategy='delete+insert',
    unique_key=('date_day')
) }}

with new_events as (

    select * from {{ ref('events') }}

    {% if is_incremental() %}
    where date_day >= date_add(current_date, -1)
    {% endif %}

)

select
    date_day,
    count(*) as users

from new_events
group by 1
```

into the following sql run code:
```sql
create temporary view databricks_incremental__dbt_tmp as

    with new_events as (

        select * from analytics.events


        where date_day >= date_add(current_date, -1)


    )

    select
        date_day,
        count(*) as users

    from events
    group by 1

;

insert into table analytics.databricks_incremental as t
replace on (t.date_day = s.date_day)
select `date_day`, `users` from databricks_incremental__dbt_tmp as s
```

The alternative incremental strategies all have certain downsides for this use-case. `insert_overwrite` does seem to require partitioned tables, and also partitioning granularity that matches the added increment. `merge` strategy doesn't work with non-unique `unique_key`s. And `replace_where` would require us to be explicit about our increment ranges via `incremental_predicates` (which seems not easy to work with for this case).

### Who will this benefit?
Anyone that would like to build incremental models based on a non-unique "matching key" (or `unique_key` in the case of `delete+insert`) for non-partitioned tables (e.g. liquid clustered tables).

I could also contribute this feature

canbekley:
Resolves #1217 

### Description

this pr implements the dbt built-in incremental strategy [delete+insert](https://docs.getdbt.com/docs/build/incremental-strategy#deleteinsert), allowing users to build incremental models based on a (potentially non-unique) `unique_key`. it does so by utilizing databricks' [INSERT INTO .. REPLACE ON](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-syntax-dml-insert-into#replace-on) semantics.

### Checklist

### DIFF
--- a/tests/unit/macros/materializations/incremental/test_delete_insert.py
+++ b/tests/unit/macros/materializations/incremental/test_delete_insert.py
@@ -57,7 +57,9 @@ class TestDeleteInsertMacros(MacroTestBase):
         self.assert_sql_equal(sql, expected)
 
     def test_delete_insert_sql_impl__multiple_incremental_predicates(self, template):
-        sql = self.render_delete_insert(template, unique_key="a", incremental_predicates=["a > 1", "b < 3"])
+        sql = self.render_delete_insert(
+            template, unique_key="a", incremental_predicates=["a > 1", "b < 3"]
+        )
         expected = """
             insert into table target as target
             replace on (target.a <=> temp.a)


dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -58,6 +62,38 @@ REPLACE WHERE {{ predicates }}
...
+  {%- set target_relation = arg_dict.get('target_relation') -%}
+  {%- set incremental_predicates = config.get('incremental_predicates') -%}
+  {%- set target_columns = (adapter.get_columns_in_relation(target_relation) | map(attribute='quoted') | list) -%}
+  {%- set unique_key = config.require('unique_key') -%}
+  {%- set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change'), default='ignore') -%}
+  {{ delete_insert_sql_impl(source_relation, target_relation, target_columns, unique_key, on_schema_change, incremental_predicates) }}
+{% endmacro %}
+
+{% macro delete_insert_sql_impl(source_relation, target_relation, target_columns, unique_key, on_schema_change, incremental_predicates) %}
+  {%- set target_cols_csv = '*' if on_schema_change == 'ignore' else (target_columns | join(', ')) -%}
```
canbekley: I'm actually not sure if we ever want to `select *` from the source/temporary relation, and now that I think about it, it could be even worse to use with `on_schema_change == 'ignore'`. in general I see two problems with a `select *`:
- the order of the selected columns matters, as databricks might try to insert data into the wrong columns if source and target column orders don't exactly match. for instance, if the column order of a dbt sql model was changed, and the delta target table was not updated, we could end up with a databricks failure or even a silent data corruption.
- if there is any sort of mismatch between source and target schemas (for instance with `on_schema_change == 'ignore'`), we could end up with additional, missing or changed columns in the source relation. `insert into .. replace on` seems to only work with exact schema matches between source and target.

so I think it might be better to just use explicit column selection based on the target schema. if a model had `on_schema_change == 'ignore'` and a column was removed from the model's sql, it would then lead to some "missing column in selection" error (which should be acceptable).

@benc-db I will push and update here. but I would also appreciate your feedback on this.

benc-db: Yeah, we recently had a reckoning with select *.  We needed to switch to using 'by name' to handle the order mismatch.  And unfortunately this doesn't work with replace on to my knowledge.  Please take a look at how we are using 'replace on' in the 1.11.latest branch.

benc-db:
Closing in favor of a branch on origin.  Thanks for the contribution, will merge shortly.

PR #1227, opened 2025-10-21
chore: Update dbt dependencies version ranges

zqureshi:
Resolves #

### Description

Bumps the dbt dependencies to make sure this works with dbt 1.10.13

### Checklist


emmyoop (commented):
Verified against PyPI that these pins are the correct ranges now.

Issue #1235, opened 2025-10-27
Streaming model creation fails with incompatible schema error


adam-staros95:
### Describe the bug
Streaming model creation fails with incompatible schema error when `--full-refresh` flag is provided to `dbt run` command. Problem appears when underlying parquet files have different columns across s3 paths.

### Steps To Reproduce
Create dbt model as follows:

```sql
-- users.sql
{% set env = get_databricks_env_name() %}
{% set s3_path = var(env ~ '_s3_path') %}
{% set model_name = 'Users' %}
{{
    config(
        catalog='bronze_'~env,
        alias=model_name|lower,
        materialized='streaming_table'
    )
}}

select *
from
    stream read_files(
        '{{ s3_path }}{{ get_model_prefix_per_env(env, model_name) }}', format => 'parquet'
    )
```
And execute `dbt run --select users --full-refresh`

### Expected behavior
Streaming table should be created and schema should be inferred correctly

### Screenshots and log output

```
Database Error in model users (models/bronze/users.sql)
  Table 'users' has a user-specified schema that is incompatible with the schema
   inferred from its query.
  "
  Streaming tables are stateful and remember data that has already been
  processed. If you want to recompute the table from scratch, please full refresh
  the table.
                
  
  Declared schema:
  root
   |-- columns list
  
  
  Inferred schema:
  root
   |-- columns list with extra columns not listed in `Declared schema`
  compiled code at target/run/<project_name>/models/bronze/users.sql
```

### System information
**The output of `dbt --version`:**
```
Core:
  - installed: 1.10.5 
  - latest:    1.10.13 - Update available!

  Your version of dbt-core is out of date!
  You can find instructions for upgrading here:
  https://docs.getdbt.com/docs/installation

Plugins:
  - databricks: 1.10.14 - Up to date!
  - redshift:   1.8.1   - Update available!
  - postgres:   1.8.2   - Update available!
  - spark:      1.9.3   - Up to date!
```

**The operating system you're using:**
macOS Sequoia

**The output of `python --version`:**
Python 3.9.6

PR #1236, opened 2025-10-27
fix(streaming_table): resolve schema inference error

adam-staros95:
Resolves ([1235](https://github.com/databricks/dbt-databricks/issues/1235))

### Description

Removing the limit in `create_temp_view` fixes schema inference issues when the underlying file structures are inconsistent between files. Based on my tests, removing the limit does not negatively affect table creation performance.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
   - Improved efficiency of column tags and comments change detection to use case-insensitive comparison
 - Use backtick quoting for everything to avoid errors with special characters ([1186](https://github.com/databricks/dbt-databricks/pull/1186))
 - Ensure column compare always uses lower case names (since Databricks stores internally as lower case) ([1190](https://github.com/databricks/dbt-databricks/pull/1190))
+- Fix incompatible schema error during streaming table creation ([1235](https://github.com/databricks/dbt-databricks/issues/1235))
 
 ### Under the Hood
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## dbt-databricks 1.10.15 (TBD)
+### Fixes
+
```
benc-db, line 14: Please rebase off 1.11.latest and update the changelog...while this is a fix, and I do want to take it, on the off chance there is perf impact, I'd like it on a separate feature release.  Thanks!

adam-staros95: Done. Regarding the performance impact — I tested my change on a Serverless SQL Warehouse and, surprisingly, I noticed better performance after removing the `LIMIT 10` from the `create or replace temporary view ` statement.
I repeated the test several times, thinking it might be a coincidence, but the results were quite consistent. Is it possible?

I actually expected no performance difference, since I believe both of these queries:

```sql
CREATE OR REPLACE TEMPORARY VIEW `users__dbt_tmp` AS
SELECT *
FROM read_files(
    's3://bucket_name/path_to_table/Users', format => 'parquet'
) LIMIT 10
```

and

```sql
CREATE OR REPLACE TEMPORARY VIEW `users__dbt_tmp` AS
SELECT *
FROM read_files(
    's3://bucket_name/path_to_table/Users', format => 'parquet'
)
```
should have similar performance, as I assume `read_files` scans all data in the S3 path before applying the LIMIT.

PR #1238, opened 2025-10-27
fix: Use CREATE OR REPLACE for managed iceberg tables

ericj-db:
Resolves #

### Description

Instead of using non-atomic DROP + CREATE, use `CREATE OR REPLACE` for managed Iceberg tables. This should help with potential replication lag race conditions that cause subsequent materializations to fail

### Checklist


dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -12,6 +12,8 @@
   {% set language = model['language'] %}
   {% set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change'), default='ignore') %}
   {% set is_delta = (catalog_relation.file_format == 'delta' and existing_relation.is_delta) %}
+  {% set is_iceberg = (catalog_relation.file_format == 'iceberg' and existing_relation.is_iceberg) %}
```
benc-db, line 15: Is this sufficient?  Our prior way of detecting iceberg was to look for table_format...is that incorporated somewhere else?

PR #1247, opened 2025-10-30
fix: delete+insert for old DBRs

benc-db:
### Description

With testing delete+insert, realized that the implementation only worked for DBR 17.1+; this PR gates based on DBR, and if the DBR is old, we provide a non-atomic mechanism that at least still accomplishes what is requested.

### Checklist


benc-db:
FYI, all our test clusters are new enough to not exercise this behavior, but I have manually validated on an older cluster.

PR #1256, opened 2025-11-11
fix: Do not set UniForm properties on managed iceberg

ericj-db:
Resolves #

### Description

There is a bug in `TblPropertiesProcessor` that is assuming that all tables with iceberg format need UniForm properties to be set. When managed iceberg behavior flag is enabled, this causes the materialization to fail

The fix is to only apply these properties when working with a UniForm table. Unfortunately, we cannot use the `file_format` to distinguish because we do not actually require the user to set `file_format` to `parquet` for managed iceberg. We infer it in the adapter based on the `table_format` and behavior flag. To work around this fact, I am setting `model.config.extra["_databricks_use_managed_iceberg"]` to pass the information to the config processor 

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -236,6 +236,9 @@ class DatabricksAdapter(SparkAdapter):
         self.add_catalog_integration(constants.DEFAULT_UNITY_CATALOG)
         self.add_catalog_integration(constants.DEFAULT_HIVE_METASTORE_CATALOG)
 
+        # Store the use_managed_iceberg flag in GlobalState for the session
+        GlobalState.set_use_managed_iceberg(bool(self.behavior.use_managed_iceberg))
+
     @property
     def _behavior_flags(self) -> list[BehaviorFlag]:
         return [
@@ -920,14 +923,6 @@ class DatabricksAdapter(SparkAdapter):
     @available.parse(lambda *a, **k: {})
     def get_config_from_model(self, model: RelationConfig) -> DatabricksRelationConfigBase:
         assert model.config, "Config was missing from relation"
-
-        # Inject the use_managed_iceberg behavior flag into the model config so
-        # processors can use it. This is needed by TblPropertiesProcessor to determine
-        # if we should add UniForm properties
-        if "_databricks_use_managed_iceberg" not in model.config.extra:
-            model.config.extra["_databricks_use_managed_iceberg"] = bool(
-                self.behavior.use_managed_iceberg
-            )
 
         if model.config.materialized == "materialized_view":
             return MaterializedViewAPI.get_from_relation_config(model)
--- a/dbt/adapters/databricks/global_state.py
+++ b/dbt/adapters/databricks/global_state.py
@@ -50,3 +50,15 @@ class GlobalState:
                 "DBT_DATABRICKS_CONNECTOR_LOG_LEVEL", "WARN"
             ).upper()
         return cls.__connector_log_level
+
+    __use_managed_iceberg: ClassVar[Optional[bool]] = None
+
+    @classmethod
+    def set_use_managed_iceberg(cls, value: bool) -> None:
+        """Set the use_managed_iceberg flag for the session."""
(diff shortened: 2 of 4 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -920,6 +920,15 @@ def get_relation_config(self, relation: DatabricksRelation) -> DatabricksRelatio
     @available.parse(lambda *a, **k: {})
     def get_config_from_model(self, model: RelationConfig) -> DatabricksRelationConfigBase:
         assert model.config, "Config was missing from relation"
+
+        # Inject the use_managed_iceberg behavior flag into the model config so
+        # processors can use it. This is needed by TblPropertiesProcessor to determine
+        # if we should add UniForm properties
+        if "_databricks_use_managed_iceberg" not in model.config.extra:
+            model.config.extra["_databricks_use_managed_iceberg"] = bool(
```
benc-db: this should be a constant if you're going to use this key multiple places.  I also wonder if we should just store this in GlobalState since it will be true for the entire session, and then we won't need to stuff it on the model config.

dbt/adapters/databricks/impl.py
```diff
@@ -236,6 +236,9 @@ def __init__(self, config: Any, mp_context: SpawnContext) -> None:
         self.add_catalog_integration(constants.DEFAULT_UNITY_CATALOG)
         self.add_catalog_integration(constants.DEFAULT_HIVE_METASTORE_CATALOG)
 
+        # Store the use_managed_iceberg flag in GlobalState for the session
+        GlobalState.set_use_managed_iceberg(bool(self.behavior.use_managed_iceberg))
```
benc-db, line 240: I think this might not be defined yet.  I'm looking above and seeing that I have left over comments but missing code (because I removed the behavior I added the guard for).  Anyway, I don't think we can assume that the flag is defined yet because dbt-core is nonsense, and they call this sometimes without the flags in the case of dbt debug.  Can you verify the behavior if you call dbt debug?

ericj-db: I tried running `dbt debug` against this change with the flag both disabled/enabled. In both cases, it ran fine. I temporarily bumped `dbt-databricks` version to 1.12.0 locally to make sure I was testing against the right package

```
eric.jang@Y5QG6GXNW1 dbt-query-tag-demo % dbt debug                   
17:49:00  Running with dbt=1.11.0-b4
17:49:00  dbt version: 1.11.0-b4
17:49:00  python version: 3.10.16
17:49:00  python path: /Users/eric.jang/miniconda/bin/python
17:49:00  os info: macOS-15.7.1-arm64-arm-64bit
17:49:01  Using profiles dir at /Users/eric.jang/dbt-query-tag-demo/profiles
17:49:01  Using profiles.yml file at /Users/eric.jang/dbt-query-tag-demo/profiles/profiles.yml
17:49:01  Using dbt_project.yml file at /Users/eric.jang/dbt-query-tag-demo/dbt_project.yml
17:49:01  adapter type: databricks
17:49:01  adapter version: 1.12.0
17:49:01  Configuration:
17:49:01    profiles.yml file [OK found and valid]
17:49:01    dbt_project.yml file [OK found and valid]
17:49:01  Required dependencies:
17:49:01   - git [OK found]

17:49:01  Connection:
17:49:01    host: https://e2-dogfood.staging.cloud.databricks.com
17:49:01    http_path: /sql/1.0/warehouses/e9ef6a527497a91d
17:49:01    catalog: eric_jang
17:49:01    schema: recommendation
17:49:01  Registered adapter: databricks=1.12.0
17:49:01  [WARNING]: Use managed Iceberg tables when table_format is iceberg. When this flag is disabled, UniForm is used instead.
You may opt into the new behavior sooner by setting `flags.use_managed_iceberg` to `True` in `dbt_project.yml`.
Visit https://docs.getdbt.com/reference/global-configs/behavior-changes for more information.
17:49:02    Connection test: [OK connection ok]

17:49:02  All checks passed!
```

PR #1258, opened 2025-11-12
feat: Bump dbt-core upper bound for 1.10.latest, and target CI dbt deps branches with python 3.9 support

QMalcolm:
### Description

- Bumps dbt-core upper bound to allow for `dbt-core 1.10.15`
- Changes the target branches of dbt deps in CI to target branches containing support for python 3.9

### Checklist

- [X] I have run this code in development and it appears to resolve the stated issue
- [X] This PR includes tests, or tests are not required/relevant for this PR
- [X] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section.

PR #1259, opened 2025-11-12
feat: Bump dbt-core upper bound for 1.10.latest, and target CI dbt deps branches with python 3.9 support

benc-db:
Migrated from https://github.com/databricks/dbt-databricks/pull/1258


benc-db:
> ✅ I would approve it if I had the power 😂

Yeah, I'm going to merge after functional tests run, assuming no new issues discovered.

PR #1273, opened 2025-11-25
Fix: escape and truncate default query tag values

ericj-db:
Resolves #

### Description

Fix to ensure that default (i.e. auto) query tag values are not invalid
- Escape special characters (FWIW, model names are more restrictive than query tag values but doesn't hurt to have extra layer of defense)
- Truncate values greater than 128 characters

Tested using a model with name `accepted_values_backhaulcontractschedule_site_category_3__Non_Served_Sites__New_Sites__Served_Sites__Non_HS_Sites__Served_Site__CHARACTERS_BEYOND_128`

Screenshot 2025-11-25 at 3 11 37 PM

### Checklist

### DIFF
--- a/tests/unit/test_query_tags.py
+++ b/tests/unit/test_query_tags.py
@@ -144,12 +144,18 @@ class TestQueryTagsUtils:
             "key1": "value:with:colons",
             "key2": "value,with,commas",
             "key3": r"value\with\backslashes",
+            "key4": r"path\to:file,v1",
+            "key5": r"a\b:c,d\e:f,g",
+            "key6": r"start\,middle:,end",
         }
         result = QueryTagsUtils.process_default_tags(tags)
 
         assert result["key1"] == r"value\:with\:colons"
         assert result["key2"] == r"value\,with\,commas"
         assert result["key3"] == r"value\\with\\backslashes"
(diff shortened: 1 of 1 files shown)


dbt/adapters/databricks/utils.py
```diff
@@ -181,6 +189,24 @@ def validate_query_tags(tags: dict[str, str], source: str = "") -> None:
...
 
+    @staticmethod
+    def process_default_tags(tags: dict[str, str]) -> dict[str, str]:
+        """Process default tags: escape special characters and truncate long values."""
+        processed = {}
+        for key, value in tags.items():
+            escaped_value = QueryTagsUtils.escape_tag_value(key, value, "Default tags")
+
+            if len(escaped_value) > 128:
+                logger.warning(
```
benc-db: I'm not sure we want to emit a warning, because this is not something the user can really take action on.  Maybe just debug?

benc-db (dismissed):
Approve with minor comment.

dbt/adapters/databricks/utils.py
ericj-db: Good point - downgraded it to debug level

PR #1294, opened 2026-01-06
feat: Implement row filter functionality

sd-db:
Introduces support for Unity Catalog row filters across table, incremental, materialized view, and streaming table materializations to enable fine-grained row-level security

[One Pager/PRD](https://drive.google.com/file/d/1gG1JeeEidcNGkuy9cScnbk0tWwwbLGps/view?usp=sharing) [Internal Doc]
[Implementation Plan](https://drive.google.com/file/d/1dBxOzqgEWa6PnBKZb_AGR-TPfV0dvFAq/view?usp=sharing) [Internal Doc]

### Testing

- Added functional tests to verify behaviour across the possible materialisation types [table/incremental/MV/ST] for CREATE/ALTER/DROP behaviour across V1/V2 materialisations
- Added sufficient unit tests, to check correctness of helper function/macros

### Checklist
- [x] Code has been tested in development
- [x] Relevant tests have been added
- [x] CHANGELOG.md has been updated with this feature

### DIFF
--- a/tests/unit/macros/relations/test_row_filter_macros.py
+++ b/tests/unit/macros/relations/test_row_filter_macros.py
@@ -383,8 +383,7 @@ class TestDropRowFilterIfExists(MacroTestBase):
         return ["macros/relations/components", "macros"]
 
     def test_skips_hive_metastore(self, template_bundle):
-        """Test that Hive Metastore relations are skipped entirely.
-        """
+        """Test that Hive Metastore relations are skipped entirely."""
         template_bundle.relation.is_hive_metastore = Mock(return_value=True)
 
         sql = self.run_macro(


benc-db:
Let's target the 1.12.latest branch since this is a significant feature.

dbt/adapters/databricks/impl.py
```diff
@@ -930,7 +930,7 @@ def get_config_from_model(self, model: RelationConfig) -> DatabricksRelationConf
             return MaterializedViewAPI.get_from_relation_config(model)
         elif model.config.materialized == "streaming_table":
             return StreamingTableAPI.get_from_relation_config(model)
-        elif model.config.materialized == "incremental":
+        elif model.config.materialized in ("incremental", "table"):
```
benc-db: Tables get completely replaced, so they do not need to processed with relation config approach.

dbt/include/databricks/macros/materializations/table.sql
```diff
@@ -24,11 +24,13 @@
     {% else %}
       {% if safe_create and existing_relation.can_be_renamed %}
         {{ safe_relation_replace(existing_relation, staging_relation, intermediate_relation, compiled_code) }}
+        {{ process_config_changes(target_relation) }}
```
benc-db: Incremental processing is not part of table materialization as a table materialization is complete replacement.  Instead, the row filters should be added at table create time.

dbt/include/databricks/macros/materializations/table.sql
```diff
@@ -24,11 +24,13 @@
...
       {% else %}
         {% if existing_relation and (existing_relation.type != 'table' or not (existing_relation.can_be_replaced and adapter.resolve_file_format(config) in ('delta', 'iceberg'))) -%}
           {{ adapter.drop_relation(existing_relation) }}
         {%- endif %}
         {{ create_table_at(target_relation, intermediate_relation, compiled_code) }}
+        {{ process_config_changes(target_relation) }}
```
benc-db: Same as above

dbt/include/databricks/macros/materializations/table.sql
```diff
@@ -62,6 +64,21 @@
     {% endif %}
     {%- do apply_tags(target_relation, tags) -%}
 
+    {#-- Handle row filter changes for V1 --#}
+    {%- if existing_relation and not target_relation.is_hive_metastore() -%}
```
benc-db: Same comment as above

dbt/include/databricks/macros/relations/components/row_filter.sql
```diff
@@ -0,0 +1,192 @@
...
+Implements row-level security via Unity Catalog ROW FILTER clause.
+
+ARCHITECTURAL NOTE: Why Both CREATE and ALTER Paths Exist
+---------------------------------------------------------
+This file contains intentional duplication between:
+- get_create_row_filter_clause() - reads raw config.get('row_filter'), validates in Jinja
+- alter_set_row_filter() - receives pre-validated RowFilterConfig from Python
+
+This duplication exists because:
+
```
benc-db: This comment block is confused.  Because row filters do not depend on column metadata, they should be implemented the same way for V1 and V2, i.e. specified as part of the core create clause.  For incremental changes (incremental table, MV, ST), they would be applied via alter.

dbt/include/databricks/macros/relations/components/row_filter.sql
```diff
@@ -0,0 +1,192 @@
...
+        {{ alter_set_row_filter(target_relation, row_filter) }}
+      {%- endcall -%}
+    {%- endif -%}
+  {%- endif -%}
+{%- endmacro -%}
+
+
+{#-- ===== CREATE CLAUSE MACROS ===== --#}
+
+{%- macro qualify_row_filter_function(function, relation) -%}
```
benc-db, line 138: We should probably not do this in a macro, rather in python.

dbt/include/databricks/macros/relations/table/create.sql
```diff
@@ -41,6 +41,7 @@
   {{ file_format_clause(catalog_relation) }}
   {{ databricks__options_clause(catalog_relation) }}
   {{ partition_cols(label="partitioned by") }}
+  {{ get_create_row_filter_clause(target_relation) }}
```
benc-db, line 44: Because these are added here, the changes made in the table materialization are strictly redundant.

dbt/include/databricks/macros/materializations/table.sql
sd-db: Actually there is one case where this was required, when a table had a `row-filter` earlier which we removed. Currently we don't handle that in CREATE but I can. IMO that would actually be much simpler and remove the need to do the updates here. I will proceed to do that instead...

dbt/include/databricks/macros/relations/table/create.sql
sd-db: acked+resolved in latest commit.

dbt/include/databricks/macros/relations/components/row_filter.sql
sd-db: Makes sense, fixed in the latest commit. This is a miss from my part, thx !

dbt/include/databricks/macros/materializations/table.sql
sd-db: I have made the changes locally to move everything for tables to CREATE flow. `Incremental processing` for tables is no longer required/used.

dbt/adapters/databricks/impl.py
sd-db: I have made the changes locally to move everything for tables to CREATE flow. Incremental processing for tables is no longer required/used.

dbt/include/databricks/macros/materializations/table.sql
sd-db: Fixed+Resolved in latest commit.

dbt/include/databricks/macros/materializations/table.sql
sd-db: Fixed+Resolved in latest commit

dbt/include/databricks/macros/relations/components/row_filter.sql
sd-db: We already do have this in Python as well for the `ALTER` flow, issue is with tables which require setting `row-filters` in the `CREATE` flow, which is why the duplication of code. I have updated the comment to note this more specifically as well. I don't think it is possible but if there is a cleaner approach let me know will explore...

benc-db: we can do pretty much anything in python by putting the method on the adapter, and then calling it from jinja with adapter.<method_name>.  You can keep this sort of stuff as jinja, as there is no existing consistency as far as what MUST be jinja or MUST be python, other than Jinja templates the final SQL.  Your call.

PR #1296, opened 2026-01-08
Fix: Retain multiple foreign-keys from one table to another after an incremental run

Ewan-Keith:
### Description

Encountered a bug when:
  * `"use_materialization_v2": False`
  * `config.contract.enforced: true`
  * Multiple foreign keys are specified from one table to another (e.g. a fact table with multiple date values, each of which has a foreign key constrain to a common data dimension)
  * An initial full run of the model with the FK constraints is carried out, followed by an incremental run.

The bug is that on the initial full run the multiple FK constraints to the common table are created correctly on the Databricks table, but then when an incremental run is carried out these FKs are removed from the Databricks table. Single FKs to other tables seem to be retained (this isn't replicated in the test case in the interest of keeping it minimal).

The bug is replicated in the integration test [TestIncrementalMultipleFKsToSameTableNoV2](https://github.com/databricks/dbt-databricks/compare/main...Ewan-Keith:dbt-databricks:consistenty-persist-databricks-constraints?expand=1#diff-64ad276353f7ed237d5072874d1fe18c67310f90493ce7dc8a9fc69756925335R380), before the fix is applied this passes the first assertion (after the initial full run) but fails the second assertion (after the incremental run) with an empty set of constraints returned. A matching test is also present replicating the logic when using materialization_v2 [TestIncrementalMultipleFKsToSameTable](https://github.com/databricks/dbt-databricks/compare/main...Ewan-Keith:dbt-databricks:consistenty-persist-databricks-constraints?expand=1#diff-64ad276353f7ed237d5072874d1fe18c67310f90493ce7dc8a9fc69756925335R334) which confirms the issue is not seen with v2 materialization (this passes without any actual code changes).

The fix I've landed on in `dbt/include/databricks/macros/materializations/incremental/incremental.sql` is to treat constraints as one of the configuration changes that is explicitly applied on an incremental run when changes are detected (the same as for `tags`, `tblproperties`, and `liquid_clustering`). This gets the failing test passing.

I don't think this is necessarily a root-cause fix, I think the root issue is likely a change in constraints being falsely detected in this instance, but I wasn't sure where to start on that, and this fix seems sensible (even if the root issue is resolved it seems likely we'd want changes to constraints to be applied without forcing a full-refresh where this is possible?). There's a unit test checking if there are any obvious bugs in the constraint diff checker but this doesn't turn anything up (passes as expected).

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@
 
 - Fix `hard_deletes: invalidate` incorrectly invalidating active records in snapshots (thanks @Zurbste!) ([#1281](https://github.com/databricks/dbt-databricks/issues/1281))
 - Fix serverless Python model environment configuration: use `environment_version` instead of deprecated `client` field. Users can now specify custom environment versions via `python_job_config.environments`. ([#1286](https://github.com/databricks/dbt-databricks/pull/1286))
-- Fix foreign-key on an incremental table to a primary key on a non-incremental table being lost after incremental run
 
 ### Under the Hood
 


dbt/include/databricks/macros/materializations/incremental/incremental.sql
```diff
@@ -186,6 +187,9 @@
         {% if liquid_clustering is not none %}
           {% do apply_liquid_clustered_cols(target_relation, liquid_clustering) %}
         {% endif %}
+        {% if constraints %}
```
benc-db: What i'm confused by is your description made it sound like constraints were being dropped, but we haven't removed or altered any logic that would cause them to drop.  I'm digging in, but if you could provide more context, like what SQL is showing up in the logs, that would be appreciated.

benc-db: By any chance, is the table with the PK not incremental?  If the PK table is not incremental, during create or replace it's PK will be destroyed and recreated, which causes the cascading delete of the FKs.  If the PK table were incremental, then no PKs or FKs should have been deleted.  So the fix here is basically saying, incremental tables, if your constraints don't match, whether due to deletion or due to the user changing it, reapply.

Ewan-Keith: yeah that's it! I was having trouble replicating the case so reduced our exact issue down in the integration test, thought it was the multiple keys that were the issue but total red herring, it's the inc -> non-inc like you say.

I've updated the integration tests to reflect this and replicated the behaviour (fails on v1 before we make the change to `incremental.sql`, passing after).

Ewan-Keith:
> @Ewan-Keith not sure if you can see [this](https://github.com/databricks/dbt-databricks/actions/runs/20837141631/job/59864039651), but we have a failure in the E2E tests:
> 
> ```
> 01:01:10  1 of 1 START sql incremental model test17679204186774943539_test_constraints.stg_numbers  [RUN]
> 01:01:10  [WARNING]: Use revamped materializations based on separating create and insert.  This allows more performant column comments, as well as new column features.
> You may opt into the new behavior sooner by setting `flags.use_materialization_v2` to `True` in `dbt_project.yml`.
> Visit https://docs.getdbt.com/reference/global-configs/behavior-changes for more information.
> 01:01:17  1 of 1 ERROR creating sql incremental model test17679204186774943539_test_constraints.stg_numbers  [ERROR in 6.88s]
> 01:01:17  
> 01:01:17  Finished running 1 incremental model in 0 hours 0 minutes and 12.18 seconds (12.18s).
> 01:01:17  
> 01:01:17  Completed with 1 error, 0 partial successes, and 0 warnings:
> 01:01:17  
> 01:01:17  Failure in model stg_numbers (models/stg_numbers.sql)
> 01:01:17    Database Error in model stg_numbers (models/stg_numbers.sql)
>   
>   [PARSE_SYNTAX_ERROR] Syntax error at or near '('. SQLSTATE: 42601 (line 2, pos 113)
>   
>   == SQL ==
>   /* {"app": "dbt", "dbt_version": "1.11.2", "dbt_databricks_version": "1.11.3", "databricks_sql_connector_version": "4.1.3", "profile_name": "test", "target_name": "default", "node_id": "model.test.stg_numbers"} */
>   ALTER TABLE `peco`.`test17679204186774943539_test_constraints`.`stg_numbers` ADD CONSTRAINT fk_n FOREIGN KEY (n) (n) REFERENCES test17679204186774943539_test_constraints.raw_numbers
> ```
> 
> In TestIncrementalForeignKeyExpressionConstraint::test_incremental_foreign_key_constraint
> 
> You also need to add annotations to your new tests to tell it to skip testing for "databricks_cluster" profile, since Hive doesn't support constraints.

Yeah I can see the failure, will try to find some time to understand why this change is causing the issue. Have added the test annotation, must've just gotten lost at some point in editing.

Ewan-Keith:
Think I tracked down the issue, the changes interacted badly with the code to render FK statement suffixes using the `expression` syntax (haven't used it myself so was a bit of a blindspot).

I think the re-application of the FK constraints on the incremental model actually raised a pre-existing issue where the python rendering of the FK sql differed from the jinja used for the initial creation. The failing functional test defines a FK constraint like:
```yaml
expression: (n) REFERENCES schema.raw_numbers
```

The original `_render_suffix()` always prepended `FOREIGN KEY (columns)` then appended the expression. This resulted in invalid sql:
```sql
FOREIGN KEY (n) (n) REFERENCES schema.raw_numbers
```

I fixed this in the `_render_suffix()` method but that caused one of the unit tests to fail. This was because expression can either specify the FK columns or not, the unit test that failed didn't specify columns, so:
```python
expression="REFERENCES other_table (other_id) DEFERRABLE"
```
became:

```sql
 FOREIGN KEY REFERENCES other_table (other_id) DEFERRABLE
```

instead of:

```sql
FOREIGN KEY (id, other) REFERENCES other_table (other_id) DEFERRABLE
```

Spelling all this out because while I think the new fix should resolve these issues (all unit tests and functional tests under `tests/functional/adapter/constraints/` passing locally) it's hardly an elegant solution.. Matching on `(` at the start of the string to see if columns are specified in the expression or not and handling appropriately.

benc-db:
Hi @Ewan-Keith we didn't make the cut off for the 1.11.4 release, can you update the changelog change to be part of the 1.11.5 release?  We'll plan to put that out next week with your fix.  Rerunning tests now.

Ewan-Keith:
no worries! that's the change log updated.

CHANGELOG.md
```diff
@@ -8,6 +14,7 @@
 
 - Fix `hard_deletes: invalidate` incorrectly invalidating active records in snapshots (thanks @Zurbste!) ([#1281](https://github.com/databricks/dbt-databricks/issues/1281))
 - Fix serverless Python model environment configuration: use `environment_version` instead of deprecated `client` field. Users can now specify custom environment versions via `python_job_config.environments`. ([#1286](https://github.com/databricks/dbt-databricks/pull/1286))
+- Fix foreign-key on an incremental table to a primary key on a non-incremental table being lost after incremental run
```
sd-db: nit: cleanup from 1.11.4 logs (already released)

Ewan-Keith: ah I thought I'd cleaned that one up! done now.

Ewan-Keith:
> @Ewan-Keith all tests are passing, are you ready for merge?

Absolutely! Whenever suits :smile:

PR #1319, opened 2026-02-03
fix: track tblproperties managed by dbt (#1314)

### LINKED ISSUE #1314: running materialized_view always results in a full rebuild
### Describe the bug
Re-running a materialized view always recreates the view, rather than triggering a refresh.

### Steps To Reproduce
1. Create a materialized_view
2. Run it
3. Run it again and watch it get recreated

### Expected behavior
After the first run, subsequent runs (assuming there are no changes to the model or config) will trigger a refresh of the view, not a rebuild.

### System information
```
Core:
  - installed: 1.11.2

Plugins:
```

Linux

Python 3.13.11

### Additional context
Digging around the code, this is being caused by the huge number of tblproperties that Databricks is automatically applying to the materialized view. Although there is a list of properties to ignore, it only covers a small subset of the properties that Databricks is applying. Trying to manage this via a hardcoded list of properties is going to be a losing battle, as they are constantly adding new ones.

polivbr:
Resolves #1314

### Description

Rather than relying on an ever changing list of tblproperty keys to ignore, track (in a tblproperty) which keys have been set in the model, and are therefore being managed by dbt.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Fix foreign-key on an incremental table to a primary key on a non-incremental table being lost after incremental run
+- Track tblproperties which are managed by dbt, rather than relying on ignore list
 
 ### Under the Hood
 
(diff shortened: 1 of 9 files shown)


sd-db:
Closing in favour of #1350. I have tested that E2E and did a few small fixes so that all looks good. Thanks again for the contributions @polivbr !!

PR #1321, opened 2026-02-05
feat: Add Notebook-scoped packages for command submits or notebook job run

### LINKED ISSUE #1320: Notebook-scoped libraries feature not enable in dbt-databricks
### Describe the bug

We are trying to use All-Purpose clusters for dbt Python models. The [PythonCommandSubmitter](https://github.com/databricks/dbt-databricks/blob/25caa2a14ed0535f08f6fd92e29b39df1f453e4d/dbt/adapters/databricks/python_models/python_submissions.py#L62-L89) class is currently agnostic of packages defined in the model config.

We were able to work around this by using the `upload_notebook` flag (internally handled by [PythonNotebookUploader](https://github.com/databricks/dbt-databricks/blob/25caa2a14ed0535f08f6fd92e29b39df1f453e4d/dbt/adapters/databricks/python_models/python_submissions.py#L92-L127)), which installs the packages at the cluster system-scope.

This represents a significant problem for us: as the project grows, different Python models will require different packages or versions, eventually leading to dependency collisions. While this is less of an issue in production (where we use ephemeral Job Clusters), it severely hinders the Development Experience (DX). We want to use shared All-Purpose clusters in development to speed up iteration without package conflicts.

Databricks provides [Notebook-Scoped libraries](https://docs.databricks.com/aws/en/libraries/notebooks-python-libraries) specifically to avoid this. Since dbt can already compile the code into a Databricks Notebook, the submitters should be updated to leverage this feature.

### Steps To Reproduce
By running this simple model you can test the issue:

```python
import pandas as pd
from rapidfuzz import fuzz

def get_fuzz_ration(a,b):
    return fuzz.ratio(a,b)

def model(dbt, session):
    """
    Simple test model to verify rapidfuzz works on the cluster.
    Should complete in under 5 seconds.
    """
    # Configuration - using same cluster and package
    dbt.config(
        materialized="table",
        cluster_id=<an_all_purpose_cluster_id>,
        packages=['rapidfuzz==3.13.0'],
        submission_method='all_purpose_cluster',
        create_notebook=True,
        timeout=600
    )    

    # Create a tiny test dataset (just 3 rows)
    test_data = pd.DataFrame({
        'id': [1, 2, 3],
        'text1': ['hello world', 'databricks test', 'rapidfuzz check'],
        'text2': ['hello world!', 'databrick test', 'rapid fuzz check']
    })

    # Compute simple fuzzy scores
    test_data['similarity_score'] = [
        get_fuzz_ration(a, b)
        for a, b in zip(test_data['text1'], test_data['text2'])
    ]

    return test_data
```
1. Toggle Notesbook to False using a "vanilla" cluster: The model fails because the package is not found.

2. Set Notesbook to True: The model works, but the package is installed at the cluster system level (affecting all other users and models).
 
### Expected behavior
The compiled code should include a `%pip install <package>` command at the beginning of the notebook/script.

Ideally, we could introduce a flag like `use_notebook_scoped_libraries: true`. This would ensure that:

1. Packages are isolated to the specific dbt run.

2. The solution works across all compute types (Job Clusters, Serverless, and All-Purpose) since `%pip` is the standard for modern Databricks runtimes.

3. The solution will work no matter the data security access (Shared or single user)

Running the model with `use_notebook=True` works, but the dependency persists in the cluster environment even after the job ends:

Image

Image

### System information
```
(dbt-data-model) ➜  dbt_databricks git:(test/python-model) ✗ uv run dbt --version
Core:
  - installed: 1.11.2

Plugins:
```

**The operating system you're using:** MacOS Tahoe Version 26.2

**The output of `python --version`:** Python 3.11.14

fedemgp:
Resolves #1320

### Description

Add feature to install packages in Notebook-scoped environment for PythonCommandSubmitt and PythonNotebookUploader classes

#### Execution Test

I made the changes and tested it in our environment, looking that the compiled code now have the prepended package installation

##### All purpose cluster

image

##### Serverless cluster

image

#### Job cluster

image

### Checklist


tejassp-db (commented):
Have you taken this into account - The [Databricks docs](https://docs.databricks.com/aws/en/libraries/notebooks-python-libraries#manage-libraries-with-pip-commands) state "Starting with Databricks Runtime 13.0 %pip commands do not automatically restart the Python process. If you install a new package or update an existing package, you may need to use dbutils.library.restartPython() to see the new packages."
- https://docs.databricks.com/aws/en/libraries/notebooks-python-libraries#manage-libraries-with-pip-commands
- https://docs.databricks.com/aws/en/libraries/restart-python-process


Also have you considered the [recommendation on driver node sizes for notebook scoped libraries](https://docs.databricks.com/aws/en/libraries/notebooks-python-libraries#how-large-should-the-driver-node-be-when-working-with-notebook-scoped-libraries)

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -63,21 +63,41 @@ class PythonCommandSubmitter(PythonSubmitter):
...
+
+        # Prepend the pip install command to the compiled code
+        return f"{pip_install_cmd}\n\n{compiled_code}"
 
     @override
     def submit(self, compiled_code: str) -> None:
         logger.debug("Submitting Python model using the Command API.")
 
+        # Prepare code with notebook-scoped package installation if needed
+        code_to_execute = self._prepare_code_with_packages(compiled_code)
```
sd-db: We should ensure we only call this path if `notebook_scoped_libraries` is set.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -103,6 +123,21 @@ def __init__(self, api_client: DatabricksApiClient, parsed_model: ParsedPythonMo
             else {}
         )
         self.notebook_access_control_list = parsed_model.config.notebook_access_control_list
+        self.packages = (
+            parsed_model.config.packages if parsed_model.config.notebook_scoped_libraries else []
+        )
+
+    def _prepare_code_with_packages(self, compiled_code: str) -> str:
```
sd-db: instead of duplication, better to extract into a re-usable function component. Maybe something like 

```
def prepare_code_with_packages(compiled_code: str, packages: list[str], separator: str = "\n\n") -> str:
```

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -252,16 +292,24 @@ def get_library_config(
...
 
-    for package in packages:
-        if index_url:
-            libraries.append({"pypi": {"package": package, "repo": index_url}})
-        else:
-            libraries.append({"pypi": {"package": package}})
+    # Only add packages to cluster-level libraries if not using notebook-scoped
+    if not notebook_scoped_libraries:
+        for package in packages:
+            if index_url:
```
sd-db, line 312: Let us look to add support for `index_url` in notebook scoped packages as well. Since this is supported today; it would be weird if some model silently breaks if something breaks if an user who has set this switches to using notebook scoped packages.

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -103,6 +123,21 @@ def __init__(self, api_client: DatabricksApiClient, parsed_model: ParsedPythonMo
...
+        """Prepend notebook-scoped package installation commands to the compiled code."""
+        if not self.packages:
+            return compiled_code
+
+        # Build the %pip install command for notebook-scoped packages
+        pip_install_cmd = "%pip install " + " ".join(self.packages)
+        logger.debug(f"Adding notebook-scoped package installation: {pip_install_cmd}")
+
+        # Prepend the pip install command to the compiled code
+        return f"{pip_install_cmd}\n\n# COMMAND ----------\n{compiled_code}"
```
sd-db: As recommended by @tejassp-db, let us add `dbutils.library.restartPython()` after the `pip_install_cmd`

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -103,6 +123,21 @@ def __init__(self, api_client: DatabricksApiClient, parsed_model: ParsedPythonMo
...
+            parsed_model.config.packages if parsed_model.config.notebook_scoped_libraries else []
+        )
+
+    def _prepare_code_with_packages(self, compiled_code: str) -> str:
+        """Prepend notebook-scoped package installation commands to the compiled code."""
+        if not self.packages:
+            return compiled_code
+
+        # Build the %pip install command for notebook-scoped packages
+        pip_install_cmd = "%pip install " + " ".join(self.packages)
```
sd-db: maybe can try `%pip install -q {packages}` to reduce noise in notebook output

sd-db (commented):
Thx for the PR. Took a look at the code changes(didn't review tests yet) and added comments.

fedemgp:
> Have you taken this into account - The [Databricks docs](https://docs.databricks.com/aws/en/libraries/notebooks-python-libraries#manage-libraries-with-pip-commands) state "Starting with Databricks Runtime 13.0 %pip commands do not automatically restart the Python process. If you install a new package or update an existing package, you may need to use dbutils.library.restartPython() to see the new packages."
> 
> * https://docs.databricks.com/aws/en/libraries/notebooks-python-libraries#manage-libraries-with-pip-commands
> * https://docs.databricks.com/aws/en/libraries/restart-python-process
> 
> Also have you considered the [recommendation on driver node sizes for notebook scoped libraries](https://docs.databricks.com/aws/en/libraries/notebooks-python-libraries#how-large-should-the-driver-node-be-when-working-with-notebook-scoped-libraries)

I didn't take into account the recommendation to restart python process, thanks for that. I will refactor this draft PR to make it more production ready (avoiding duplicated code) and consider that. 

About the how large should the driver node be, this solution is agnostic about cluster size. It will depend on the team that previously defined the cluster (no matter if it was an all purpose cluster, job cluster, whatever).

dbt/adapters/databricks/python_models/python_submissions.py
fedemgp: already added support to index-url also at notebook scoped packages [here](https://github.com/fedemgp/dbt-databricks/blob/add_notebook_scoped_packages/dbt/adapters/databricks/python_models/python_submissions.py#L51-L53)

dbt/adapters/databricks/python_models/python_submissions.py
fedemgp: now the private function checks if notebook_scoped_libraries is set before adding this header

dbt/adapters/databricks/python_models/python_submissions.py
fedemgp: added it into its base clase PythonSubmitter, i had to make some adjustments so any submitter is able to use uit

dbt/adapters/databricks/python_models/python_submissions.py
fedemgp: added

dbt/adapters/databricks/python_models/python_submissions.py
fedemgp: sure thing

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -335,11 +389,14 @@ def __init__(
         tracker: PythonRunTracker,
         uploader: PythonNotebookUploader,
         config_compiler: PythonJobConfigCompiler,
+        parsed_model: ParsedPythonModel,
     ) -> None:
         self.api_client = api_client
         self.tracker = tracker
         self.uploader = uploader
         self.config_compiler = config_compiler
+        self.parsed_model = parsed_model
```
sd-db: nit: I don't think we are using `self.parsed_model` anywhere

dbt/adapters/databricks/python_models/python_submissions.py
```diff
@@ -9,20 +9,55 @@
...
+            return compiled_code
+
+        index_url = (
+            f"--index-url {self.packages_config.index_url}"
+            if self.packages_config.index_url
+            else ""
+        )
+        # Build the %pip install command for notebook-scoped packages
+        packages = " ".join(self.packages_config.packages)
+        pip_install_cmd = f"%pip install {index_url} -q {packages}"
```
sd-db: nit: can replace with `f"%pip install {index_url} -q {packages}".replace("  ", " ")`. The current logic is leading to double-spaces when `{index_url}` is not provided. This behaviour is also encoded in the tests (_checking for double-spaces..._)

sd-db (commented):
I would recommend functional tests as well. I checked locally that this looks good. Would be good to add to the PR. Otherwise looks good outside of minor comments. Thx again for the PR !!

dbt/adapters/databricks/python_models/python_submissions.py
fedemgp: You are right, i will remove it

dbt/adapters/databricks/python_models/python_submissions.py
fedemgp: 👌

fedemgp:
Hello @tejassp-db, were you able to take a look at the latest changes?

sd-db:
Hi @fedemgp sorry for the late reply, I was on extended leave. Thanks for the functional test + fixes. I see the tests are failing though. The fix for this should be quite simple

- Both model fixtures in `fixtures.py` are missing `return df` — without it dbt can't materialize the result and the tests fail. Add `return df` at the end of both `notebook_scoped_packages_cmd_api_model` and `notebook_scoped_packages_notebook_run_model`
- `TestNotebookScopedPackagesCommandAPI` references the wrong fixture — it uses `notebook_scoped_packages_notebook_run_model` (which has create_notebook=True) instead of `notebook_scoped_packages_cmd_api_model` (which has create_notebook=False). This means both test classes exercise the same notebook-upload path and the Command API path has no coverage.

You can refer to the fixes [here](https://github.com/fedemgp/dbt-databricks/commit/8dc24e15adfebebf8f1925a41fcddd90b83dc0ad) (verified running against a live cluster).

sd-db:
> I see the tests are failing though

The failing [github action run](https://github.com/databricks/dbt-databricks/actions/runs/24386382873) should have more details on the failure reason

fedemgp:
> > I see the tests are failing though
> 
> The failing [github action run](https://github.com/databricks/dbt-databricks/actions/runs/24386382873) should have more details on the failure reason


You are totally right. I thought I had run the tests locally, but I guess I was wrong. I got confused because the PR skipped the step. I've fixed them now and ran them locally. My tests passed, but some unrelated ones failed. I think they might be failing due to a lack of configuration in my dev environment.

I pushed a new commit applying the fixes you recommended. Thanks!

sd-db (approved):
Changes look good, thanks for the PR.

fedemgp:
> Changes look good, thanks for the PR.

Glad to hear it, do you have an ETA of when the change will be released?

PR #1326, opened 2026-02-18
feat: Enable concurrent microbatch execution

### LINKED ISSUE #914: support concurrent_batches for microbatch
### Describe the feature
see [slack thread](https://getdbt.slack.com/archives/CNGCW8HKL/p1737643036733319)

setting `concurrent_batches=true` for `incremental_strategy=microbatch` yields
```
[WARNING]: Found 1 microbatch model with the `concurrent_batches` config set to true, but the databricks adapter does not support running batches concurrently. Batches will be run sequentially.
```

@benc-db followed the [dbt guide](https://github.com/dbt-labs/dbt-adapters/discussions/371#support-considerations) to activate concurrency, which yielded
```
[DELTA_CONCURRENT_APPEND] ConcurrentAppendException: Files were added to the root of the table by a concurrent update. Please try the operation again.
```
qouting him:

> There are ways to allow concurrent writes, but it looks like they may be dependent on the actual schema of the table: [link](https://learn.microsoft.com/en-us/azure/databricks/optimizations/isolation-level#concurrentappendexception)

considering that the whole 'microbatch' feature was developed especially with large datasets in mind, concurrency would be really powerful.

given an incremental batch of 5 days, we tried using jinja to loop through these 5 days and then unioning them (to not overload the memory with too many rows), but that got "optimized away" by the internal query optimizer. but it also just creates annoying complexity on top of your models.

### Additional context
none

### Who will this benefit?
clients with large datasets/heavy operations, where the 'row count' bottleneck does not scale linearly but explodes after some threshold (depending on memory)

yes, but no idea how for this topic

wmjones:
Resolves #914

### Description

Declares `MicrobatchConcurrency` adapter capability so dbt-core 1.9+ can execute microbatch incremental batches in parallel threads instead of sequentially.

Per reviewer feedback from @sd-db, the capability is **gated behind the `use_concurrent_microbatch` behavior flag** (default: `false`). Users opt in via:

```yaml
# dbt_project.yml
flags:
  use_concurrent_microbatch: true
```

When the flag is disabled (default), `adapter.supports(MicrobatchConcurrency)` returns `False` and dbt-core falls back to sequential batch execution — identical to current behavior.

### Implementation

1. **`USE_CONCURRENT_MICROBATCH` behavior flag** — module-level `BehaviorFlag` with `default=False`, registered in `_behavior_flags`
2. **`supports()` instance method override** — intercepts `Capability.MicrobatchConcurrency` and gates on the flag; delegates all other capabilities to `super().supports()`
3. **`MicrobatchConcurrency` removed from `_capabilities` dict** — the `supports()` override is the sole gatekeeper (if the capability stayed in `_capabilities`, `super().supports()` would return `True` regardless of the flag)
4. **Unit tests** — disabled-by-default, enabled-with-flag, guard test (capability not declared as Full in `_capabilities`), delegation regression test

### Integration test findings (Databricks cluster, `batch_size='day'`, 31 batches, `--threads 4`)

| Run | Flag | Batches OK | Notes |
|-----|------|-----------|-------|
| Sequential | `false` | 31/31 | Correct fallback, warning emitted |
| Concurrent | `true` | 13-30/31 | `DELTA_CONCURRENT_APPEND` on non-partitioned tables |

**Key finding: `REPLACE WHERE` predicates are non-overlapping** (each batch is exactly one `batch_size` wide, regardless of `lookback`), **but non-partitioned Delta tables still conflict** because Delta's `WriteSerializable` isolation cannot verify non-overlap at the file level — it conservatively rejects concurrent conditional overwrites to the same table root.

A secondary error class (`DELTA_METADATA_CHANGED`) occurs when dbt-databricks applies `SET TBLPROPERTIES` (e.g., `autoCompact`) per batch, conflicting with concurrent writes.

**Safe configurations for concurrent microbatch:**
- Partition the target table by `event_time` at the same granularity as `batch_size` (allows Delta to verify non-overlap at the partition level)
- Avoid per-batch `tblproperties` changes
- `DATABRICKS_SKIP_OPTIMIZE=true` prevents post-write OPTIMIZE conflicts (but does not prevent the core `REPLACE WHERE` conflict)

**`lookback` is NOT a factor** — it only controls which batches are generated (shifts the start of the batch list backwards), not the width of any individual batch's `REPLACE WHERE` predicate.

**The `default=false` is the correct safe choice.** The default can be flipped to `true` around 1.12 once partitioning guidance and/or retry logic is established.

**Prior art:** dbt-snowflake added the same capability in [dbt-snowflake#1259](https://github.com/dbt-labs/dbt-snowflake/pull/1259).

### Checklist

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -261,12 +261,7 @@ class DatabricksAdapter(SparkAdapter):
         ]
 
     def supports(self, capability: Capability) -> bool:
-        # Override as instance method to gate on behavior flag.
-        # Safe because all dbt-core call sites invoke supports() on an adapter
-        # instance, never on the class directly (verified across 6 call sites
-        # in dbt-core 1.11.2). If a future dbt-core version calls
-        # DatabricksAdapter.supports(cap) at the class level, this would need
-        # to become a different pattern.
+        # Gate MicrobatchConcurrency on the use_concurrent_microbatch behavior flag.
         if capability == Capability.MicrobatchConcurrency:
             return bool(self.behavior.use_concurrent_microbatch)
         return super().supports(capability)
--- a/tests/unit/test_adapter_capabilities.py
+++ b/tests/unit/test_adapter_capabilities.py
@@ -179,17 +179,3 @@ class TestAdapterCapabilities:
         """Test that MicrobatchConcurrency is enabled when behavior flag is on."""
         adapter.behavior.use_concurrent_microbatch = True
         assert adapter.supports(Capability.MicrobatchConcurrency)
-
-    def test_microbatch_concurrency_not_declared_in_capabilities(self):
-        """MicrobatchConcurrency must not be declared with Full support in _capabilities.
-
-        The supports() override is the sole gatekeeper. CapabilityDict is a defaultdict,
-        so missing keys auto-create with Support.Unknown (falsy). This test ensures no one
-        accidentally re-adds MicrobatchConcurrency as Full, which would bypass the flag gate.
-        """
-        assert not bool(DatabricksAdapter._capabilities[Capability.MicrobatchConcurrency])
-
-    def test_supports_delegates_other_capabilities(self, adapter):
-        """Non-MicrobatchConcurrency capabilities should still use _capabilities dict."""
-        assert adapter.supports(Capability.TableLastModifiedMetadata)
-        assert adapter.supports(Capability.SchemaMetadataByRelations)


dbt/adapters/databricks/impl.py
```diff
@@ -247,8 +257,20 @@ def _behavior_flags(self) -> list[BehaviorFlag]:
             USE_MATERIALIZATION_V2,
             USE_REPLACE_ON_FOR_INSERT_OVERWRITE,
             USE_MANAGED_ICEBERG,
+            USE_CONCURRENT_MICROBATCH,
         ]
 
+    def supports(self, capability: Capability) -> bool:
+        # Override as instance method to gate on behavior flag.
```
sd-db: nit: comment comes across as too verbose/captures a lot more implementation specifics than required. Better to remove beyond a one-liner at max

tests/unit/test_adapter_capabilities.py
```diff
@@ -171,6 +171,25 @@ def test_insert_by_name_required_version_string(self):
...
+    def test_microbatch_concurrency_disabled_by_default(self, adapter):
+        """Test that MicrobatchConcurrency is disabled by default (behavior flag off)."""
+        assert not adapter.supports(Capability.MicrobatchConcurrency)
+
+    def test_microbatch_concurrency_enabled_with_flag(self, adapter):
+        """Test that MicrobatchConcurrency is enabled when behavior flag is on."""
+        adapter.behavior.use_concurrent_microbatch = True
         assert adapter.supports(Capability.MicrobatchConcurrency)
+
+    def test_microbatch_concurrency_not_declared_in_capabilities(self):
```
sd-db: This test seems overly specific and not that helpful.

tests/unit/test_adapter_capabilities.py
```diff
@@ -171,6 +171,25 @@ def test_insert_by_name_required_version_string(self):
...
+        """MicrobatchConcurrency must not be declared with Full support in _capabilities.
+
+        The supports() override is the sole gatekeeper. CapabilityDict is a defaultdict,
+        so missing keys auto-create with Support.Unknown (falsy). This test ensures no one
+        accidentally re-adds MicrobatchConcurrency as Full, which would bypass the flag gate.
+        """
+        assert not bool(DatabricksAdapter._capabilities[Capability.MicrobatchConcurrency])
+
+    def test_supports_delegates_other_capabilities(self, adapter):
```
sd-db: Same comment as on `test_microbatch_concurrency_not_declared_in_capabilities`; test is overly specific...

wmjones:
@sd-db Thanks for the review! Pushed `d63ae152` addressing all three comments:

1. Trimmed the `supports()` comment to a one-liner
2. Removed `test_microbatch_concurrency_not_declared_in_capabilities`
3. Removed `test_supports_delegates_other_capabilities`

The two core tests (disabled-by-default + enabled-with-flag) remain.

PR #1332, opened 2026-02-24
fix: quote catalog names in SHOW SCHEMAS commands (#1325)

### LINKED ISSUE #1325: Catalog names with hyphens not properly quoted in SHOW SCHEMAS commands Github Actions
### Describe the bug
When using a Unity Catalog name containing hyphens (or other special characters that require quoting), dbt-databricks generates unquoted SQL in `SHOW SCHEMAS` commands, causing failures on Github Actions.

### Steps To Reproduce
1. Use a Unity Catalog with hyphens in the name (e.g., `data_engineering-uc-dev`)
2. Configure dbt profile with this catalog
3. Run `dbt test` or any command that checks for schema existence

### Expected behavior
Catalog identifiers should be automatically quoted with backticks:
SHOW SCHEMAS IN 'data_engineering-uc-dev' LIKE 'schema_name'

### Actual behavior
SHOW SCHEMAS IN data_engineering-uc-dev LIKE 'schema_name'

Encountered an error:
Database Error
  Database Error

[INVALID_IDENTIFIER] The unquoted identifier data_engineering-uc-dev is invalid 
and must be back quoted as: 'data_engineering-uc-dev'.
Unquoted identifiers can only contain ASCII letters ('a' - 'z', 'A' - 'Z'), digits ('0' - '9'), 
and underbar ('_'). Unquoted identifiers must also not start with a digit.
Different data sources and meta stores may impose additional restrictions on valid[INVALID_IDENTIFIER] The unquoted identifier data_engineering-uc-dev is invalid 
'data_engineering-uc-dev'.
Unquoted identifiers can only contain ASCII letters ('a' - 'z', 'A' - 'Z'), digits ('0' - '9'), 
and underbar ('_'). Unquoted identifiers must also not start with a digit.
Different data sources and meta stores may impose additional restrictions on valid

### System information

  - installed: 1.11.2
Plugins:

macos

Python 3.11.11

aarushisingh04:
resolves #1325

### Description

when a unity catalog name contains hyphens or other special characters (e.g., `data_engineering-uc-dev`), `SHOW SCHEMAS` commands fail with `INVALID_IDENTIFIER` because the catalog name is not quoted.

**fix:** added backtick quoting around `{{ database }}` in the `list_schemas_sql` and `check_schema_exists_sql` macros in `metadata.sql`.

**before:**
```sql
SHOW SCHEMAS IN data_engineering-uc-dev
```

**after:**
```sql
SHOW SCHEMAS IN `data_engineering-uc-dev`
```

added 4 unit tests covering normal, hyphenated, and null database names for both macros.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
-## dbt-databricks next
+## dbt-databricks 1.11.6 (TBD)
 
+### Features
+
+- Enable concurrent microbatch execution via `MicrobatchConcurrency` capability,
+  gated behind the `use_concurrent_microbatch` behavior flag (default: `false`).
+  Opt in via `flags: {use_concurrent_microbatch: true}` in `dbt_project.yml`
+  ([#914](https://github.com/databricks/dbt-databricks/issues/914))
+  
 ### Fixes
 
 - Fix catalog names with special characters (e.g., hyphens) not being quoted in `SHOW SCHEMAS` commands, causing `INVALID_IDENTIFIER` errors ([#1325](https://github.com/databricks/dbt-databricks/issues/1325))
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -141,6 +141,17 @@ USE_MANAGED_ICEBERG = BehaviorFlag(
     ),
 )  # type: ignore[typeddict-item]
 
+USE_CONCURRENT_MICROBATCH = BehaviorFlag(
+    name="use_concurrent_microbatch",
+    default=False,
+    description=(
+        "Enable concurrent execution of microbatch incremental batches. "
+        "When enabled, dbt will run microbatch batches in parallel threads "
+        "instead of sequentially. The microbatch strategy itself requires "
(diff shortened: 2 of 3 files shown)


dbt/include/databricks/macros/adapters/metadata.sql
```diff
@@ -33,7 +33,7 @@ SHOW TABLES IN {{ relation.render() }}
 
 {% macro list_schemas_sql(database) %}
 {% if database %}
-  SHOW SCHEMAS IN {{ database }}
+  SHOW SCHEMAS IN `{{ database }}`
```
sd-db: nit: can we use `adapter.quote()` instead of raw back-ticks "``". It is the preferred practice in the codebase and protects against double-quoting. I think without it might also be fine here but is recommended and a smallish change.

aarushisingh04: @sd-db thanks for the check, it's done.

sd-db (approved):
Thanks for the PR !!

aarushisingh04:
@mstfkmlbsbdk2023 happy to help 👍

PR #1336, opened 2026-02-26
feat(PECOBLR-1766): Python UDF Support

sd-db:
## Description

Users want to create Python UDFs using dbt's functions feature on Databricks, but the current dbt-databricks adapter restricts UDF creation to SQL only. This PR looks to add support for Python UDFs (of the [following form/format + more](https://learn.microsoft.com/en-us/azure/databricks/udf/unity-catalog))

```
# def + explicit call -- for more complex logic
def compute(value):
    return value * 2
return compute(value)

# Inline return -- parameters are available as local variables
return value * 2
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## dbt-databricks 1.12.0 (TBD)
 
-- Add support for metric views as a materialization ([1285](https://github.com/databricks/dbt-databricks/pull/1285))
-- Add support for row filters [[1294](https://github.com/databricks/dbt-databricks/pull/1294)]
-- Add support for Python UDFs ([1336](https://github.com/databricks/dbt-databricks/pull/1336))
+### Features
+
+- Add support for metric views as a materialization ([#1285](https://github.com/databricks/dbt-databricks/pull/1285))
+- Add support for row filters ([#1294](https://github.com/databricks/dbt-databricks/pull/1294))
+- Add support for Python UDFs ([#1336](https://github.com/databricks/dbt-databricks/pull/1336))
 
 ## dbt-databricks 1.11.5 (Feb 19, 2026)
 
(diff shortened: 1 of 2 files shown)


tests/functional/adapter/functions/test_udfs.py
```diff
@@ -1,5 +1,291 @@
...
+from dbt_common.events.event_catcher import EventCatcher
 
+DATABRICKS_PYTHON_UDF_BODY = """
+def price_for_xlarge(price):
+    return price * 2
 
+return price_for_xlarge(price)
+"""
+
+# Databricks-compatible Python UDF YAML (using Python 3.11)
```
benc-db: Minor style thing to consider: historically the big blocks of config/sql have been put into in a fixtures file.  It's not a requirement, just pointing it out.

tests/functional/adapter/functions/test_udfs.py
```diff
@@ -1,5 +1,291 @@
...
+        result = run_dbt(["build", "--debug"], callbacks=[sql_event_catcher.catch])
+
+        # Verify build succeeded
+        assert len(result.results) == 1
+        node_result = result.results[0]
+        assert node_result.status == RunStatus.Success
+        assert isinstance(node_result.node, FunctionNode)
+        assert node_result.node.name == "price_for_xlarge"
+
+        # Verify the generated SQL contains Databricks Python UDF syntax
```
benc-db: all of the stuff from 100 to 121 is somewhat extra if line 127 passes.  I think those level of assertions are better off in the unit tests of the macro

benc-db (approved):
Approve with minor test style comments

tests/functional/adapter/functions/test_udfs.py
sd-db: Makes sense, removed the redundancy.

tests/functional/adapter/functions/test_udfs.py
sd-db: Added a fixtures file.

PR #1339, opened 2026-02-27
feat: support key-only databricks_tags

### LINKED ISSUE #1270: How do I apply a valueless/key-only column tags?
[As we all know](https://docs.getdbt.com/reference/resource-configs/databricks-configs#configuring-columns), it is possible to use Databricks Column Tags in both schema.yml and as config. The documentation gives this example for apply the tag data_classification with the value public:

```yaml
models:
  - name: customers
    columns:
      - name: customer_id
        databricks_tags:
          data_classification: "public"
```

But how would I apply a valueless tag such as the [governed tag](https://docs.databricks.com/aws/en/database-objects/tags#governed-tags) class.phone_number?

I tried this
```yaml
models:
  - name: customers
    columns:
      - name: customer_id
        databricks_tags:
          class.phone_number
```

But it raised the error Runtime Error in model customers (models/customers.sql)
  databricks_tags must be a dictionary

I also tried [setting it to null](https://yaml.org/type/null.html)
```yaml
models:
  - name: customers
    columns:
      - name: customer_id
        databricks_tags:
          class.phone_number: ~
```
But it raises the error.
ErrorClass=INVALID_PARAMETER_VALUE.INVALID_PARAMETER_VALUE] Tag value None is not an allowed value for tag policy key class.name. Allowed values: []

Could we document this to avoid having to guess?

canbekley:
Resolves #1270

### Description

Provides a solution to add support for configuring key-only `databricks_tags` for tables and columns. Utilizes existing databricks sql patterns, which treat values of key-only patterns as empty strings (when queried from `information_schema.tags` and also when applied via sql: `ALTER TABLE ... SET TAGS ('key_only_tag' = '', ...)`)

### Checklist


dbt/adapters/databricks/relation_configs/column_tags.py
```diff
@@ -79,7 +79,7 @@ def from_relation_config(cls, relation_config: RelationConfig) -> ColumnTagsConf
             if databricks_tags:
                 if isinstance(databricks_tags, dict):
                     set_column_tags[col["name"]] = {
-                        str(k): str(v) for k, v in databricks_tags.items()
+                        str(k): str(v or "") for k, v in databricks_tags.items()
```
sd-db: This collapses all Python-falsy values to `""`, which is incorrect. Things like

- {priority: 0} → tag value "0"
- {enabled: false} → tag value "False"

silently become key-only tags which is a regression.
 
- {priority: 0} → tag value "" -- silently becomes key-only
- {enabled: false} → tag value "" -- silently becomes key-only

sd-db: We should change to something like `"" if v is None else str(v)` which is narrower and limits to exactly `None`, leaving out 0/False/etc

dbt/adapters/databricks/relation_configs/column_tags.py
```diff
@@ -55,7 +55,7 @@ def from_relation_results(cls, results: RelationResults) -> ColumnTagsConfig:
                 # row contains [column_name, tag_name, tag_value]
                 column_name = str(row[0])
                 tag_name = str(row[1])
-                tag_value = str(row[2])
+                tag_value = str(row[2] or "")
```
sd-db: Noted, below why this can be risky. In this case risk is lower as `information_schema` returns type STRINGs, but we should still look to fix this to follow the same pattern - `"" if row[2] is None else str(row[2])`

dbt/adapters/databricks/relation_configs/tags.py
```diff
@@ -33,7 +33,7 @@ def from_relation_results(cls, results: RelationResults) -> TagsConfig:
 
         if table:
             for row in table.rows:
-                tags[str(row[0])] = str(row[1])
+                tags[str(row[0])] = str(row[1] or "")
```
sd-db: Similar case as in `column_tags`

dbt/adapters/databricks/relation_configs/tags.py
```diff
@@ -43,7 +43,7 @@ def from_relation_config(cls, relation_config: RelationConfig) -> TagsConfig:
         if not tags:
             return TagsConfig(set_tags=dict())
         if isinstance(tags, dict):
-            tags = {str(k): str(v) for k, v in tags.items()}
+            tags = {str(k): str(v or "") for k, v in tags.items()}
```
sd-db: Similar case as in `column_tags`

dbt/include/databricks/macros/relations/tags.sql
```diff
@@ -30,8 +30,8 @@
 
 {% macro alter_set_tags(relation, tags) -%}
   ALTER {{ relation.type.render() }} {{ relation.render() }} SET TAGS (
-    {% for tag in tags -%}
-      '{{ tag }}' = '{{ tags[tag] }}' {%- if not loop.last %}, {% endif -%}
+    {% for key, value in tags.items() -%}
```
sd-db: This is a behavioural no-op, this is not required.

tests/functional/adapter/column_tags/fixtures.py
```diff
@@ -13,7 +13,7 @@
       - name: account_number
         databricks_tags:
           pii: "true"
-          sensitive: "true"
+          sensitive: ""
```
sd-db: We should not change existing fixtures but rather add new fixtures as this is in a way reducing coverage. Let us do what we did in the table tag fixtures and add a new value.

tests/functional/adapter/column_tags/fixtures.py
```diff
@@ -29,7 +29,7 @@
       - name: account_number
         databricks_tags:
           pii: "true"
-          sensitive: "true"
```
sd-db, line 32: Same as above

dbt/adapters/databricks/relation_configs/column_tags.py
sd-db: We should add an unit test for this case as well, i.e. for `{"databricks_tags": {"priority": 0, "enabled": False}`

dbt/adapters/databricks/relation_configs/tags.py
sd-db: Also good to add tests so that this behaviour baked in (_no regression for Falsy values_)

sd-db (changes requested):
Have added some comments on changes, apart from that would be good to have a functional test. We can specially test the scenario in #1270 as a test to verify.

sd-db:
Hi @canbekley — thanks a lot for driving this one, and apologies for the delay on our side. Quick heads-up: this is a quarterly priority for our team and I'm looking to land it this week. If you have bandwidth to continue, please let me know. Otherwise, I can drive the change to completion — I'll push the remaining fixes directly to your branch so your commits stay and you remain co-author on anything I add. Please let me know either way when you get a chance.

canbekley:
hi @sd-db , thank you for your thorough review! I can commit the suggested changes later today and let you know.

canbekley:
hi @sd-db , I applied your suggestions. can you check again please?

dbt/adapters/databricks/relation_configs/column_tags.py
canbekley: good catch! I applied your suggestion added regression tests

sd-db:
Pushed two small follow-ups directly to the branch (`68e4cb9a`):

- **Functional coverage for YAML `null`**: `tests/functional/adapter/column_tags/` fixtures now include a `null_value:` (bare-key) entry alongside the existing `key_only: ""`, so the exact input form reported in #1270 round-trips end-to-end.
- **CHANGELOG**: fixed the `settings` typo and added mention of `None` as an accepted input.

sd-db:
I have also rebased this to `1.12.latest` as this might be regarded as a behaviour change as well, hence merging into `1.12.latest` (next) is safer

sd-db (approved):
Changes look good, thx for the PR !

PR #1341, opened 2026-03-03
feature: add optimize() call to snapshot materialization (#1335)

### LINKED ISSUE #1335: Run the optimize on snapshots with liquid clustering
### Describe the feature
When building a model, the "OPTIMIZE" query is executed automatically after each run. I want same feature for snapshots.

Using posthooks.

### Additional context
Ideally I want to change the config of the snapshots to adjust the liquid clustering columns even for existing tables.
I also considering some kind of default (dbt_scd_id, dbt_valid_to and config.id)

### Who will this benefit?
Increased performance for work with snapshots.

No.

aarushisingh04:
partially resolves #1335

### Description

1. added `optimize()` call to the snapshot materialization, matching existing behavior in table and incremental materializations. when a snapshot is configured with `liquid_clustered_by`, `zorder`, or `auto_liquid_cluster`, the `OPTIMIZE` query will now execute automatically after each snapshot run.

2. the change is a single-line addition to `snapshot.sql`, placed after `persist_constraints` and before `post-hooks` (again, consistent with the pattern in `table.sql` and `incremental.sql`).

**note**: the original issue also mentioned default clustering columns and `ALTER TABLE` support for config changes. these are intentionally excluded from this PR to keep scope minimal and can be addressed in follow-up PRs if needed. i believe they need a design decision that should be discussed by/with maintainers first due to having a larger scope.

### Checklist

- [x] I have run this code in development and it appears to resolve the stated issue (**unit tests pass :** this change mirrors the identical optimize() call pattern already used in `table.sql` and `incremental.sql`.)

### DIFF
--- a/tests/functional/adapter/simple_snapshot/test_snapshot.py
+++ b/tests/functional/adapter/simple_snapshot/test_snapshot.py
@@ -33,7 +33,7 @@ class TestSnapshotCheck(BaseSnapshotCheck):
 
 
 @pytest.mark.skip_profile("databricks_uc_cluster", "databricks_cluster")
-class TestSnapshotLiquidClustering(BaseSimpleSnapshot):
+class TestSnapshotLiquidClustering(TestSnapshot):
     @pytest.fixture(scope="class")
     def project_config_update(self):
         return {"snapshots": {"+liquid_clustered_by": "id"}}


tests/functional/adapter/simple_snapshot/test_snapshot.py
```diff
@@ -32,6 +32,13 @@ class TestSnapshotCheck(BaseSnapshotCheck):
     pass
 
 
+@pytest.mark.skip_profile("databricks_uc_cluster", "databricks_cluster")
+class TestSnapshotLiquidClustering(BaseSimpleSnapshot):
```
sd-db: This test is failing, my initial comment was incorrect. We need to extend `TestSnapshot` instead of `BaseSimpleSnapshot`.Tested and confirmed this locally and is working. 

The reason is that `BaseSimpleSnapshot` has an `add_fact_column` method that uses varchar(200) default null, which is incompatible with Databricks. `TestSnapshot` class works around this by overriding `add_fact_column` to use string.

aarushisingh04: oh i wouldn't have caught that one, changed it accordingly now. thanks for the explanation 👍

PR #1350, opened 2026-03-10
fix: properly check for unapplied tblproperties configs in existing r…

### LINKED ISSUE #1314: running materialized_view always results in a full rebuild

canbekley:
Resolves #1314

### Description

Aquiring TblProperties changesets is highly error-prone and not scalable. The old code filtered out Databricks-internal properties (like delta.minReaderVersion, clusteringColumns, etc.) when reading from the relation from a static "ignore list" that needed to be updated continuously. With the new approach, `from_relation_results` now includes all properties from the relation. Since `get_diff` only looks at what config tblproperties are currently missing or incorrect, these extra databricks-internal properties in other are harmless, they'll never appear in the diff.

### Checklist


TiPPeX2:
Cool, 
Is there something that holds us back here :) ?
@sd-db ?

sd-db (approved):
Thanks a lot for the PR, I have done some changes to fix a bug that might have affected STs + added tests. Overall looks good !!

PR #1355, opened 2026-03-12
fix: capability detection for named compute (`databricks_compute`)

trouze:
Resolves #1354 

### Description

Fixes a timing bug where `table_format='iceberg'` (and other capability-gated features) fail with "iceberg requires DBR 14.3+" when the model targets a named compute via `databricks_compute` model config, even though the compute supports the required DBR version.

### Fix

One-line change: call `_cache_dbr_capabilities()` eagerly in `_create_fresh_connection()` before reading the cache. This ensures the DBR version is queried and cached at connection creation time rather than waiting for the lazy open(). The call is idempotent (guarded by if `http_path` not in cache), so for default compute with a warm cache it's a no-op.

### Key changes
- `dbt/adapters/databricks/connections.py`: Added `self._cache_dbr_capabilities(creds, conn.http_path)` before `conn.capabilities = self._get_capabilities_for_http_path(conn.http_path)` in `_create_fresh_connection()`
- `tests/unit/test_connection_manager.py`: Added 3 unit tests validating eager caching for named compute, fallback behavior when cache is empty, and idempotency for default compute with a warm cache

### Checklist

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -208,6 +208,23 @@ class DatabricksConnectionManager(SparkConnectionManager):
                 is_sql_warehouse=not is_cluster,
             )
 
+    @classmethod
+    def _try_cache_dbr_capabilities(cls, creds: DatabricksCredentials, http_path: str) -> None:
+        """Like _cache_dbr_capabilities, but only writes to the cache when the version query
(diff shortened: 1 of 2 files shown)


sd-db (approved):
Changes look good, thanks for the contribution !!

PR #1358, opened 2026-03-13
fix: implicit view schema changes with alter view approach

canbekley:
Resolves #1356

> [!NOTE]
> Stacked on #1444 (resolves #1357), per maintainer request to split this work. please merge that PR first

### Description

Views materialized with `view_update_via_alter` (v2 materialization) were skipping `ALTER VIEW AS` whenever dbt's query diff reported no change, so changes in underlying tables (e.g. a new column appearing in a `select *` source) never propagated to the view.

Fix: force `ALTER VIEW AS <query>` on every invocation by replacing `QueryProcessor` with a `ViewQueryProcessor` whose `get_diff` always returns self. Drop the no-op branch in the view materialization (always alter, never skip).

With #1357's reapply-after-`alter_query` block now in place, the existing `ColumnCommentsProcessor` is also removed from `ViewConfig` every alter-view run wipes column comments and reapplies them unconditionally, so a separate diff path is no longer needed.

### Checklist

### DIFF
--- a/dbt/include/databricks/macros/relations/view/alter.sql
+++ b/dbt/include/databricks/macros/relations/view/alter.sql
@@ -16,6 +16,12 @@
   {% endif %}
   {% if query %}
     {{ alter_query(target_relation, query.query) }}
+    {% if config.persist_column_docs() and model.columns %}
+      {#-- ALTER VIEW AS <query> wipes all column comments, so reapply them here. --#}
+      {%- set existing_columns = adapter.get_columns_in_relation(target_relation) -%}
+      {%- set columns_to_persist = adapter.get_persist_doc_columns(existing_columns, model.columns) -%}
+      {{ alter_column_comment(target_relation, columns_to_persist) }}
+    {% endif %}
   {% endif %}
   {% if column_comments %}
     {{ alter_column_comments(target_relation, column_comments.comments) }}
--- a/dbt/include/databricks/macros/relations/view/alter.sql
+++ b/dbt/include/databricks/macros/relations/view/alter.sql
@@ -23,7 +23,4 @@
       {{ alter_column_comment(target_relation, columns_to_persist) }}
     {% endif %}
   {% endif %}
-  {% if column_comments %}
-    {{ alter_column_comments(target_relation, column_comments.comments) }}
-  {% endif %}
 {% endmacro %}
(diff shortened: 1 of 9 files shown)


tejassp-db:
Thanks for the PR. Will look into it.

dbt/include/databricks/macros/relations/view/alter.sql
```diff
@@ -16,8 +16,11 @@
   {% endif %}
   {% if query %}
     {{ alter_query(target_relation, query.query) }}
-  {% endif %}
-  {% if column_comments %}
-    {{ alter_column_comments(target_relation, column_comments.comments) }}
+    {% if config.persist_column_docs() and model.columns %}
+      {#-- ALTER VIEW AS <query> will wipe all column comments in a column_list, so we need to reapply them here. --#}
```
tejassp-db: Can you explain this comment please?

tejassp-db (changes requested):
It looks like there are two different fixes - column comments and view refresh, in this PR. Can you please split it into two separate PRs?

dbt/include/databricks/macros/relations/view/alter.sql
canbekley: that's the issue described in #1357. basically every `ALTER VIEW AS` command that changes the view defintion will currently (as of dbr 18.1) drop all existing column comments, without adding them back. this issue already existed, but #1356 makes #1357 strictly worse, because we would run `ALTER VIEW AS` on every invocation.

canbekley:
> It looks like there are two different fixes - column comments and view refresh, in this PR. Can you please split it into two separate PRs?

good point, I should definitely split this in two separate PRs. there is going to be an order dependency though, as we would need to merge the column comment fix before the bahavioural changes that enforce `ALTER VIEW AS` invocations. I will create a new PR for the column comment fix.

canbekley:
hi @tejassp-db , this PRs is now split into two dependent PRs: column comment fix in #1444 (which would need to be merged first) and the view schema refresh fix in the PR. I also updated the descriptions accordingly. could you please have another look?

sd-db:
Hi @canbekley I have reviewed this and overall looks good, but since results in a behaviour change I would want to target `1.13.latest` here. The release for which I am planning near end of august, let me know if that works out ?

sd-db (approved):
Overall changes look good, added a few more tests as well. If all looks good I can rebase to `1.13.latest` and look to merge, wdyt @canbekley ?

canbekley:
hi @sd-db, thank you for picking this up and adding the tests. I think shipping this pr with `1.13.latest` is reasonable, since it substantially changes view materialization behaviour

sd-db:
>  I think shipping this pr with 1.13.latest is reasonable, since it substantially changes view materialization behaviour

Thanks let me look to port.

PR #1363, opened 2026-03-24
feat(PECOBLR-2328): bump databricks-sql-connector upper bound to <4.1.6

tejassp-db:
## Summary

- Bumps `databricks-sql-connector` upper bound from `<4.1.4` to `<4.1.6`, allowing users to install connector `4.1.5`
- Connector `4.1.5` introduced [`_respect_server_retry_after_header`](https://github.com/databricks/databricks-sql-python/pull/756), which users can now opt into via `connection_parameters` in `profiles.yml`

## Context

Customers using `use_materialization_v2: true` experience duplicate rows when the server returns HTTP 503 after already committing an INSERT. The connector blindly retries, causing data to be written twice. With `_respect_server_retry_after_header: true`, retries only occur when the server explicitly sends a `Retry-After` header, preventing duplicate writes from infrastructure-level 503s.

## Test plan

- [x] All 744 unit tests pass (`hatch run unit`)
- [x] E2E verification with a dummy dbt project against UC SQL endpoint:
  - `dbt debug` + `dbt run` succeed with default profile (connector default `False` applied)
  - `dbt debug` + `dbt run` succeed with `_respect_server_retry_after_header: true` explicitly set


pyproject.toml
```diff
@@ -24,7 +24,7 @@ classifiers = [
 dependencies = [
     "click>=8.2.0, <9.0.0",
     "databricks-sdk>=0.68.0, <0.78.0",
-    "databricks-sql-connector[pyarrow]>=4.1.1, <4.1.4",
+    "databricks-sql-connector[pyarrow]>=4.1.5, <4.2.0",
```
tejassp-db: @jprakash-db Is this acceptable - since we don't want it to move beyond 4.2.0?

sd-db: - I think we should have tighter pinning today <4.1.6
- Any reason to update the lower version as well to 4.1.5 ? We can do this here but not required and we never know if this might break some env. IMO let us not update the minimum supported version

sd-db (commented):
Let us also update the CHANGELOG.md to list that we are updating the `python-sql-connector` dependency

dbt/adapters/databricks/credentials.py
```diff
@@ -73,6 +73,7 @@ def __pre_deserialize__(cls, data: dict[Any, Any]) -> dict[Any, Any]:
         data.setdefault("connection_parameters", {})
         data["connection_parameters"].setdefault("_retry_stop_after_attempts_count", 30)
         data["connection_parameters"].setdefault("_retry_delay_max", 60)
+        data["connection_parameters"].setdefault("_retry_server_directed_only", False)
```
sd-db: There were comments on the original PR, because of which I had changed the name for the config to [`_respect_server_retry_after_header`](http://github.com/databricks/databricks-sql-python/pull/757/changes#diff-af9774cb6d4c19ac9e50a40c7a1b1adc09fac7a0932d1077c9da24ccfdb5c4a9R193)

sd-db (dismissed):

sd-db (approved):

Issue #1338, opened 2026-02-26
make microbatch strategy support insert-by-name


data-blade:
### Describe the feature
recent adapter additions brought `insert-by-name` behavior to some processes that do not natively support it, see [here](https://github.com/databricks/dbt-databricks/issues/1025) and [here](https://github.com/databricks/dbt-databricks/issues/1211). one place that is still lacking that behavior is the `microbatch` strategy. microbatch models still break upon column order change (unless the data types still match, then it will silently produce wrong data).

### Describe alternatives you've considered
add complex validation that detects and alerts when code schema and database schema do not align for microbatch models.

### Additional context
dealing with column order is so annoying that, for us, this is seen as a blocker for migrating `insert_overwrite` models to `microbatch`

### Who will this benefit?
clients that have large incremental models suitable for microbatch, who do not want to deal with the frequent issues that arise from column order mismatches.

### Are you interested in contributing this feature?
potentially ;)

PR #1364, opened 2026-03-26
fix: convert workflow job spec to SDK dataclasses before jobs.create() (#1338)

aarushisingh04:
resolves #1360

### description

`WorkflowJobApi.create()` passes plain dicts to `workspace_client.jobs.create(**converted_job_spec)`. the databricks sdk's `jobs.create()` iterates tasks calling `v.as_dict()` on each (confirmed via `JobsAPI.create` source inspection). plain dicts have no `.as_dict()` which caused the `'dict' object has no attribute 'as_dict'` error.

1. the fix uses `JobSettings.from_dict()` + `.as_shallow_dict()` to convert the plain dict job spec into proper sdk dataclasses before calling `jobs.create()`. 
2. `as_shallow_dict()` is the correct choice here : it unpacks top-level fields as kwargs (matching the `jobs.create()` signature) while keeping nested values as typed `Task` / `NotebookTask` / etc. dataclasses that satisfy `.as_dict()`. 
3. this is the same pattern already used by `update_job_settings()` in the same class.

**changes:**
- `dbt/adapters/databricks/api_client.py` - `WorkflowJobApi.create()` now deserializes the dict via `JobSettings.from_dict()` and passes `.as_shallow_dict()` to the sdk. inline comment documents the sdk internals that make this the correct choice.
- `tests/unit/api_client/test_workflow_job_api.py` - tests strengthened to assert `isinstance(task, Task)`, call `.as_dict()` on each task and verify output (directly proving the original `AttributeError` cannot recur), and add a new `test_create__invalid_job_spec_raises` covering the `JobSettings.from_dict()` error path.

### checklist

- [x] i have run this code in development and it appears to resolve the stated issue
- [x] this pr includes tests, or tests are not required/relevant for this pr
- [x] i have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section.

### DIFF
--- a/dbt/adapters/databricks/api_client.py
+++ b/dbt/adapters/databricks/api_client.py
@@ -801,7 +801,11 @@ class WorkflowJobApi:
         try:
             # Convert job_spec to be compatible with jobs.create
             converted_job_spec = self._convert_job_spec_for_create(job_spec)
-            create_response = self.workspace_client.jobs.create(**converted_job_spec)
+            # Convert plain dicts to proper SDK dataclass objects via JobSettings.
+            # The SDK's jobs.create() expects typed objects (e.g. Task, not dict)
+            # and calls .as_dict() on them internally, which fails on plain dicts.
+            job_settings = JobSettings.from_dict(converted_job_spec)
+            create_response = self.workspace_client.jobs.create(**job_settings.as_shallow_dict())
             job_id = str(create_response.job_id)
             logger.info(f"New workflow created with job id {job_id}")
             return job_id
--- a/dbt/adapters/databricks/api_client.py
+++ b/dbt/adapters/databricks/api_client.py
@@ -802,8 +802,12 @@ class WorkflowJobApi:
             # Convert job_spec to be compatible with jobs.create
             converted_job_spec = self._convert_job_spec_for_create(job_spec)
             # Convert plain dicts to proper SDK dataclass objects via JobSettings.
-            # The SDK's jobs.create() expects typed objects (e.g. Task, not dict)
-            # and calls .as_dict() on them internally, which fails on plain dicts.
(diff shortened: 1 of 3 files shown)


dbt/adapters/databricks/api_client.py
```diff
@@ -801,7 +801,15 @@ def create(self, job_spec: dict[str, Any]) -> str:
         try:
             # Convert job_spec to be compatible with jobs.create
             converted_job_spec = self._convert_job_spec_for_create(job_spec)
-            create_response = self.workspace_client.jobs.create(**converted_job_spec)
+            # Convert plain dicts to proper SDK dataclass objects via JobSettings.
```
sd-db: comment is too verbose, maybe remove/simplify ?

tests/unit/api_client/test_workflow_job_api.py
```diff
@@ -59,7 +59,15 @@ def test_create__success(self, api, workspace_client):
         result = api.create(job_spec)
 
         assert result == "456"
-        workspace_client.jobs.create.assert_called_once_with(**job_spec)
+        workspace_client.jobs.create.assert_called_once()
```
sd-db, line 62: I think we can remove the comments from the tests as well

sd-db (dismissed):

dbt/adapters/databricks/api_client.py
aarushisingh04: done, simplified

tests/unit/api_client/test_workflow_job_api.py
aarushisingh04: yeah done, removed. thanks!

sd-db (approved):
Changes look good, thanks for the PR !

PR #1387, opened 2026-04-10
perf: skip unnecessary metadata fetch calls for tags when not configured

tejassp-db:
**Summary**

- Skips fetch_tags and fetch_column_tags queries to information_schema during incremental and view materializations when the model has no tags configured
- Falls back to fetching metadata when model config is unavailable (safe default)
- No behavior change for models with tags configured — diffs still computed correctly

**What changed**

- Added requires_server_metadata_for_diff() to TagsConfig and ColumnTagsConfig to signal when server metadata  is needed
- IncrementalTableAPI._describe_relation and ViewAPI._describe_relation now check model config before making tag fetch calls
- get_relation_config accepts the model config to pass through to _describe_relation
- Fixed duplicate TagsProcessor in StreamingTableConfig.config_components

**Test plan**

- Unit tests: empty tags (skip), non-empty tags (fetch), null config fallback (fetch), both tags + column tags present, hive_metastore (skip)
- Functional tests: override fetch_tags/fetch_column_tags macros to raise errors, confirming calls are actually skipped when expected

  PECOBLR-2497


dbt/adapters/databricks/impl.py
```diff
@@ -964,15 +967,17 @@ def parse_columns_and_constraints(
         return enriched_columns, parsed_constraints
 
     @available.parse(lambda *a, **k: {})
-    def get_relation_config(self, relation: DatabricksRelation) -> DatabricksRelationConfigBase:
+    def get_relation_config(
+        self, relation: DatabricksRelation, relation_config: DatabricksRelationConfigBase
```
sd-db: I think the naming here is very confusing for the new parameter, further compunded by the fact that the return type is also the same. What we should be always distinguishing is that we are passing `desired_config` as a new parameter and we expect to get `existing_config` as return. In this function we should rename the parameter from `relation_config` to `desired_config` to be less confusing

dbt/adapters/databricks/impl.py
```diff
@@ -1053,12 +1058,15 @@ def config_type(cls) -> type[DatabricksRelationConfig]:
 
     @classmethod
     def get_from_relation(
-        cls, adapter: DatabricksAdapter, relation: DatabricksRelation
+        cls,
+        adapter: DatabricksAdapter,
+        relation: DatabricksRelation,
+        relation_config: DatabricksRelationConfigBase,
```
sd-db: As above rename to `desired_config`

dbt/adapters/databricks/impl.py
```diff
@@ -1070,7 +1078,10 @@ def get_from_relation_config(cls, relation_config: RelationConfig) -> Databricks
     @classmethod
     @abstractmethod
     def _describe_relation(
-        cls, adapter: DatabricksAdapter, relation: DatabricksRelation
+        cls,
+        adapter: DatabricksAdapter,
+        relation: DatabricksRelation,
+        relation_config: DatabricksRelationConfigBase,
```
sd-db: rename parameter to `desired_config`

dbt/adapters/databricks/impl.py
```diff
@@ -1080,19 +1091,22 @@ def _describe_relation(
...
     ) -> DatabricksRelationConfig:
         """Get the relation config from the relation."""
 
-        relation_config = super().get_from_relation(adapter, relation)
+        full_relation_config = super().get_from_relation(adapter, relation, relation_config)
```
sd-db: `existing_config = super().get_from_relation(adapter, relation, relation_config)`

dbt/adapters/databricks/impl.py
```diff
@@ -1104,7 +1118,10 @@ def config_type(cls) -> type[MaterializedViewConfig]:
 
     @classmethod
     def _describe_relation(
-        cls, adapter: DatabricksAdapter, relation: DatabricksRelation
+        cls,
+        adapter: DatabricksAdapter,
+        relation: DatabricksRelation,
+        relation_config: DatabricksRelationConfigBase,
```
sd-db: Same comment as above on renaming parameter name for clarity

dbt/adapters/databricks/impl.py
```diff
@@ -1153,16 +1173,35 @@ def config_type(cls) -> type[IncrementalTableConfig]:
...
     ) -> RelationResults:
         results = {}
         kwargs = {"relation": relation}
 
         if not relation.is_hive_metastore():
-            results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
-            results["information_schema.column_tags"] = adapter.execute_macro(
-                "fetch_column_tags", kwargs=kwargs
+            table_tag_config = (
```
sd-db: To be more clear in our naming similarly best to change this to `desired_table_tag_configs`

dbt/adapters/databricks/impl.py
```diff
@@ -1153,16 +1173,35 @@ def config_type(cls) -> type[IncrementalTableConfig]:
...
+                relation_config.config.get(TagsProcessor.name) if relation_config else None
+            )
+            if table_tag_config is None or table_tag_config.requires_server_metadata_for_diff():
```
sd-db, line 1208: This condition check is a bit confusing, can you help explain a scenario where we would need the `table_tag_config is None` path?

dbt/adapters/databricks/impl.py
```diff
@@ -1153,16 +1173,35 @@ def config_type(cls) -> type[IncrementalTableConfig]:
...
     ) -> RelationResults:
         results = {}
         kwargs = {"relation": relation}
 
         if not relation.is_hive_metastore():
-            results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
-            results["information_schema.column_tags"] = adapter.execute_macro(
-                "fetch_column_tags", kwargs=kwargs
+            table_tag_config = (
+                relation_config.config.get(TagsProcessor.name) if relation_config else None
```
sd-db: `relation_config` is of type `DatabricksRelationConfigBase`. According to the function singature it should not be None. Either we change the function signature or we remove the None check from here

dbt/adapters/databricks/impl.py
```diff
@@ -1191,15 +1230,26 @@ def config_type(cls) -> type[ViewConfig]:
 
     @classmethod
     def _describe_relation(
-        cls, adapter: DatabricksAdapter, relation: DatabricksRelation
+        cls,
```
sd-db, line 1256: Similar comments as with `IncrementalTableAPI`

tests/functional/adapter/views/test_view_metadata_fetch_skips.py
```diff
@@ -0,0 +1,72 @@
...
+                "+view_update_via_alter": True,
+            },
+        }
+
+    def test_second_view_run_fails_when_tag_fetch_is_required(self, project):
+        # The first run creates the view; the second run exercises the existing-relation
+        # alter/config-diff path where adapter.get_relation_config() may fetch tags.
+        util.run_dbt(["run"])
+        _, logs = util.run_dbt_and_capture(["run"], expect_pass=False)
+        util.assert_message_in_logs("fetch_tags should not be called", logs)
```
sd-db: We should not rely on logs to check for behaviour, let us figure out some other way to check for the same

tests/functional/adapter/incremental/test_incremental_metadata_fetch_skips.py
```diff
@@ -0,0 +1,69 @@
...
+    @pytest.fixture(scope="class")
+    def macros(self):
+        return {"fail_if_tag_fetch_called.sql": FAIL_IF_TAG_AND_COLUMN_TAG_FETCH_CALLED_MACROS}
+
+    def test_second_incremental_run_fails_when_column_tag_fetch_is_required(self, project):
+        # The first run creates the relation; the second run exercises the existing-relation
+        # path where adapter.get_relation_config() may attempt metadata fetches.
+        util.run_dbt(["run"])
+        _, logs = util.run_dbt_and_capture(["run"], expect_pass=False)
+        util.assert_message_in_logs("fetch_column_tags should not be called", logs)
```
sd-db: We should not rely on logs to check for behaviour, let us figure out some other way to check for the same

dbt/adapters/databricks/impl.py
```diff
@@ -1153,16 +1173,35 @@ def config_type(cls) -> type[IncrementalTableConfig]:
...
+                relation_config.config.get(TagsProcessor.name) if relation_config else None
+            )
+            if table_tag_config is None or table_tag_config.requires_server_metadata_for_diff():
+                results["information_schema.tags"] = adapter.execute_macro(
+                    "fetch_tags", kwargs=kwargs
+                )
+            else:
+                results["information_schema.tags"] = None
+
+            column_tag_config = (
```
sd-db, line 1217: Similar comments on naming + checks on `desried_column_tag_config`

dbt/adapters/databricks/impl.py
```diff
@@ -964,15 +967,17 @@ def parse_columns_and_constraints(
         return enriched_columns, parsed_constraints
 
     @available.parse(lambda *a, **k: {})
-    def get_relation_config(self, relation: DatabricksRelation) -> DatabricksRelationConfigBase:
+    def get_relation_config(
```
sd-db, line 973: Since this function is also exposed at a Jinja layer this is a breaking API change for external macro callers.Due to this I would be more comfortable with the change targeting 1.12 rather

sd-db (changes requested):
Overall intentation of the change looks good, there are a few things specifically where we need changes

- Function signature - make new param optional + naming + other related changes
- Coverage - Can you confirm on coverage across all supported materialisation types (MV/ST missing; Views don't have support for column_tags)
- Functional test changes

dbt/adapters/databricks/impl.py
sd-db: I checked around in the macros and the names `model_config` and `exsiting_config` seem to be established patterns. We should use that instead to be in sync. So better to rename the parameter `relation_config` --> `model_config`

dbt/adapters/databricks/impl.py
tejassp-db: Fixed.

dbt/adapters/databricks/impl.py
tejassp-db: Now it is the case, since we made argument optional.

dbt/adapters/databricks/impl.py
tejassp-db: Suffix `_config` and surrounding code makes it obvious that its coming from the user configuration.

dbt/adapters/databricks/impl.py
tejassp-db: Fixed.

dbt/adapters/databricks/impl.py
tejassp-db: Fixed.

dbt/adapters/databricks/impl.py
```diff
@@ -1080,19 +1093,22 @@ def _describe_relation(
...
     def get_from_relation(
-        cls, adapter: DatabricksAdapter, relation: DatabricksRelation
+        cls,
+        adapter: DatabricksAdapter,
+        relation: DatabricksRelation,
+        model_config: Optional[DatabricksRelationConfigBase] = None,
     ) -> DatabricksRelationConfig:
         """Get the relation config from the relation."""
 
-        relation_config = super().get_from_relation(adapter, relation)
```
sd-db, line 1094: nit: let us revert back the name here to `relation_config` from `full_relation_config` as the original reason for variable name shadowing is no longer applicable here

tejassp-db: Fixed.

dbt/adapters/databricks/impl.py
```diff
@@ -1116,8 +1135,13 @@ def _describe_relation(
         results["information_schema.views"] = get_first_row(
             adapter.execute_macro("get_view_description", kwargs=kwargs)
         )
-        results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
         results["show_tblproperties"] = adapter.execute_macro("fetch_tbl_properties", kwargs=kwargs)
+
+        table_tag_config = model_config.config.get(TagsProcessor.name) if model_config else None
+        if table_tag_config is None or table_tag_config.requires_server_metadata_for_diff():
```
sd-db, line 1150: While this is correct, it is quite difficult to reason  on the `table_tag_config is None` route and why that leads to a macro to `fetch_tags`. We should add a note here that this ie the default behaviour that was present before our change (i.e. model_config is None)

dbt/adapters/databricks/impl.py
```diff
@@ -1116,8 +1135,13 @@ def _describe_relation(
         results["information_schema.views"] = get_first_row(
             adapter.execute_macro("get_view_description", kwargs=kwargs)
         )
-        results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
         results["show_tblproperties"] = adapter.execute_macro("fetch_tbl_properties", kwargs=kwargs)
+
+        table_tag_config = model_config.config.get(TagsProcessor.name) if model_config else None
```
sd-db, line 1149: rename to `model_table_tag_config` we should differentiate between desired(model) configs and existing (relation) configs better

dbt/adapters/databricks/impl.py
```diff
@@ -1153,16 +1180,33 @@ def config_type(cls) -> type[IncrementalTableConfig]:
...
+        model_config: Optional[DatabricksRelationConfigBase] = None,
     ) -> RelationResults:
         results = {}
         kwargs = {"relation": relation}
 
         if not relation.is_hive_metastore():
-            results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
-            results["information_schema.column_tags"] = adapter.execute_macro(
-                "fetch_column_tags", kwargs=kwargs
+            table_tag_config = model_config.config.get(TagsProcessor.name) if model_config else None
```
sd-db, line 1207: Same comment as earlier on `MaterializedViewAPI._describe_relation()` on naming + comment (maybe we can add the comment once in any function and that is good enough)

sd-db (commented):
Added some more comments, overall much better. I still feel we should add a comment in all the materialisation types where we are doing this optimisation that this only works because we don't support deleting/removing tags through dbt-databricks, as that will break the optimisations done and better to encode as a comment

dbt/adapters/databricks/impl.py
sd-db: Variable naming should still be intentional and descriptive, we should update it to `model_column_tag_config`

dbt/adapters/databricks/impl.py
sd-db: This is no longer relevant

dbt/adapters/databricks/impl.py
sd-db: Added a separate comment to rename to `model_table_tag_config`

dbt/adapters/databricks/impl.py
sd-db: This is done

tests/functional/adapter/views/test_view_metadata_fetch_skips.py
```diff
@@ -0,0 +1,78 @@
+import pytest
+from dbt.artifacts.schemas.results import RunStatus
+from dbt.tests import util
+
+from tests.functional.adapter.helpers import FAIL_IF_TAG_FETCH_CALLED_MACROS
+
+VIEW_WITHOUT_TAGS_SQL = """
```
sd-db: we should move this inside fixtures file

tests/functional/adapter/helpers.py
```diff
@@ -0,0 +1,44 @@
...
+  {{ exceptions.raise_compiler_error("fetch_tags should not be called") }}
+{% endmacro %}
+
+{% macro fetch_column_tags(relation) %}
+  {{ exceptions.raise_compiler_error("fetch_column_tags should not be called") }}
+{% endmacro %}
+"""
+
+
+def get_model_config(project, relation: BaseRelation):
```
sd-db: Do we need this ? Since `model_config` is optional in the tests we can just not pass it and that also tests for other behaviour overall. Look at the places where this gets called it is most a no-op. I think this is mostly an artifact of the config not being Optional earlier. Since now it is we can look to remove as well.

tests/unit/relation_configs/test_materialized_view_config.py
```diff
@@ -154,7 +154,7 @@ def test_describe_relation_fetches_tags(self):
         adapter = MagicMock()
         adapter.execute_macro.return_value = MagicMock()
 
-        results = MaterializedViewAPI._describe_relation(adapter, MagicMock())
+        results = MaterializedViewAPI._describe_relation(adapter, MagicMock(), Mock())
```
sd-db: nit: while the change is okay, it would be better if just did not pass `model_config` here, i.e no change as `model_config` being `None`, triggers the actual tag fetch macro which is what the test is looking for. This test was added when there no optimisation to not fetch and better to test that path instead of the else path.

sd-db (commented):
Reviewing tests 1/2

tests/functional/adapter/helpers.py
sd-db: Currently this is adding extra complexity while mostly being no-op

tests/unit/test_adapter.py
```diff
@@ -1311,6 +1321,195 @@ def test_get_columns_reraises_other_database_errors(
             )
 
 
+class TestDescribeRelationMetadataFetchPlanning:
```
sd-db, line 1326: nit: These tests look really good, but we missed MV materealisation here which would have this even more complete. Fine without as well.

sd-db (commented):
Reviewing Tests 2/2

tests/unit/test_adapter.py
tejassp-db: added.

dbt/adapters/databricks/impl.py
tejassp-db: Config only comes from user and its obvious here.

sd-db (approved):
Changes look good, thanks !

PR #1393, opened 2026-04-13
chore: remove temporary pull_request trigger from warmDepsCache

sd-db:
## Summary
- Remove the temporary `pull_request` trigger that was added to `warmDepsCache.yml` for testing
- The cache warmer should only run on push to main, daily schedule, and manual dispatch

## Test plan
- [ ] Verify warmDepsCache no longer triggers on PRs


sd-db:
Superseded by PR #1394 which removes the temp trigger directly.

PR #1401, opened 2026-04-14
fix: enable TestWorkflowJob functional test that was unreachable on all profiles

sd-db:
## Summary

- Fix `TestWorkflowJob` functional test that was **never actually running** due to three compounding bugs:
  1. `skip_profile` listed all 3 available profiles (`databricks_cluster`, `databricks_uc_sql_endpoint`, `databricks_uc_cluster`) — test was skipped everywhere
  2. Used `simple_python_model` which hardcodes `submission_method='serverless_cluster'` in `dbt.config()`, overriding the YAML's `workflow_job` setting — test was exercising the wrong code path
  3. `workflow_schema` included `max_retries` which is not a valid `jobs.create()` parameter

resolves follow-up to #1360

### changes

- Add `workflow_python_model` fixture without `submission_method` in `dbt.config()` so the YAML schema's `submission_method: workflow_job` takes effect
- Remove `databricks_uc_cluster` from skip list so the test runs on at least one profile
- Remove invalid `max_retries` from `workflow_schema`

### checklist

- [x] i have run this code in development and it appears to resolve the stated issue
- [x] this pr includes tests, or tests are not required/relevant for this pr
- [x] i have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section.

## Test plan

- [x] Verified test was SKIPPED on all 3 profiles before the fix
- [x] Verified test PASSES on `databricks_uc_cluster` profile after the fix (ran against live cluster)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Fixes
 
 - Fix `workflow_job` Python model submission method failing with dictionary attribute error ([#1360](https://github.com/databricks/dbt-databricks/issues/1360))
+- Fix `TestWorkflowJob` functional test that was unreachable on all profiles due to incorrect skip list, wrong model fixture, and invalid `max_retries` parameter ([#1360](https://github.com/databricks/dbt-databricks/issues/1360))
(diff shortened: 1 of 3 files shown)


benc-db (dismissed):
Approved assuming the test passes :P

tejassp-db (approved):

PR #1409, opened 2026-04-20
fix: prevent spurious MicrobatchConcurrency warning on every run (#1406)

### LINKED ISSUE #1406: MicrobatchConcurrency warning emitted even when no microbatch models are present
### Describe the bug

When using the dbt-databricks adapter (v1.11.6+), dbt emits the following warning on every run and in Studio status — even when the project contains **no microbatch incremental models**:
```
[WARNING]: Enable concurrent execution of microbatch incremental batches. When enabled, dbt will run microbatch batches in parallel threads instead of sequentially. The microbatch strategy itself requires dbt-core 1.9+; this flag controls whether batches run concurrently.
You may opt into the new behavior sooner by setting `flags.use_concurrent_microbatch` to `True` in `dbt_project.yml`.
Visit https://docs.getdbt.com/reference/global-configs/behavior-changes for more information.
```

This appears to be caused by `dbt-databricks` advertising the `MicrobatchConcurrency` capability to `dbt-core` unconditionally, which triggers the behavior change warning regardless of whether the project actually uses the microbatch strategy.

### Steps To Reproduce

1. Use `dbt-databricks` v1.11.6+
2. Ensure the project has no models using `materialized: incremental` with `strategy: microbatch`
3. Run any dbt command (eg, dbt run, dbt build) or check the status details in dbt Studio/IDE
4. Observe the [WARNING] above emitted on every run

### Expected behavior
The warning should only fire when the adapter supports `MicrobatchConcurrency` and the project actually contains microbatch models. Projects that do not use the microbatch strategy should not see this warning at all.

n/a

### System information
```
dbt-core 1.9+, dbt-databricks 1.11.6
```

Mac OS

n/a

### Additional context
This is potentially breaking for customers using `--warn-error` or `--warn-error-options={"error": "all"}`, as the warning would be elevated to an error - causing runs to fail despite no misconfiguration in the project.

The suggested fix would be to gate the `MicrobatchConcurrency` capability advertisement (or the associated warning) on whether the project actually contains microbatch models.

sd-db:
Fixes #1406.

`supports(Capability.MicrobatchConcurrency)` read the `use_concurrent_microbatch` behavior flag via `bool(...)`, which goes through the warning-firing `BehaviorFlagRendered.setting` path. dbt-core probes this capability on every parse (via `check_forcing_batch_concurrency`), so the one-shot `BehaviorChangeEvent` fired on every run regardless of whether the project contained microbatch models — breaking `--warn-error` builds.

Read via `.no_warn` instead — same gating, no event.

## Test plan
- [x] `hatch run unit` → 742 passed, 0 regressions
- [x] New `test_microbatch_concurrency_probe_does_not_warn`

### DIFF
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -261,9 +261,9 @@ class DatabricksAdapter(SparkAdapter):
         ]
 
     def supports(self, capability: Capability) -> bool:
-        # Gate MicrobatchConcurrency on the use_concurrent_microbatch behavior flag.
         if capability == Capability.MicrobatchConcurrency:
-            return bool(self.behavior.use_concurrent_microbatch)
+            # `.no_warn` avoids a BehaviorChangeEvent on dbt-core's per-parse probe.
+            return self.behavior.use_concurrent_microbatch.no_warn
         return super().supports(capability)
 
     def quote(self, identifier):  # type: ignore[override,no-untyped-def]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## dbt-databricks next (TBD)
+
+### Fixes
+
+- Fix spurious `MicrobatchConcurrency` behavior-change warning firing on every run regardless of whether the project contained microbatch models ([#1406](https://github.com/databricks/dbt-databricks/issues/1406))
+
 ## dbt-databricks 1.11.7 (Apr 17, 2026)
 
 ### Features
--- a/tests/unit/test_adapter_capabilities.py
+++ b/tests/unit/test_adapter_capabilities.py
@@ -7,6 +7,7 @@ from unittest.mock import Mock, patch
 
 import pytest
 from dbt.adapters.capability import Capability
+from dbt_common.events.event_manager_client import add_callback_to_manager
 from dbt_common.exceptions import DbtConfigError
 
 from dbt.adapters.databricks.dbr_capabilities import DBRCapabilities, DBRCapability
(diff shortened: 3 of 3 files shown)


dbt/adapters/databricks/impl.py
```diff
@@ -261,9 +261,9 @@ def _behavior_flags(self) -> list[BehaviorFlag]:
         ]
 
     def supports(self, capability: Capability) -> bool:
-        # Gate MicrobatchConcurrency on the use_concurrent_microbatch behavior flag.
         if capability == Capability.MicrobatchConcurrency:
-            return bool(self.behavior.use_concurrent_microbatch)
+            # `.no_warn` avoids a BehaviorChangeEvent on dbt-core's per-parse probe.
+            return self.behavior.use_concurrent_microbatch.no_warn
```
tejassp-db, line 266: Is there any way around using `use_concurrent_microbatch` such that it is more tightly tied with the declared behavior flags. Right now it is a property name which has to match what was defined earlier in `USE_CONCURRENT_MICROBATCH.name`.

tejassp-db (dismissed):
It would be nice if there is a abstraction to track behavior flags, and the silenced ones are always called with `.no_warn`.

dbt/adapters/databricks/impl.py
sd-db: Makes sense, we can do this as part of a refactor later for flags; when looking at warnings holistically. Though I am not sure if this will lead to a lot of code quality improvements

PR #1414, opened 2026-04-21
fix: prevent _cache_dbr_capabilities from caching None-version entries (#1398)

### LINKED ISSUE #1398: bug: _cache_dbr_capabilities permanently poisons capability cache when version query fails
### Describe the bug

`_cache_dbr_capabilities` unconditionally writes to the class-level `_dbr_capabilities_cache` dict, even when `_query_dbr_version` returns `None`. Once written, the `if http_path not in cls._dbr_capabilities_cache` guard prevents any subsequent call from overwriting it. This permanently disables all capability-gated features for that `http_path` for the lifetime of the dbt process.

The `_try_cache_dbr_capabilities` method introduced in #1355 correctly guards against this with `if dbr_version is not None`, but only for the eager path in `_create_fresh_connection()`. The authoritative path in `open()` still uses the unguarded `_cache_dbr_capabilities`.

### Root cause

[`connections.py:200-209`](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/connections.py#L200-L209):

```python
@classmethod
def _cache_dbr_capabilities(cls, creds, http_path):
    if http_path not in cls._dbr_capabilities_cache:
        is_cluster = is_cluster_http_path(http_path, creds.cluster_id)
        dbr_version = cls._query_dbr_version(creds, http_path)  # can return None

        cls._dbr_capabilities_cache[http_path] = DBRCapabilities(
            dbr_version=dbr_version,        # None written to cache
            is_sql_warehouse=not is_cluster,
        )
```

`DBRCapabilities(dbr_version=None)` makes `has_capability()` return `False` for all version-gated capabilities ([`dbr_capabilities.py:104`](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/dbr_capabilities.py#L104)).

The check-then-write guard (`if http_path not in cls._dbr_capabilities_cache`) prevents subsequent calls from correcting the poisoned entry.

### Trigger conditions

`_query_dbr_version` returns `None` when:
1. `credentials_manager` is `None` at call time (raises `DbtRuntimeError`, caught by `except Exception: pass` at [`connections.py:195-196`](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/connections.py#L195-L196))
2. Cluster is starting up / not yet reachable
3. Transient network error during the `SET spark.databricks.clusterUsageTags.sparkVersion` query

### Impact

All 7 capability-gated features are permanently disabled for the affected `http_path`:

| Capability | Min Version | Effect When Disabled |
|------------|-------------|----------------------|
| `ICEBERG` | 14.3 | **Hard error**: "iceberg requires DBR 14.3+" |
| `INSERT_BY_NAME` | 12.2 | Silent fallback to positional insert (risk of wrong column mapping with schema drift) |
| `REPLACE_ON` | 17.1 | Silent fallback to non-replace_on strategy |
| `COMMENT_ON_COLUMN` | 16.1 | Column comments silently skipped |
| `JSON_COLUMN_METADATA` | 16.2 | Fallback to legacy column introspection |
| `TIMESTAMPDIFF` | 10.4 | Fallback to non-timestampdiff SQL |
| `STREAMING_TABLE_JSON_METADATA` | 17.1 | Fallback to legacy metadata path |

`ICEBERG` is the most visible — it raises a hard error. Others silently degrade.

### Contributing factor: silent exception swallowing

[`connections.py:195-196`](https://github.com/databricks/dbt-databricks/blob/main/dbt/adapters/databricks/connections.py#L195-L196):

```python
except Exception:
    pass
```

This catches ALL errors in `_query_dbr_version` without any logging, making the cache poisoning invisible to users. There is no diagnostic output when the version query fails — the user only sees the downstream capability error.

### Suggested fix

Apply the same `None`-guard pattern that `_try_cache_dbr_capabilities` already uses:

```python
@classmethod
def _cache_dbr_capabilities(cls, creds, http_path):
    if http_path not in cls._dbr_capabilities_cache:
        is_cluster = is_cluster_http_path(http_path, creds.cluster_id)
        dbr_version = cls._query_dbr_version(creds, http_path)

        if dbr_version is not None:
            cls._dbr_capabilities_cache[http_path] = DBRCapabilities(
                dbr_version=dbr_version,
                is_sql_warehouse=not is_cluster,
            )
```

This allows subsequent calls (on retry or from another thread) to attempt the version query again rather than being blocked by a poisoned entry. The retry logic already exists in `open()` via `exponential_backoff`.

Additionally, adding debug-level logging to the `except Exception` block in `_query_dbr_version` would make failures diagnosable.

### Related

- #1354 — the named compute bug that exposed this code path
- #1355 — introduced `_try_cache_dbr_capabilities` with the `None` guard, but only for the eager path

### System information

Identified via code review of `connections.py` on main branch (post-#1355 merge). Affects all versions with the capability caching system.

sd-db:
Fixes #1398.

`_cache_dbr_capabilities` was caching `DBRCapabilities(dbr_version=None)` when `_query_dbr_version()` failed, and the idempotency guard then blocked every retry — silently disabling all version-gated features for the life of the process.

Fix: apply the same `None`-guard already used by the sibling `_try_cache_dbr_capabilities`, and switch the downstream read in `open()` to `.get(..., DBRCapabilities())` so a now-possibly-missing entry doesn't `KeyError`. Failure-case behavior is unchanged for the current connection; subsequent `open()` calls can now retry instead of staying stuck.

Follow-ups tracked in #1413.

## Test plan

- [x] `hatch run unit tests/unit/test_connection_manager.py` — 745 pass.
- [x] New `TestCacheDbr` regression mirrors `TestTryCacheDbr` + a `test_retry_succeeds_after_transient_failure` case specific to #1398 (verified it fails on `main`, passes with the fix).
- [ ] CI functional run.

### DIFF
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -199,22 +199,12 @@ class DatabricksConnectionManager(SparkConnectionManager):
 
     @classmethod
     def _cache_dbr_capabilities(cls, creds: DatabricksCredentials, http_path: str) -> None:
-        if http_path not in cls._dbr_capabilities_cache:
-            is_cluster = is_cluster_http_path(http_path, creds.cluster_id)
-            dbr_version = cls._query_dbr_version(creds, http_path)
-            if dbr_version is not None:
-                cls._dbr_capabilities_cache[http_path] = DBRCapabilities(
-                    dbr_version=dbr_version,
-                    is_sql_warehouse=not is_cluster,
-                )
+        """Cache DBR capabilities for an http_path on first successful version query.
 
-    @classmethod
-    def _try_cache_dbr_capabilities(cls, creds: DatabricksCredentials, http_path: str) -> None:
-        """Like _cache_dbr_capabilities, but only writes to the cache when the version query
-        actually succeeds. This prevents a failed eager lookup (e.g. credentials_manager not yet
-        set, cluster still spinning up) from storing a None-version entry that the idempotency
-        guard in open() would later treat as authoritative, causing all capability checks to
-        return False.
+        Only writes when the version query succeeds: a failed lookup (credentials_manager
+        not yet set, cluster spinning up) must not store a None-version entry, since the
+        idempotency guard would then treat it as authoritative and disable every
+        capability check for the rest of the process.
         """
         if http_path not in cls._dbr_capabilities_cache:
             is_cluster = is_cluster_http_path(http_path, creds.cluster_id)
@@ -305,7 +295,7 @@ class DatabricksConnectionManager(SparkConnectionManager):
         conn.http_path = QueryConfigUtils.get_http_path(query_header_context, creds)
         conn.thread_identifier = cast(tuple[int, int], self.get_thread_identifier())
         conn._query_header_context = query_header_context
-        self._try_cache_dbr_capabilities(creds, conn.http_path)
+        self._cache_dbr_capabilities(creds, conn.http_path)
         conn.capabilities = self._get_capabilities_for_http_path(conn.http_path)
         conn.handle = LazyHandle(self.open)
 
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -194,8 +194,8 @@ class DatabricksConnectionManager(SparkConnectionManager):
                     result = cursor.fetchone()
                     if result:
                         return SqlUtils.extract_dbr_version(result[1])
-        except Exception:
-            pass
+        except Exception as e:
+            logger.debug(f"Failed to query DBR version for http_path={http_path}: {e}")
 
         return None
 
--- a/dbt/adapters/databricks/connections.py
+++ b/dbt/adapters/databricks/connections.py
@@ -167,8 +167,9 @@ class DatabricksConnectionManager(SparkConnectionManager):
         databricks_conn = cast(DatabricksDBTConnection, conn)
         return is_cluster_http_path(databricks_conn.http_path, conn.credentials.cluster_id)
 
-    def _get_capabilities_for_http_path(self, http_path: str) -> DBRCapabilities:
-        return self._dbr_capabilities_cache.get(http_path, DBRCapabilities())
+    @classmethod
+    def _get_capabilities_for_http_path(cls, http_path: str) -> DBRCapabilities:
+        return cls._dbr_capabilities_cache.get(http_path, DBRCapabilities())
 
     @classmethod
     def _query_dbr_version(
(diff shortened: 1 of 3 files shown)


dbt/adapters/databricks/connections.py
```diff
@@ -202,11 +202,11 @@ def _cache_dbr_capabilities(cls, creds: DatabricksCredentials, http_path: str) -
         if http_path not in cls._dbr_capabilities_cache:
             is_cluster = is_cluster_http_path(http_path, creds.cluster_id)
             dbr_version = cls._query_dbr_version(creds, http_path)
-
-            cls._dbr_capabilities_cache[http_path] = DBRCapabilities(
-                dbr_version=dbr_version,
-                is_sql_warehouse=not is_cluster,
-            )
+            if dbr_version is not None:
```
tejassp-db: Now `_cache_dbr_capabilities` and `_try_cache_dbr_capabilities` are the exact same methods. Can we get rid of one of them?

dbt/adapters/databricks/connections.py
```diff
@@ -506,9 +506,9 @@ def connect() -> DatabricksHandle:
                 if conn:
                     databricks_connection.session_id = conn.session_id
                     cls._cache_dbr_capabilities(creds, databricks_connection.http_path)
-                    databricks_connection.capabilities = cls._dbr_capabilities_cache[
-                        databricks_connection.http_path
-                    ]
+                    databricks_connection.capabilities = cls._dbr_capabilities_cache.get(
```
tejassp-db: Can we use `_get_capabilities_for_http_path` which does the exact same thing? It is currently an instance method, we might have to convert it to a class method. We should be using the same method across the whole class for getting capabilities.

tejassp-db (changes requested):

dbt/adapters/databricks/connections.py
sd-db: I wanted to do this in a followup as it was out of scope here but makes sense, let me do this here.

dbt/adapters/databricks/connections.py
sd-db: Done

dbt/adapters/databricks/connections.py
tejassp-db: Should we log this?

tejassp-db (dismissed):

dbt/adapters/databricks/connections.py
sd-db: added

---
_This comment was generated with [GitHub MCP](http://go/mcps)._

tejassp-db (approved):

PR #1421, opened 2026-04-23
fix: warn on contract.enforced=true for materialized_view

### LINKED ISSUE #1279: Materialized views silently ignore `contract.enforced: true`
### Describe the bug
When setting [`+materialized: materialized_view`](https://docs.getdbt.com/reference/resource-configs/databricks-configs#materialized-views-and-streaming-tables) on a model simultaneously with [`+contract: { enforced: true }`](https://docs.getdbt.com/reference/resource-configs/contract), I expected dbt to warn me if my documented column names/types got out of sync with my model SQL... or at least warn me if it's not supported!

I've had it set for a few weeks and just now realized that a model of mine was out of sync. 🙃 I then tested changing the materialization to `table` in dev and sure enough I got a contract failure.

I vaguely recall noticing that the docs say it's ["Models materialized as `table`, `view`, and `incremental` ..."](https://docs.getdbt.com/reference/resource-configs/contract), but I think at the time I internally thought that `+materialized: materialized_view` was essentially a sub-case of `+materialized: view` so didn't think my use case was excluded.

It seems easy enough to add this to the remaining materialization types for dbt-databricks.

### Steps To Reproduce

Create a SQL model with the above configs, with a list of column names/types in `my_model.yml` which does not match the columns defined in `my_model.sql`. Call `dbt run`; it'll succeed with no warnings or errors.

### Expected behavior
Either:
1. something along the lines of ⚠️ `[WARNING]: model contract unenforceable on models materialized as 'materialized_view' (models/model.sql)`, *or*
2. ❌ `Compilation Error in model my_model (models/my_model.sql): Contracted models require data_type to be defined for each column. Please ensure that the column name and data_type are defined within the YAML configuration for the [...] column(s).`

### System information
```
$ dbt --version
WARNING:thrift.transport.sslcompat:using legacy validation callback
Core:
  - installed: 1.10.13
  - latest:    1.10.15 - Update available!

Plugins:
  - databricks: 1.10.14 - Update available!
  - spark:      1.9.3   - Up to date!

  At least one plugin is out of date with dbt-core.
```

MacOS 15.7.1 (24G231)

Python 3.12.5

### Additional context

I've (manually) explored the call graph a bit starting from a search in my project's folder (including inclusions from its venv):
Image

From my spelunking it looks like dbt-databricks allows setting up _constraints_ on materialized views -- but won't do the preflight "contract" check to for column names/types consistency.

sd-db:
Resolves #1279

Upstream dbt does not support contracts on materialized views ([docs](https://docs.getdbt.com/docs/mesh/govern/model-contracts)). dbt-databricks was silently accepting `contract.enforced: true`, hiding yml/SQL drift. This PR emits a warning pointing at the dbt docs. Strictly additive; no existing build outcome changes.

### Checklist

- [x] I have updated the `CHANGELOG.md`

### DIFF
--- a/dbt/include/databricks/macros/relations/materialized_view/create.sql
+++ b/dbt/include/databricks/macros/relations/materialized_view/create.sql
@@ -24,6 +24,9 @@
   {%- set model_columns = model.get('columns', {}) -%}
   {%- set contract_config = config.get('contract') -%}
   {%- if contract_config and contract_config.enforced -%}
+    {%- do exceptions.warn(
+         "model '" ~ model.name ~ "' sets contract.enforced=true, but contracts are not supported for materialized_view. See https://docs.getdbt.com/docs/mesh/govern/model-contracts"
+       ) -%}
     {%- set model_constraints = model.get('constraints', []) -%}
   {%- else -%}
     {%- set model_constraints = [] -%}
--- a/dbt/include/databricks/macros/relations/materialized_view/create.sql
+++ b/dbt/include/databricks/macros/relations/materialized_view/create.sql
@@ -25,7 +25,7 @@
   {%- set contract_config = config.get('contract') -%}
   {%- if contract_config and contract_config.enforced -%}
     {%- do exceptions.warn(
-         "model '" ~ model.name ~ "' sets contract.enforced=true, but contracts are not supported for materialized_view. See https://docs.getdbt.com/docs/mesh/govern/model-contracts"
+         "contract.enforced=true on materialized_view '" ~ model.name ~ "': not supported by dbt (https://docs.getdbt.com/docs/mesh/govern/model-contracts). dbt-databricks provides best-effort support that may change without notice."
        ) -%}
     {%- set model_constraints = model.get('constraints', []) -%}
   {%- else -%}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## dbt-databricks 1.11.8 (TBD)
+
+### Fixes
+
+- Warn when `contract.enforced: true` is set on a `materialized_view` model ([#1279](https://github.com/databricks/dbt-databricks/issues/1279))
+
 ## dbt-databricks 1.11.7 (Apr 17, 2026)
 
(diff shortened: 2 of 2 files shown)


dbt/include/databricks/macros/relations/materialized_view/create.sql
```diff
@@ -24,6 +24,9 @@
   {%- set model_columns = model.get('columns', {}) -%}
   {%- set contract_config = config.get('contract') -%}
   {%- if contract_config and contract_config.enforced -%}
+    {%- do exceptions.warn(
+         "model '" ~ model.name ~ "' sets contract.enforced=true, but contracts are not supported for materialized_view. See https://docs.getdbt.com/docs/mesh/govern/model-contracts"
+       ) -%}
     {%- set model_constraints = model.get('constraints', []) -%}
```
tejassp-db, line 30: If `contract` is not enforced in materialized views, shouldn't `model_constraints` be assigned to `[]` like the else branch?

sd-db: The warning is a half solve at this point. Some model constraints are supported today while some aren't. At this moment it is difficult to solve either way

- Either we support constraints fully for MVs and deviate from dbt docs.
- We remove support on what we have now and that is a breaking change.

I wanted some more discussion on this, but a warning here atleast tells the customer that this is undocumented/unsupported behaviour.

sd-db: Maybe I can edit the comment (it should still be a warning) that contract enforcement is not fully supported for `materialised_views` with some more specific details on the same.

sd-db: I have changed the wording, to specifically point out that a) `dbt-core` does not support contract enforcement on MVs and b) any behaviour you see for enforcement in the adapter is incidental.

- The warning is required and is a strictly positive change on something that is confusing today
- To solve this we need to more data points and post talking to customers we can look change things.

tejassp-db (approved):

PR #1424, opened 2026-04-25
fix: refresh materialized_view when databricks_tags is set (#1419)

### LINKED ISSUE #1419: materialized_view materialization does not refresh data when no configuration changes are detected (only re-applies tags/grants)
### Describe the bug

When running `dbt run` against an existing `materialized_view` model whose SQL and configuration have not changed since the last run, `dbt-databricks` does **not** issue a `REFRESH MATERIALIZED VIEW` statement (nor a `CREATE OR REPLACE MATERIALIZED VIEW`). The model is reported as successfully run, but the underlying data is not refreshed. The only data-affecting statements issued against the relation are `ALTER MATERIALIZED VIEW … SET TAGS (…)` and grant-related statements.

This contradicts:

1. The dbt-core materialized view spec ([docs.getdbt.com/reference/resource-configs/materialized#creation-precedence](https://docs.getdbt.com/reference/resource-configs/materialized?version=1.11#creation-precedence)), which states that when there are no configuration changes the default action for the type should be performed — *"e.g. apply refresh for a materialized view."*
2. The `dbt-databricks` configuration documentation, which states *"Even when schedules are set, dbt will request that the materialization be refreshed manually when run."*

Command run:
```
dbt run --select <my_mv_model> --target dev
```

Result: dbt reports `OK created sql materialized_view model … [OK in 4.20s]`, but the MV's data has not been refreshed.

### Steps To Reproduce

1. In a Unity Catalog–enabled workspace on a serverless SQL Warehouse, create a `materialized_view` model and let dbt build it once. Minimal example model config:

   ```yaml
   materialized: materialized_view
   on_configuration_change: apply
   persist_docs: {columns: true, relation: true}
   grants:
     select: ["<some_group>"]
   databricks_tags:
     deployment: DBT
   post-hook:
     - "GRANT USE SCHEMA ON SCHEMA {{ model.database }}.{{ model.schema }} TO `<some_group>`"
     - "GRANT USE CATALOG ON CATALOG {{ model.database }} TO `<some_group>`"
   ```

   Model body is a straightforward `SELECT … FROM <upstream_a> JOIN <upstream_b> ON …`.

2. Mutate data in the upstream sources so that the MV would, if refreshed, return different rows.

3. Without changing the model SQL or config, run:
   ```
   dbt run --select <my_mv_model>
   ```

4. Observe via dbt debug logs and/or Databricks SQL Warehouse Query History the exact statements issued against the MV.

### Expected behavior

Per the dbt-core spec and the dbt-databricks docs, the "no configuration changes detected" branch of the `materialized_view` materialization should perform the type's default action — i.e. issue `REFRESH MATERIALIZED VIEW <relation>` (or otherwise trigger an update of the underlying DLT pipeline) — so that the MV's data is refreshed.

The full sequence of statements / debug events issued by the adapter against the MV node is (paraphrased, with identifiers redacted):

1. `describe extended <relation>`
2. `SELECT * FROM system.information_schema.views WHERE … table_name = '<mv_name>'`
3. `SHOW TBLPROPERTIES <relation>`
4. *(debug log line only — no SQL issued)* `Databricks adapter: Pipeline <uuid> completed successfully`
5. Debug: `Applying ALTER to: <relation>`
6. `ALTER MATERIALIZED VIEW <relation> SET TAGS ('deployment' = 'DBT')`
7. Debug: `Applying tags to relation {'deployment': 'DBT'}`
8. `ALTER MATERIALIZED VIEW <relation> SET TAGS ('deployment' = 'DBT')` *(emitted a second time — see Additional context)*
9. Debug: `Applying column tags to relation <relation>`
10. `show grants on <relation>` → `All grants are in place, no revocation or granting needed.`
11. The configured `post-hook` statements: `GRANT USE SCHEMA …`, `GRANT USE CATALOG …`
12. `OK created sql materialized_view model <relation> [OK in 4.20s]`

Notable observations:

- **No `REFRESH MATERIALIZED VIEW` statement is ever issued.**
- **No `CREATE OR REPLACE MATERIALIZED VIEW` statement is ever issued.**
- The `Databricks adapter: Pipeline <uuid> completed successfully` debug line is logged ∼1.2 seconds after the previous SQL statement returns. This is far too short to be a real refresh of a non-trivial MV; it looks like the adapter is inspecting the existing pipeline state rather than triggering a new update. (If a real update were triggered and was a no-op, that would be useful to surface in the logs explicitly.)
- `ALTER … SET TAGS` is emitted twice in succession with identical arguments — likely a separate minor bug.

I can provide a sanitised debug log excerpt on request.

### System information

```
Core:
  - installed: 1.11.8

Plugins:
  - databricks: 1.11.7

databricks-sql-connector: 4.1.5
```

**The operating system you're using:** Linux (executed inside a Databricks Workflows "dbt" task on Azure Databricks)

**The output of `python --version`:** Python provided by the Databricks dbt task runtime (happy to confirm exact patch version if useful)

### Additional context

- Compute: serverless SQL Warehouse, Unity Catalog enabled.
- The MV is backed by the standard managed DLT pipeline that Databricks creates for materialized views. Pipeline state in the DLT UI confirms no new update was started during the dbt invocation in question; only an existing terminal-state update is visible.
- This is reproducible across multiple `materialized_view` models in our project; it does not appear to depend on the specific model SQL.
- The duplicate `ALTER MATERIALIZED VIEW … SET TAGS` statement noted above looks like a separate small bug worth a glance while in this code path.
- **Impact:** users relying on `dbt run` to keep MVs fresh per the documented contract silently end up with stale data. The dbt task reports success and emits no warning, so the staleness is invisible from the dbt side.
- **Current workarounds in use:**
  - Adding an explicit `post_hook="REFRESH MATERIALIZED VIEW " ~ this` (or similar) to the model
  - Attaching a `SCHEDULE` to the MV in Databricks and not relying on dbt for refresh.
  Both work, but should not be necessary if the documented behaviour is the intended one.
- If the current behaviour is in fact intentional, the dbt-core materialized view reference and the `dbt-databricks` configuration docs should be updated, as both currently imply a refresh will occur on `dbt run`.

sd-db:
Fixes #1419. MVs configured with `databricks_tags` were silently going stale on `dbt run` — only `ALTER ... SET TAGS` was issued, never `REFRESH MATERIALIZED VIEW`.

`MaterializedViewAPI._describe_relation` was the only `*API._describe_relation` not fetching `information_schema.tags`, so existing tags always parsed as `{}`. Any model with `databricks_tags` produced a spurious tag diff that routed the build to the alter path instead of refresh. One-line fix mirroring `ViewAPI` / `IncrementalTableAPI`.

## Test plan
- [x] New functional test + unit regression guard
- [x] Full MV functional suite green (21/21) on serverless SQL warehouse
- [x] Full unit suite green (742 passed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## dbt-databricks 1.11.8 (TBD)
+
+### Fixes
+
+- Fix `materialized_view` models with `databricks_tags` silently going stale on `dbt run`. `MaterializedViewAPI._describe_relation` was not fetching `information_schema.tags`, so existing tags always parsed as empty, producing a spurious tag diff that routed the materialization to `ALTER ... SET TAGS` instead of `REFRESH MATERIALIZED VIEW` ([#1419](https://github.com/databricks/dbt-databricks/issues/1419))
+
 ## dbt-databricks 1.11.7 (Apr 17, 2026)
 
 ### Features
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -1080,6 +1080,7 @@ class MaterializedViewAPI(DeltaLiveTableAPIBase[MaterializedViewConfig]):
         results["information_schema.views"] = get_first_row(
             adapter.execute_macro("get_view_description", kwargs=kwargs)
         )
+        results["information_schema.tags"] = adapter.execute_macro("fetch_tags", kwargs=kwargs)
         results["show_tblproperties"] = adapter.execute_macro("fetch_tbl_properties", kwargs=kwargs)
         return results
 
--- a/tests/functional/adapter/materialized_view_tests/test_refresh_with_tags.py
+++ b/tests/functional/adapter/materialized_view_tests/test_refresh_with_tags.py
@@ -1,0 +1,65 @@
+import pytest
+from dbt.adapters.base import BaseRelation
+from dbt.tests import util
+
+from dbt.adapters.databricks.relation import DatabricksRelationType
+
+SEED_CSV = """id,value
+1,100
+2,200
+""".lstrip()
+
+MV_WITH_TAGS = """
+{{ config(
+    materialized='materialized_view',
+    on_configuration_change='apply',
+    databricks_tags={'deployment': 'DBT'},
+) }}
+select * from {{ ref('mv_seed') }}
+"""
+
+
+@pytest.mark.dlt
+@pytest.mark.skip_profile("databricks_cluster", "databricks_uc_cluster")
+class TestMaterializedViewRefreshesWithTags:
+    @pytest.fixture(scope="class")
+    def seeds(self):
+        return {"mv_seed.csv": SEED_CSV}
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"mv_with_tags.sql": MV_WITH_TAGS}
+
+    @staticmethod
+    def _row_count(project, relation: BaseRelation) -> int:
+        return project.run_sql(f"select count(*) from {relation}", fetch="one")[0]
+
+    def test_mv_refreshes_on_rerun_when_tags_set(self, project):
+        util.run_dbt(["seed"])
+        util.run_dbt(["run", "--models", "mv_with_tags"])
+
+        mv = project.adapter.Relation.create(
+            identifier="mv_with_tags",
+            schema=project.test_schema,
+            database=project.database,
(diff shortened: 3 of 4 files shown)


tejassp-db (commented):
Are all metadata for MV covered now?

tejassp-db (approved):

sd-db:
> Are all metadata for MV covered now?

I have not done a holistic check but resolved what was reported in the current issue/bug report. I can take a look at this as a followup

PR #1429, opened 2026-04-28
fix: suppress use_managed_iceberg warning for non-Iceberg projects

### LINKED ISSUE #1266: use_managed_iceberg warning popping up for projects that don't use iceberg at all
### Describe the bug
dbt projects are seeing the following warning emitted:
```
Use managed Iceberg tables when table_format is iceberg. When this flag is disabled, UniForm is used instead.
You may opt into the new behavior sooner by setting "flags use_managed _iceberg to 'True" in "dbt_project.yml*.
Visit https://docs-getdbt.com/reference/global-configs/behavior-changes for more information.
```
_even when_ their project isn't using iceberg at all.

### Expected behavior
We should suppress this warning if the user isn't using iceberg.

### Additional context
this is the PR that added this flag -> https://github.com/databricks/dbt-databricks/pull/1146
docs here -> https://docs.getdbt.com/reference/global-configs/databricks-changes#use-managed-iceberg

tejassp-db:
## Summary

Fixes #1266.

dbt-core's `BehaviorFlag` emits a deprecation warning the first time a flag with `default=False` is read. As a result, every dbt run — even on Delta-only projects with zero Iceberg models — logged a confusing `use_managed_iceberg` warning, with no clean way to silence it short of opting in to the flag itself. 

This PR fixes this by accessing the flag with `non_warn` in the code paths. Also flipped the order of comparison in the jinja macro, now warning fires only if its a iceberg model and `use_managed_icerberg` is set to `false`.

PECOBLR-2638


tests/functional/adapter/iceberg/test_managed_iceberg_warning.py
```diff
@@ -0,0 +1,34 @@
...
+    """A Delta-only project must not emit the `use_managed_iceberg` deprecation warning on
+    `dbt run`."""
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"delta_model.sql": fixtures.basic_table}
+
+    def test_use_managed_iceberg_warning_absent(self, project):
+        _, logs = util.run_dbt_and_capture(["--debug", "run"])
+        util.assert_message_in_logs(_iceberg_warning_text, logs, False)
```
sd-db: functional tests are used to check on overall behaviour. Checking messages in logs is too brittle/implementation specific and defeats the purpose of functional tests. Better to have this in unit tests if required.

sd-db: For testing of this change we should not use functional tests

dbt/adapters/databricks/impl.py
```diff
@@ -281,7 +282,7 @@ def is_uniform(self, config: BaseConfig) -> bool:
                     ", 'table', or 'snapshot'."
                 )
             if (
-                self.behavior.use_managed_iceberg
+                self.get_behavior_flag_no_warn(USE_MANAGED_ICEBERG["name"])
```
sd-db, line 285: nit: this is actually safe as `is_uniform` check is called only if we are checking for iceberg compat mode.

dbt/adapters/databricks/impl.py
```diff
@@ -993,7 +1004,7 @@ def get_column_tags_from_model(self, model: RelationConfig) -> Optional[ColumnTa
     @available
     def resolve_file_format(self, config: BaseConfig) -> str:
         if config.get("table_format") == constants.ICEBERG_TABLE_FORMAT:
-            if self.behavior.use_managed_iceberg:
+            if self.get_behavior_flag_no_warn(USE_MANAGED_ICEBERG["name"]):
```
sd-db: nit: same as earlier, this is an intentional call

tests/unit/test_adapter.py
```diff
@@ -1351,3 +1352,35 @@ def test_is_uniform_with_invalid_materialization_error(
             DbtConfigError, match="When table_format is 'iceberg', materialized must be"
         ):
             adapter.is_uniform(mock_config)
+
+    def test_is_uniform_does_not_fire_managed_iceberg_event(
```
sd-db: We should remove this test, as this is testing for implementation detail on something that is user-preference. `is_uniform` path is only called if iceberg/iceberg-compat. In such a case even if we trigger with warning that is fine. So while code wise this is correct, intention wise this is probably encoding wrong assumptions

sd-db (changes requested):
Overall change looks good, recommended some changes w.r.t tests.Additionally I have made some `nit` comments in the code which are more of a FYI though and not a need to address.

tests/functional/adapter/iceberg/test_managed_iceberg_warning.py
tejassp-db: fixed.

tests/unit/test_adapter.py
tejassp-db: removed.

sd-db (approved):
Changes look good, have triggered integration tests.

PR #1430, opened 2026-04-28
fix: auto-quote backtick-rendered refs in metric_view YAML body (#1361)

sd-db:
Closes #1361.

## Bug

Bare `{{ ref('x') }}` in a `metric_view` `source:` field renders to `` `db`.`schema`.`x` ``, which the YAML scanner rejects:

```
[METRIC_VIEW_INVALID_VIEW_DEFINITION] Failed to parse YAML:
found character '`' that cannot start any token
```

## Fix

Add `SqlUtils.yaml_quote_backtick_values` — wraps backtick-rendered identifiers in mapping values with `"..."` before the body is embedded in `$$..$$`. Preserves SQL identifier escaping for non-trivial names (hyphens, reserved words). Called from the metric_view create and alter macros.

Same shape as `SqlUtils.clean_sql` and the `STREAM` strip in `streaming_table/create.sql`.

## Test plan

- [x] Unit: 10 parametrized cases (`tests/unit/test_handle.py`) — idempotence, multi-line, indented, inline backticks left alone, etc.
- [x] Functional: `TestMetricViewBareRef` runs a bare-ref metric_view + round-trips a `MEASURE()` query
- [x] Full `tests/functional/adapter/metric_views/` suite (18 tests) — no regressions
- [x] `pre-commit run --all-files` clean


tejassp-db (changes requested):
Does this parse the whole YAML file or just the `source` field? If we intend to only change `source` field can we make it more specific and not whole YAML wide.

Also the regex might need tightening. Valid identifiers as mentioned in this [identifier syntax doc](https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-identifiers). Following samples break in the current implementation

```
pytest.param(
    "source: `schema`.`name` # a comment",
    'source: "`schema`.`name` # a comment',
    id="two_part_with_comment",
),

# Very rare case, but it is valid syntax. I don't know if we should solve for this?
pytest.param(
    "source: `a``b`.`schema`.`table`",
    'source: "`a``b`.`schema`.`table`"',
    id="escaped_backtick_in_identifier",
),
```

sd-db:
> Does this parse the whole YAML file or just the `source` field? If we intend to only change `source` field can we make it more specific and not whole YAML wide.
> 

Good point, will limit this to only source field


> ```
> pytest.param(
>     "source: `schema`.`name` # a comment",
>     'source: "`schema`.`name` # a comment',
>     id="two_part_with_comment",
> ),
> ```

Yes this is something we should look to solve for


> ```
> # Very rare case, but it is valid syntax. I don't know if we should solve for this?
> pytest.param(
>     "source: `a``b`.`schema`.`table`",
>     'source: "`a``b`.`schema`.`table`"',
>     id="escaped_backtick_in_identifier",
> ),
> ```

Here pragmatically I think we should not solve for such cases and it is fine if this fails. We are specifically looking for well intentioned patterns for source ref that we want to parse for

tejassp-db (approved):

PR #1431, opened 2026-04-29
fix: suppress warning on materialization v2 behavior flags.

### LINKED ISSUE #1089: How to silence deprecation warning about materialization v2?
It is pretty annoying to constantly see 10x warnings about materialization V2, it makes it harder to read our dbt logs.

Even though we explicitly disabled the flag, it's still spamming our logs.

How do we silence the warning? I couldn't figure out how to do it from the docs.

tejassp-db:
Fixes https://github.com/databricks/dbt-databricks/issues/1089.

dbt-core's BehaviorFlag emits a deprecation warning the first time a flag with default=False is read. As a result, every dbt run with incremental materialization logs a confusing use_materialization_v2 warning on every model, with no clean way to silence it short of opting in to the flag itself.
                                                                                                                                                                                                              
This PR fixes this by accessing the flag with no_warn in the code paths. A new get_behavior_flag_no_warn helper on the adapter is decorated with @available so Jinja can call it at runtime. All four materialization macros that gate v1/v2 (table.sql, view.sql, seeds.sql, incremental.sql) now route through that helper, which reads the flag via BehaviorFlagRendered.no_warn and bypasses the BehaviorChangeEvent entirely.   

PECOBLR-2649

### DIFF
--- a/tests/functional/adapter/materialized_view_tests/test_basic.py
+++ b/tests/functional/adapter/materialized_view_tests/test_basic.py
@@ -139,14 +139,6 @@ class TestMaterializedViews(TestMaterializedViewsMixin, MaterializedViewBasic):
         )
         assert results[0][1] == expected_struct_type
 
-    def test_use_materialization_v2_warning_absent(self, project, my_table):
-        """A project on the default `use_materialization_v2` setting must not emit the
-        deprecation warning. Running a `table` model exercises the v2 check."""
-        _, logs = util.run_dbt_and_capture(["run", "--models", my_table.identifier])
-        warning_substring = USE_MATERIALIZATION_V2["description"][:40]
-        util.assert_message_in_logs(warning_substring, logs, False)
-
-
 @pytest.mark.dlt
 @pytest.mark.skip_profile("databricks_cluster", "databricks_uc_cluster")
 class TestMaterializedViewLiquidClustering(TestMaterializedViewsMixin):
--- a/tests/functional/adapter/materialized_view_tests/test_basic.py
+++ b/tests/functional/adapter/materialized_view_tests/test_basic.py
@@ -6,6 +6,7 @@ from dbt.tests import util
 from dbt.tests.adapter.materialized_view import files
 from dbt.tests.adapter.materialized_view.basic import MaterializedViewBasic
 
+from dbt.adapters.databricks.impl import USE_MATERIALIZATION_V2
 from tests.functional.adapter.materialized_view_tests import fixtures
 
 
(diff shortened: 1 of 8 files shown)


tests/functional/adapter/materialized_view_tests/test_basic.py
```diff
@@ -138,6 +139,13 @@ def test_materialized_view_complex_types(self, project):
         )
         assert results[0][1] == expected_struct_type
 
+    def test_use_materialization_v2_warning_absent(self, project, my_table):
```
sd-db: We should not add any functional test for this change. Functional check general behaviour and should not be used to check for logs/messages. This is better to be had as unit tests.

tests/unit/test_adapter.py
```diff
@@ -1351,3 +1352,39 @@ def test_is_uniform_with_invalid_materialization_error(
             DbtConfigError, match="When table_format is 'iceberg', materialized must be"
         ):
             adapter.is_uniform(mock_config)
+
+
+class TestMaterializationV2BehaviorFlag(DatabricksAdapterBase):
+    """The `use_materialization_v2` deprecation warning must not fire on every dbt run.
+    The fix routes all 4 Jinja-side accesses through `adapter.get_behavior_flag_no_warn(...)`,
```
sd-db: nit: make comment less verbose/not leak specific implementation details.

sd-db (changes requested):
Code changes look good, best to remove functional tests for the change

tests/unit/test_adapter.py
tejassp-db: fixed

tests/functional/adapter/materialized_view_tests/test_basic.py
tejassp-db: removed

tests/functional/adapter/materialized_view_tests/test_basic.py
```diff
@@ -6,6 +6,7 @@
 from dbt.tests.adapter.materialized_view import files
 from dbt.tests.adapter.materialized_view.basic import MaterializedViewBasic
 
+from dbt.adapters.databricks.impl import USE_MATERIALIZATION_V2
```
sd-db: nit: stray line

sd-db (dismissed):
Changes look good, there seems to be a stray line (_probably an artifact from the recentl fixes, that we forgot ro remove_)

tests/functional/adapter/materialized_view_tests/test_basic.py
tejassp-db: fixed

sd-db (approved):
Changes look good !!

PR #1432, opened 2026-04-29
feat: Replace information schema with describe calls

tejassp-db:
### Description                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                   
  Replace expensive `information_schema` queries with a single `DESCRIBE TABLE EXTENDED ... AS JSON` call for metadata fetching in `_describe_relation` methods. Gated behind `DBRCapability.DESCRIBE_TABLE_EXTENDED_AS_JSON` (DBR 17.3+), falls back to existing info_schema queries on older runtimes.             
   
  **Changes:**                                                                                                                                                                                                                                                                                                       
                  
  - `IncrementalTableAPI._describe_relation`: replaces 4 info_schema queries (PK, FK, non-null constraints, column masks) with AS JSON parsing                                                                                                                                                                       
  - `MaterializedViewAPI._describe_relation`: replaces `get_view_description` (info_schema.views) with AS JSON parsing
  - `ViewAPI._describe_relation`: same as MV                                                                                                                                                                                                                                                                         
  - New `DatabricksDescribeJsonMetadata` parser class with composite PK/FK support                                                                                                                                                                                                                                   
  - New `is_describe_as_json_supported()` gating method (checks HMS, foreign table, capability)
  - New `is_foreign_table` property on `DatabricksRelation`                                                                                                                                                                                                                                                          
  - New `describe_table_extended_as_json` Jinja macro
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
  **Testing:**
                                                                                                                                                                                                                                                                                                                     
  - 56 unit tests: parser, processor roundtrip, diff validation, edge cases
  - Capability boundary tests (17.2 vs 17.3)
  - `is_describe_as_json_supported` unit tests (UC, HMS, foreign table, no capability)
  - Jinja macro SQL generation test                                                                                                                                                                                                                                                                                  
  - Functional tests validated on both code paths:
    - SQL warehouse (DBR 17.3+): all 14 constraint/mask tests passed via DESCRIBE AS JSON path                                                                                                                                                                                                                       
    - UC cluster (DBR 16.4 LTS): all 14 tests passed via information_schema fallback path                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                     
  ### Checklist                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                     
PECOBLR-2546


sd-db:
On the failure,

```
FAILED tests/functional/adapter/python_model/test_python_model.py::TestSpecifyingHttpPath::test_singular_tests - AssertionError: dbt exit state did not match expected
```

this is known flaky test so can be ignored (_I will look to solve for this separately_)

dbt/adapters/databricks/impl.py
```diff
@@ -153,6 +154,17 @@
     ),
 )  # type: ignore[typeddict-item]
 
+USE_DESCRIBE_AS_JSON = BehaviorFlag(
+    name="use_describe_as_json",
```
sd-db: nit: should we rename to something like `use_describe_as_json_for_metadata`. The naming can be a bit confusing as we are already using use descibe as json even without this flag, wdyt ?

dbt/adapters/databricks/impl.py
```diff
@@ -403,6 +416,24 @@ def require_capability(self, capability: DBRCapability) -> None:
                 f"Current connection does not meet this requirement."
             )
 
+    def is_describe_as_json_supported(self, relation: DatabricksRelation) -> bool:
```
sd-db, line 421: Again I feel we should rename this, this check is specifically for using DECRIBE TABLE EXTENDED AS JSON for getting metadata we should mark it as such as there can be other places where we are using `DESCRIBE... AS JSON` that don't go through this check. Some suggestions `is_describe_as_json_relation_metadata_supported()` or maybe `can_use_describe_json_for_relation_metadata()`

dbt/adapters/databricks/impl.py
sd-db: Something like `use_describe_as_json_for_relation_metadata` maybe

dbt/adapters/databricks/impl.py
```diff
@@ -403,6 +416,24 @@ def require_capability(self, capability: DBRCapability) -> None:
...
+            and not relation.is_foreign_table
+            and self.has_capability(DBRCapability.DESCRIBE_TABLE_EXTENDED_AS_JSON)
+            and bool(self.behavior.use_describe_as_json.no_warn)
+        )
+
+    def fetch_json_metadata(self, relation: DatabricksRelation) -> dict[str, Any]:
+        """Fetch the JSON metadata for a relation using DESCRIBE TABLE EXTENDED AS JSON."""
+        kwargs = {"relation": relation}
+        describe_results = self.execute_macro("describe_table_extended_as_json", kwargs=kwargs)
+        json_metadata = json.loads(describe_results.rows[0].get("json_metadata"))
```
sd-db: There is a chance for this to error out, we should raise a `DbtRuntimeError` in-case the macro fails and the result is not as per expectations. This assumes that the query always returns a success and can raise any sort of `IndexError`, `TypeError` or even `JSONDecodeError`.

Once we add more defensive code + more specific error codes here also look to add tests.

dbt/adapters/databricks/impl.py
```diff
@@ -1239,3 +1302,372 @@ def _describe_relation(
...
+            non_null_constraints=cls.parse_non_null_constraints(json_metadata),
+            primary_key_constraints=cls.parse_primary_key_constraints(json_metadata),
+            row_filters=cls.parse_row_filter(json_metadata),
+            view_description=cls.parse_view_description(json_metadata),
+        )
+
+    @classmethod
+    def parse_column_masks(cls, json_metadata: dict[str, Any]) -> "Table":
+        """Parse json metadata into an agate Table of column masks (info_schema format)."""
+        # Lazy load to improve startup time
```
sd-db, line 1423: Any reason for this optimisation ? Is it because we don't want to call the modules when user does not have the capability active ? I don't think module import is particularly expensive and maybe best to have this at the top of the file for simplicity

dbt/adapters/databricks/impl.py
```diff
@@ -1239,3 +1302,372 @@ def _describe_relation(
...
+            column_masks=cls.parse_column_masks(json_metadata),
+            foreign_key_constraints=cls.parse_foreign_key_constraints(json_metadata),
+            non_null_constraints=cls.parse_non_null_constraints(json_metadata),
+            primary_key_constraints=cls.parse_primary_key_constraints(json_metadata),
+            row_filters=cls.parse_row_filter(json_metadata),
+            view_description=cls.parse_view_description(json_metadata),
+        )
+
+    @classmethod
+    def parse_column_masks(cls, json_metadata: dict[str, Any]) -> "Table":
```
sd-db, line 1409: For all the parse methods we are heavily relying on the structure of the reponse from the `DESCRIBE...JSON` call. While this works. I feel it might have been much better to define our out class/data-model and load the results from the `DESCRIBE...JSON` call into the data-model. Basically a DAO layer. Here we are at the mercy of the server and whatever validations we might/might not add on correctness and there can still be edge-cases.

sd-db (commented):
PR Review 1/x. Added a few comments on structure

dbt/adapters/databricks/impl.py
```diff
@@ -1239,3 +1302,372 @@ def _describe_relation(
...
+        """Parse json metadata into an agate Table of column masks (info_schema format)."""
```
sd-db, line 1422: I am also thinking if there are a few other things we can do here. I see thaat with parsing (and without a proper data-model) there can be exceptions. Can you wrap all the parsing functions inside a `try ... except` block and raise descriptive `DbtRuntimeError`, something like

```
...
except Exception as e:
    raise DbtRuntimeError(f"Failed to parse column masks: {e}")
```

I would rather have the model fail with a descriptive error on whenever there is a parsing error

tejassp-db: Fixed.

dbt/adapters/databricks/impl.py
tejassp-db: Sticking to earlier conventions.

dbt/adapters/databricks/impl.py
tejassp-db: This method specifically is to check whether "DESCRIBE TABLE EXTENDED AS JSON" is supported. The method should not take responsibility of how the call response is consumed.

dbt/adapters/databricks/impl.py
tejassp-db: Fixed.

dbt/adapters/databricks/impl.py
tejassp-db: Future.

dbt/adapters/databricks/impl.py
tejassp-db: Renamed to `use_describe_as_json_for_relation_metadata`.

dbt/adapters/databricks/impl.py
sd-db: The reason for raising this was actually for the behaviour flag check --> `bool(self.behavior.use_describe_as_json_for_relation_metadata.no_warn)`; since this is dependent on the flag it is a more specific version/flavour, we can look to either move this out or rename

sd-db (commented):
Overall looks good, made some minor comments

sd-db (approved):

tejassp-db:
> Integration results for PR #1432 — UC cluster ✅ success · SQL warehouse ❌ failure · All-purpose cluster ✅ success · Shard coverage ✅ success
> 
> [Run details](https://github.com/databricks/dbt-databricks/actions/runs/25847514027).

Flaky integration test, this code path does not touch those tests or their code path.

PR #1435, opened 2026-04-30
fix: missing metadata handling in is_hudi and is_iceberg

aarushisingh04:
### description

`can_be_replaced` calls `is_iceberg` without first checking whether metadata is available. on relations that haven't fetched extended metadata yet (`metadata=None`) this hits an unconditional `assert self.metadata is not None` and crashes the dbt run with an `AssertionError`.

the fix 

- updates `is_hudi` and `is_iceberg` to safely return `False` when `self.metadata is None`, making their behavior consistent with other metadata properties like `owner` and `stats`. 
- unit tests have also been added to ensure properties evaluate gracefully on un-hydrated relations.

### checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixes
 
+- Fix latent AssertionError when evaluating `is_hudi` or `is_iceberg` properties on relations without metadata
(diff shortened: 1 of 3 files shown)


sd-db (approved):
Changes look good, thanks for the PR!

PR #1438, opened 2026-05-04
test: faster integration test run

sd-db:
## Summary

Cuts integration-test wall-clock time by **~64%** through matrix sharding + smarter shard assignment + automatic flake retry.

## Baseline (latest green main: run)

| profile | main baseline | this PR (exp-11) | change |
|---|---|---|---|
| cluster | 74.7 min | **27.4 min** (max shard) | **−63%** |
| uc-cluster | 99.7 min | **45.7 min** (max shard) | **−54%** |
| sql warehouse | 125.7 min | **31.7 min** (max shard) | **−75%** |
| **total wall** (slowest profile) | **125.7 min** | **45.7 min** | **−64%** |

Total wall = the longest job, since the 3 profiles run in parallel.

## What changed

- **Matrix sharding**: cluster split 1→2 cells, uc-cluster + sqlw split 1→3 cells (8 parallel test cells total).
- **Smart assignment** (`scripts/shard_assign.py`): greedy LPT on per-file historical wall-time (default), with `lpt_test_count` fallback.
- **dbt threads bump**: test profile from `1` → `4`.
- **Flake retry**: `pytest-rerunfailures` with `--reruns 1 --reruns-delay 120` mitigates a known UC-cluster driver flake under sharded concurrency.

## Test plan

- [x] Compare cluster / sql warehouse / uc-cluster durations against latest baseline run on `main`


scripts/shard_assign.py
```diff
@@ -0,0 +1,211 @@
...
+            all_nodeids.append(nodeid)
+
+    if not all_nodeids:
+        print(f"ERROR: no valid nodeids in {input_path}", file=sys.stderr)
+        return 2
+
+    sorted_files = sorted(file_to_tests.keys())
+
+    if args.algo == "lpt_historical_time":
+        if not args.timings:
```
tejassp-db, line 135: Should we continue when these files are missing?

pyproject.toml
```diff
@@ -69,6 +69,7 @@ dependencies = [
     "dbt-tests-adapter",
     "pytest",
     "pytest-xdist",
+    "pytest-rerunfailures",
```
tejassp-db, line 72: Should we check in the uv.lock file as well?

.github/test_timings.json
```diff
@@ -0,0 +1,275 @@
+{
```
tejassp-db, line 1: Should this file be created on a schedule?

tests/profiles.py
```diff
@@ -32,6 +32,7 @@ def _build_databricks_cluster_target(
         "connect_timeout": 5,
         "retry_all": True,
         "auth_type": os.getenv("DBT_DATABRICKS_AUTH_TYPE", "oauth"),
+        "threads": 4,
```
tejassp-db, line 35: Can our warehouses handle the load?

sd-db: Yes, warehouse load is minimal, changing `threads = 4` is a very safe change and did not result in any observable difference in metrics on the warehouse side

.github/test_timings.json
sd-db: Yes that is the plan in the long run. I plan to add another girhub action that will run on a schedule and create this file. For now the current file works and should work even if new tests are added as new test that does not have a timing get assigned the average (baseline) time

tejassp-db (approved):

pyproject.toml
sd-db: `pytest-rerunfailures` is a dev-dependency hence it is not added to `uv.lock` (same with the other dev dependencies like `pytest-xdist` etc)

scripts/shard_assign.py
sd-db: I think erroring out here is fair if the files are indeed missing (fast-failure)

PR #1440, opened 2026-05-04
feat: catalogs.yml v2 support via bridge_v2_catalog hooks

aahel:
## Summary

Adds catalogs.yml v2 support to dbt-databricks using the adapter-owned bridge architecture.

### What changed

**`impl.py`**
- Declares `Capability.CatalogsV2` — enables v2 path when `use_catalogs_v2: true` is set in `dbt_project.yml`
- Adds `_v2_to_v1_type` hook: `unity → unity`, `hive_metastore → hive_metastore`

**`catalogs/_unity.py`**
- Adeed warnings if `use_uniform` is set in config properties
- `location_root` must be non-blank if provided

**`catalogs/_hive_metastore.py`**
- Adds `file_format` enum validation in `HiveMetastoreCatalogIntegration.__init__`: must be one of `delta | parquet | hudi`

### Architecture

dbt-core parses `catalogs.yml` v2 structurally, checks `Capability.CatalogsV2`, then calls `adapter.bridge_v2_catalog(catalog)`. The base adapter template method reads `catalog.config.get("databricks", {})` and calls `_v2_to_v1_type` to resolve the v1 catalog_type. Field-level validation happens in `CatalogIntegration.__init__` — same as v1, no new interface.

Companion PRs:
- dbt-core: https://github.com/dbt-labs/dbt-core/pull/12930
- dbt-adapters: https://github.com/dbt-labs/dbt-adapters/pull/1920

## Test plan

- [x] Unit tests for `UnityCatalogIntegration.__init__` validation
- [x] Unit tests for `HiveMetastoreCatalogIntegration.__init__` validation
- [ ] Functional tests end-to-end once dbt-core and dbt-adapters PRs are merged


CHANGELOG.md
```diff
@@ -55,6 +55,7 @@
 
 ### Features
 
+- Add catalogs.yml v2 support (requires `use_catalogs_v2: true` in dbt-core)
```
sd-db: Let us move the changelog entry under `v1.22.2`

dbt/adapters/databricks/catalogs/_unity.py
```diff
@@ -13,9 +14,27 @@ class UnityCatalogIntegration(CatalogIntegration):
...
         self.file_format: Optional[str] = config.file_format
+        use_uniform = config.adapter_properties.get("use_uniform")
+        if use_uniform is not None:
+            ff = (config.file_format or "").lower()
+            if use_uniform and ff != "delta":
+                raise DbtValidationError(
+                    f"Catalog '{config.name}' unity/databricks use_uniform: true "
+                    f"requires file_format: delta"
+                )
+            if not use_uniform and ff and ff != "parquet":
```
sd-db: I think this check is overly restrivtive and can raise in incorrect scenario of a `default` table_format (not use_uniform) and underlying ff is `delta` which errors out here but is valid config.

```
table_format: default
config:
  databricks:
    use_uniform: false
    file_format: delta
```

Can you detail what we are looking to check for here ?

dbt/adapters/databricks/catalogs/_unity.py
```diff
@@ -13,9 +14,27 @@ class UnityCatalogIntegration(CatalogIntegration):
...
-        if location_root := config.adapter_properties.get("location_root"):
+        location_root = config.adapter_properties.get("location_root")
+        if location_root is not None:
+            if not str(location_root).strip():
+                raise DbtValidationError(
+                    f"Catalog '{config.name}' unity/databricks location_root cannot be blank"
+                )
             self.external_volume: Optional[str] = location_root
         self.file_format: Optional[str] = config.file_format
+        use_uniform = config.adapter_properties.get("use_uniform")
```
sd-db: I don't think the runtime respects the catalog level property `use_uniform`. Today if something is managed_iceberg or not is decided by a combination of `table_format`(tf) + `use_managed_iceberg` behaviour flag

- `tf:iceberg` + `use_managed_iceberg:False` --> uniform
- `tf:iceberg` + `use_managed_iceberg:True` --> managed iceberg

I think if we are adding validation here before that we need to add support for the flag as well

dbt/adapters/databricks/catalogs/_unity.py
aahel: you wouldnt configure use_uniform for no iceberg table formats right ? if use_uniform is not set with table_format: default it wont raise

aahel: would you prefer to not throw an error and ignore the non iceberg case ?
something like:
```
        if use_uniform is not None and (config.table_format or "").lower() == "iceberg":
```

dbt/adapters/databricks/catalogs/_unity.py
aahel: got it. 

removing the use_uniform/file_format cross-validations

dbt/adapters/databricks/impl.py
```diff
@@ -1910,3 +1917,8 @@ def _extract_backticked(cls, s: str) -> list[str]:
               returns ["cat", "sch", "tbl", "a", "b"]
         """
         return [m.group(1).replace("``", "`") for m in cls._BACKTICKED.finditer(s)]
+
```
sd-db: nit: can move this to the top. Better to have side-effect at the top for readability

dbt/adapters/databricks/catalogs/_unity.py
sd-db: Resolved, thx !

sd-db (approved):
Changes look good, thx for the PR !

PR #1444, opened 2026-05-07
fix: missing column comments after ALTER VIEW AS

canbekley:
Resolves #1357                                 

Split out from #1358 per maintainer request to land independently of the #1356 behavior change.   

### Description                                                                                   
                     
When `view_update_via_alter` is enabled and the alter path issues `ALTER VIEW <view> AS <query>`, 
Databricks silently drops every persisted column comment. The existing `if column_comments` block
in `databricks__alter_view` doesn't catch this because the column-comment diff is computed against
 the pre-alter state — when model and existing comments match, the diff is empty, so nothing
reapplies them after `ALTER VIEW AS` wipes them.

Fix: inside the `query` branch of `databricks__alter_view`, reapply persisted column comments     
after `alter_query`, gated on `config.persist_column_docs()` and `model.columns`. The pattern
mirrors `databricks__persist_docs` (`dbt/include/databricks/macros/adapters/persist_docs.sql`).   
                
The existing `if column_comments` block is kept — it still handles the disjoint case where only a 
column description changed in `schema.yml` (no query change, so the new reapply path doesn't run).
                                                                                                  
This fix becomes load-bearing once #1356 lands and `ALTER VIEW AS` runs on every invocation, but  
the bug exists today on any query change and is worth fixing standalone.
                                                                                                  
### Checklist   

- [x] I have updated the `CHANGELOG.md` and added information about my change to the              
"dbt-databricks next" section.


sd-db (approved):

Issue #1331, opened 2026-02-23
has_dbr_capability not evaluated properly with SQL warehouses


nicor88:
### Describe the bug
When bumping the adapter to dbt-databricks v1.11.5, I started to receive this error:

```
Compilation Error in model my_model (models/my_model.sql)
  macro dateadd not implemented for datepart ~ 'DAY' ~ on Spark
  
  > in macro spark__dateadd (macros/utils/dateadd.sql)
  > called by macro databricks__dateadd (macros/utils/dateadd.sql)
  > called by macro dateadd (macros/utils/dateadd.sql)
  > called by macro default__date_spine (macros/sql/date_spine.sql)
  > called by macro date_spine (macros/sql/date_spine.sql)
```
I started to get suspiscious, because the adatper was calling some Spark macros.

After further investigation, the culprit seems to be on the `{%- if adapter.has_dbr_capability('timestampdiff') -%}` was evaluated to false, and the adapter was falling back to `spark__dateadd` leading to this error.

Overriding the macro to
```
{% macro databricks__dateadd(datepart, interval, from_date_or_timestamp) %}
    timestampadd({{datepart}}, {{interval}}, {{from_date_or_timestamp}})
{%- endmacro %}
```
fix the issue. 

**I'm using a SQL serverless Warehouse** and I believe that there is a bug on how the *has_dbr_capability* is evaluated.

### Steps To Reproduce
Create a model that uses
```
with date_spine AS (

    SELECT *
    FROM (

        {{
            dbt_utils.date_spine(
                datepart="DAY",
                start_date="CAST('2024-01-01' AS DATE)",
                end_date="CAST(CONVERT_TIMEZONE('Europe/Berlin', CURRENT_TIMESTAMP()) AS DATE)"
            )
        }} -- end_date handle snowflake config time being in LA time

    )

)

select * from date_spine

```

Should lead the macro databricks__dateadd to be invoked

### Expected behavior
`adapter.has_dbr_capability('timestampdiff')`  should be evaluated to true for SQL serverless warehouses

### Screenshots and log output
Provided above inside the bug description

### System information
**The output of `dbt --version`:**
```
Core:
  - installed: 1.11.6
  - latest:    1.11.6 - Up to date!

Plugins:
  - databricks: 1.11.5 - Up to date!
  - spark:      1.10.1 - Up to date!
```

**The operating system you're using:**  MacOS 26.3 / Linux via Docker (python:3.12-slim-bullseye)

[...]

PR #1449, opened 2026-05-11
fix: return True from has_dbr_capability parse stub on SQL warehouses (#1331)

sd-db:
Motivated by [#1331](https://github.com/databricks/dbt-databricks/issues/1331).

The class-level `@available.parse(lambda *a, **k: False)` stub on `has_dbr_capability` returns a constant `False` at parse time, so every macro branching on a capability takes the legacy path during parse/compile — even on SQL warehouses, which support all capabilities we currently ship.

This is a 1.11.0 regression: the pre-1.11 stub (`compare_dbr_version`) returned `0`, and `0 >= 0` accidentally selected the modern path.

## Fix

Shadow `_parse_replacements_` at adapter `__init__` with a bound method that consults the profile's `http_path`:

- **SQL warehouse** → `True` for capabilities with `sql_warehouse_supported=True`.
- **Cluster** → `False` (real DBR version is only known at runtime).

Cluster users on supported DBR are still affected — fully fixing that requires a larger change. This PR is the contained warehouse-only fix.

## Test plan

- [x] `hatch run pytest tests/unit/test_adapter_capabilities.py -v` — 19 pass (4 new).
- [x] `hatch run pre-commit run --all-files` — clean.
- [ ] CI functional run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Fixes
 
 - Use pydantic v1-compatible API in `refresh.py` so the adapter works on environments shipping pydantic v1.
+- Fix capability-branching macros falling through to their legacy path at parse/compile time on SQL warehouses. The parse-time stub of `has_dbr_capability` now returns `True` on warehouse profiles for capabilities flagged `sql_warehouse_supported`, so macros select the modern branch during compilation instead of the legacy fallback. ([#1331](https://github.com/databricks/dbt-databricks/issues/1331))
 
 ## dbt-databricks 1.12.0 (May 14, 2026)
 
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -51,7 +51,7 @@ from dbt.adapters.databricks.connections import (
     DatabricksConnectionManager,
     DatabricksDBTConnection,
 )
-from dbt.adapters.databricks.dbr_capabilities import DBRCapability
+from dbt.adapters.databricks.dbr_capabilities import DBRCapabilities, DBRCapability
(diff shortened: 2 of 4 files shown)


nicor88:
@sd-db you can consider to add real functional tests in which for example a date-spine function from dbt-utils is used to check if everything actually works fine with a real warehouse, alternatively something artifically created like date-spine to test a real scenario. The unit tests cover some relevant aspects, but proper integration testings allows to spot regressions in a more solid way.

sd-db:
> @sd-db you can consider to add real functional tests in which for example a date-spine function from dbt-utils is used to check if everything actually works fine with a real warehouse, alternatively something artifically created like date-spine to test a real scenario. The unit tests cover some relevant aspects, but proper integration testings allows to spot regressions in a more solid way.

Good call — added a functional test in `tests/functional/adapter/utils/test_dateadd_uppercase.py` (commit d6fe3dd2).

CHANGELOG.md
```diff
@@ -9,7 +9,11 @@
 
 ### Fixes
 
+<<<<<<< sd-db/fix/issue-1331-warehouse-only-parse-stub
+- Fix capability-branching macros falling through to their legacy path at parse/compile time on SQL warehouses. The parse-time stub of `has_dbr_capability` now returns `True` on warehouse profiles for capabilities flagged `sql_warehouse_supported`, so macros select the modern branch during compilation instead of the legacy fallback. ([#1331](https://github.com/databricks/dbt-databricks/issues/1331))
+=======
```
tejassp-db: Fix merge conflicts.

sd-db: Fixed

jprakash-db (approved):

PR #1453, opened 2026-05-12
Feature/957 more auth mechanisms

### LINKED ISSUE #957: Support for more authentication mechanisms
### Describe the feature

Thanks to the work done in #827, `dbt-databricks` systematically uses Databricks' `databricks.sdk.core.Config` to configure [authentication mechanisms](https://github.com/databricks/dbt-databricks/blob/1.10.latest/dbt/adapters/databricks/credentials.py).

As of now, only a limited subset (PAT, OAuth U2M, Oauth M2M and Azure client/secret) of the authentication mechanisms are exposed. However the new implementation makes it really easy to extend it further based on all the [options offered by Databricks' SDK for Python](https://databricks-sdk-py.readthedocs.io/en/latest/authentication.html) which follows itself Databricks unified authentication.

I am very interested in implementing it for Azure CLI and Azure Managed Identity and would love to submit a PR.

- The first alternative would be to follow the current pattern in https://github.com/databricks/dbt-databricks/blob/1.10.latest/dbt/adapters/databricks/credentials.py while trying to stick as closely as possible to Databricks unified authentication behavior/variable names.
- The second alternative would be the one I mentioned in an [earlier comment](https://github.com/databricks/dbt-databricks/pull/827#discussion_r1889952643) to pass transparently the options to Databricks' SDK for Python in order to maximize the compatibility with all the options availables and be able to quickly deliver newer options as they're implemented. I would then make sure to not break the existing functionality by keeping the existing logic and only passing the options through to Databricks SDK if none of the existing options matched the configuration.

### Who will this benefit?
Any user who wants to leverage more authentication options, and potentially increase security.
For the project I am currently working on:¨
- `dbt-databricks` is the last component not leveraging Azure CLI authentication in the development environment.
- It is also the only component preventing us from deactivating Databricks PAT
- In Azure, we already rely extensively on managed identity, but have to fallback to client id/secret for `dbt-databricks`

Such a change would therefore be a significant improvement both in terms of convenience and security.

Happy to implement it 😃

ghjklw:
## What this PR does

Closes [databricks/dbt-databricks#957](https://github.com/databricks/dbt-databricks/issues/957).

**Problem:** The adapter's authentication layer had two classes of issues:
1. Many valid Databricks SDK authentication methods (`azure-cli`, `azure-msi`, `databricks-cli`, `metadata-service`, etc.) were silently blocked — `validate_creds()` required either a token or the legacy `auth_type: oauth` alias, and any other `auth_type` value was ignored rather than forwarded to the SDK.
2. Two profile fields — `oauth_scopes` and `oauth_redirect_url` — behaved incorrectly (see [OAuth bugs](#oauth-scopesredirect_url-bugs) below).

**Solution:** Auth dispatch is simplified to two branches: the existing heuristic (preserved for backward compatibility but now **deprecated with a runtime warning**) and a pass-through to the Databricks SDK `Config` for everything else. Users still on the heuristic path will see a deprecation warning directing them to set `auth_type` explicitly. This makes every auth method the SDK supports available to users without requiring changes to the adapter for each new method.

---

## Authentication type reference

The table below covers every profile configuration. Rows marked **unchanged** behave identically to the previous release; rows marked **fixed** or **new** indicate a behavioral change.

| Auth scenario | Profile fields | `auth_type` | Status |
|---|---|---|---|
| **Personal Access Token (PAT)** | `token` | _(not required)_ | Unchanged |
| **Azure Service Principal (dedicated fields)** | `azure_client_id` + `azure_client_secret` | `azure-client-secret` (inferred) | **Fixed** — `azure_tenant_id` now forwarded (was silently dropped) |
| **OAuth M2M (Databricks)** | `client_id` + `client_secret` starting with `dose` | _(not required)_ | Unchanged — heuristic tries `oauth-m2m` first. **Deprecated:** set `auth_type: oauth-m2m` explicitly |
| **Azure Service Principal (legacy fields)** | `client_id` + `client_secret` (non-`dose`) | _(not required)_ | Unchanged — heuristic tries `azure-client-secret` first. **Deprecated:** set `auth_type: azure-client-secret` explicitly |
| **OAuth U2M (browser)** | _(none)_ | `oauth` or `external-browser` | Functionally unchanged — empty `client_secret=""` no longer leaked into SDK Config |
| **OAuth U2M with custom app** | `client_id` | `oauth` / `external-browser` | Functionally unchanged — empty `client_secret=""` no longer leaked |
| **OAuth U2M inferred from `client_id` alone** | `client_id` (no `client_secret`) | `external-browser` (inferred) | **New** — `client_id` without `client_secret` now selects browser-based OAuth |
| **Azure CLI** | _(none)_ | `azure-cli` | **New** — previously blocked by `validate_creds()` and silently ignored |
| **Azure Managed Identity (system-assigned)** | _(none)_ | `azure-msi` | **New** |
| **Azure Managed Identity (user-assigned)** | `azure_client_id` | `azure-msi` | **New** |
| **Databricks CLI / config file** | `databricks_cli_profile`, `config_file` _(optional)_ | `databricks-cli` | **New** |
| **Google credentials** | `google_credentials` or `google_service_account` | `google-credentials` | **New** |
| **Instance metadata service** | `metadata_service_url` _(optional)_ | `metadata-service` | **New** |
| **OIDC token** | `oidc_token_env` or `oidc_token_filepath` | _(SDK-inferred)_ | **New** |
| **Basic auth** | `username` + `password` | _(SDK-inferred)_ | **New** |
| **Any future SDK auth type** | _(varies)_ | _(any SDK value)_ | **New** — forwarded automatically; no adapter changes needed |
| **SDK escape hatch** | `databricks_sdk_parameters: {key: value}` | _(any)_ | **New** — arbitrary SDK `Config` kwargs, merged last, override everything above |

### Additional field-forwarding fixes (previously silently dropped)

| Profile field | Previous behaviour | Now |
|---|---|---|
| `auth_type` when `token` is also set | `auth_type` ignored; only PAT used | Both forwarded — SDK decides precedence |
| `azure_tenant_id` with `azure_client_id`/`azure_client_secret` | Not forwarded to `Config` | Forwarded |
| `databricks_sdk_parameters` | Accepted in profile but discarded | Merged into `Config` kwargs for all auth paths |
| Explicit `auth_type` with `azure_client_id`/`azure_client_secret` | `auth_type` ignored; forced to `azure-client-secret` | Explicit `auth_type` wins |

### Removed field

| Field | Previous behaviour | Now |
|---|---|---|
| `oauth_redirect_url` | Accepted in profiles but had no effect | **Removed** — the Databricks SDK hardcodes `http://localhost:8020` internally and never reads this value from `Config`; accepting it created a false impression of control |

---

## OAuth `scopes`/`redirect_url` bugs

Two fields were silently broken in the previous implementation:

### `oauth_scopes` was never forwarded to the SDK

`DatabricksCredentialManager` was initialised with the `oauth_scopes` value and a default of `["all-apis", "offline_access"]` was applied. However, none of the `Config(...)` calls — `authenticate_with_external_browser()`, `authenticate_with_oauth_m2m()`, or any other path — included `scopes=...`. The field was accepted, stored, and then discarded. Users who set `oauth_scopes` in their profile to restrict or extend token scopes were silently ignored.

**Fix:** `oauth_scopes` is now passed as `scopes` to `Config` for all SDK-delegated auth paths.

### `oauth_redirect_url` had no effect

`REDIRECT_URL = "http://localhost:8020"` was defined as a module constant and used as the default for `DatabricksCredentialManager.oauth_redirect_url`. The Databricks SDK's `Config` class does not accept a redirect URL parameter — it hardcodes `http://localhost:8020` internally. Any user who configured `oauth_redirect_url` in their profile to point to a different port or host was silently ignored, and the browser OAuth flow continued to use `localhost:8020` regardless.

**Fix:** The field is removed entirely to avoid the false impression that it is configurable.

---

## Test strategy

This PR was developed test-first to make the behavioral delta between old and new explicit and reviewable:

**Step 1 — Baseline documentation** (`a31eec8`): Before any code change, a parametrized `TestAuthDispatch.test_config_kwargs` suite was written to document the exact `Config(...)` kwargs produced by every legacy dispatch path (7 cases: PAT, Azure SP with dedicated fields, dose-prefix heuristic, non-dose heuristic, `oauth` alias, `oauth` with custom `client_id`, and no credentials). A `TestValidateCreds` suite (6 cases) pinned validation rules. All 13 cases passed against `origin/main`, confirming the baseline was an accurate snapshot of existing behavior.

**Step 2 — Refactor applied** (`e57d441`): The dispatch was simplified. Exactly **2 of the 13 baseline tests began failing**, making the behavioral delta explicit: both failing cases were the external-browser paths where an empty `client_secret=""` was previously leaked into `Config`. This made the scope of the change fully visible before any test was updated.

**Step 3 — Tests updated and new coverage added** (`daff8f9`): The 2 failing cases were updated to reflect the corrected behavior (no empty `client_secret`). **15 new parametrized cases** were added covering: all new `auth_type` passthroughs (`azure-cli`, `azure-msi`, `databricks-cli`, `google-credentials`, `metadata-service`), field-forwarding fixes (`azure_tenant_id`, `databricks_sdk_parameters`, `token`+`auth_type` combined, explicit `auth_type` overriding field inference), and new profile fields. **5 new `TestValidateCreds` cases** confirm that previously-blocked SDK `auth_type` values now pass validation.

**Step 4 — `oauth_scopes` coverage** (`eda9a78`): 2 additional cases assert that `oauth_scopes` is forwarded as `scopes` for both `external-browser` (U2M) and `oauth-m2m` (M2M) flows.

**Step 5 — Deprecation warning and README** (`b5cf6b6`): The legacy heuristic now emits a `logger.warning` at runtime directing users to set `auth_type` explicitly. `_resolve_client_secret_heuristic` is documented as deprecated and notes that extra fields (e.g. `azure_tenant_id`, `oauth_scopes`) are intentionally not forwarded in the heuristic path to preserve identical legacy behaviour. `README.md` is updated: priority table order corrected (heuristic rows last, flagged as deprecated), `client_id`-only browser OAuth inference documented, and a migration callout block added.

**Final state:** 24 parametrized dispatch cases + 11 validation cases, all passing. No functional tests required changes — the connector-level behavior (how connections are opened against Databricks) is unchanged; only the `Config` kwargs constructed from profile fields change.

---

### Checklist


Alex-Teodosiu:
Hi folks, gentle ping on this PR when you have a chance. This enhancement would be really useful for my use case. Would one of you be able to take a look when you have time? Thanks!
@sd-db @tejassp-db @benc-db @jprakash-db

sd-db:
Hi @Alex-Teodosiu , this requires a lot more testing and something we are prioritising this month. What are the auth methods that are important to you, which we don't support ? We are looking at Github issues/comments to decide what methods to add support for

jayctran:
@sd-db azure-cli and azure managed identity type auth would be useful for me.

comegrant:
Support for Azure MSI auth would be a great security improvement for us.

ghjklw:
> Hi @Alex-Teodosiu , this requires a lot more testing and something we are prioritising this month. What are the auth methods that are important to you, which we don't support ? We are looking at Github issues/comments to decide what methods to add support for

@sd-db, let me know if there's anything I can do to help. A first step, if you want, might be to merge only the first commit which does nothing but adding more test cases, in order to make this PR safer. This way we could avoid modifying the test and logic at the same time 😉

ghjklw:
Does it make sense for me to resolve the (non-trivial) merge conflicts, or do you want to implement it in a different way?

sd-db:
Hi @ghjklw we might have to park the PR for a while. With `core-v2` out we want to ensure that any changes in v1 are able to be ported to v2 as well and auth is one area where the v2 behaviour is currently not that well defined. Hence we (dbt-labs + databricks) would like to cement the current behaviour first and then only look to add more methods/options here cc @xuliangs

PR #1463, opened 2026-05-15
fix(refresh): use pydantic v1-compatible API for adapter import

sd-db:
## Summary

`refresh.py` uses `@model_validator`, `model_dump`, and `model_construct` — all pydantic v2-only — but `pyproject.toml` declares `pydantic>=1.10.0, <2.13.0`. Adapter import fails on any environment running pydantic 1.x.

Switches to APIs that work on both v1 and v2: `@root_validator(skip_on_failure=True)` and `copy(update={...})`.

## Test plan

- [x] Unit tests pass
- [ ] Functional tests against MV/ST refresh schedules

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


sd-db:
Closing as duplicate of #1461 — I missed the existing PR. Same fix commits, same head ref content; #1461 is canonical.

PR #1465, opened 2026-05-15
fix(dependencies): update databricks-sdk version to <0.109.0 

lucasarruda-ciandt:
Raises the `databricks-sdk` upper bound from `<0.78.0` to `<0.109.0` (latest: `0.108.0`) and updates `uv.lock` to pin against `0.108.0`.

No code changes were required — all SDK APIs used by the adapter (`Config`, `CredentialsProvider`, `_header_factory`, and all service imports) remain compatible across this version range.

### Checklist

- [x] This PR includes tests, or tests are not required/relevant for this PR — no code changes; all 858 existing unit tests pass against `databricks-sdk==0.108.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## dbt-databricks 1.12.1 (TBD)
 
+### Under the Hood
+
+- Raise `databricks-sdk` upper bound to `<0.109.0` to support the latest SDK release.
+
 ### Fixes
 
 - Use pydantic v1-compatible API in `refresh.py` so the adapter works on environments shipping pydantic v1.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
 ]
 dependencies = [
     "click>=8.2.0, <9.0.0",
-    "databricks-sdk>=0.68.0, <0.78.0",
+    "databricks-sdk>=0.68.0, <0.109.0",
     "databricks-sql-connector[pyarrow]>=4.1.1, <4.1.6",
     "dbt-adapters>=1.22.0, <1.23.0",
     "dbt-common>=1.37.0, <1.38.0",
--- a/uv.lock
+++ b/uv.lock
@@ -350,16 +350,16 @@ wheels = [
 
 [[package]]
 name = "databricks-sdk"
-version = "0.77.0"
+version = "0.108.0"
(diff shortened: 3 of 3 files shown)


sd-db:
Hi @lucasarruda-ciandt would this work - https://github.com/databricks/dbt-databricks/pull/1474 ?

lucasarruda-ciandt:
if latest isn't feasible, at least is a lot better than what it is now. 
thanks for considering!

sd-db:
> if latest isn't feasible, at least is a lot better than what it is now.

Is there any feature you would requite explicitly from `>= 0.1.5 - <0.109`. I can then look at it. Also wanted to ask if you can confirm that all tests passed in your build. I did see test failures (expected) and did some test changes.

sd-db:
> Is there any feature you would requite explicitly from >= 0.1.5 - <0.109. I can then look at it

Hi @lucasarruda-ciandt , let me know. Will close the PR and we can upgrade later if not urgent.

lucasarruda-ciandt:
Hi there! My mistake here, I totally relied on IA telling me tests have passed, but after you said I ran by myself and realized a few indeed failed. Btw the tests are hanging at 95% here, tried twice... haven't looked further yet. Python 3.10 (as per pyproject.toml)
image

lucasarruda-ciandt:
> Is there any feature you would requite explicitly from `>= 0.1.5 - <0.109`. I can then look at it. Also wanted to ask if you can confirm that all tests passed in your build. I did see test failures (expected) and did some test changes.

I maintain a private library which heavily uses databricks-sdk and recently we introduced dbt-databricks as a dev dependency and it caused a downgrade on the former due to current constraints. I don't know which features would brake 'cause I haven't tested thoroughly but i'm pretty sure it wouldn't work as often use databricks alpha features (sometimes we even have to patch databricks-sdk by ourselves bc their API has evolved and they haven't synced to the sdk yet)... 

I'm used to keep dependencies as up to date as possible so I appreciate you bumping up that to the latest you can, thanks a lot!

sd-db:
> I'm used to keep dependencies as up to date as possible so I appreciate you bumping up that to the latest you can, thanks a lot!

Got it, 0.105 is fairly recent so I will keep that. The sdk is also in beta which is what makes the upgrade cycle difficult. I have resolved a fair bit of errors on supported both old + new versions and hopefully this enables easier future upgrades as well (_the current one was a large jump on 30+ versions_). We can keep on updating on newer releases.

PR #1469, opened 2026-05-20
fix: remove password-length from credential logs and harden view definition parsing

tomaioo:
## Summary

Two small fixes: stop logging password-length metadata in `CredentialShardEvent`, and make view-definition parsing more robust against empty or malformed DESCRIBE output.

### `credential_events.py` — stop logging password metadata

`CredentialShardEvent` previously logged password character count. That metadata can aid offline guessing even though the password itself is not logged. The event message is now a static `"Sharding credentials"` string with no password-related fields.

**Scope note:** `CredentialLoadError` and `CredentialSaveError` are unchanged in this PR. Redacting file paths in those error messages is a separate follow-up if needed.

### `query.py` — defensive view-definition parsing

- Use `startswith`/`endswith` instead of index access when stripping parentheses from `view_definition`, avoiding `IndexError` on empty strings.
- Guard against malformed `"View Text"` rows in `DescribeQueryProcessor` with a clearer `DbtRuntimeError`.

These are robustness improvements to metadata parsing, not SQL injection mitigation.


sd-db (approved):
Looks good, thx for the PR !!

Issue #1342, opened 2026-03-04
Primary key constraint applied without contract


dbeatty10:
### Describe the bug
Constraints should only be applied for enforced contracts, but after https://github.com/databricks/dbt-databricks/pull/1296 they are also being added when a contract is not enforced.

When there are `NULL` values in the column specified as a primary key constraint, this will cause builds to fail, which is a regression from previous behavior.

### Steps To Reproduce

As shared by @chadw56 in https://github.com/dbt-labs/dbt-core/issues/12400#issuecomment-3997770465:

```yaml
- name: my_second_dbt_model
  description: "A starter dbt model"
  config:
    materialized: incremental
  constraints:
    - type: primary_key
      columns: [id]
  columns:
      - name: id
        description: "The primary key for this table"
        tests:
            - unique
            - not_null
```

Run the model and see that a constraint was added.

In some cases this can cause database errors like the following:
```
Database Error in model c_financial_accounting_gl_detail 
  Cannot create the primary key `c_financial_accounting_gl_detail_pk` because its child column(s) `seqnbr` is nullable.
```

### Expected behavior

Constraints should only be applied for enforced contracts. The constraint should not be applied in this case because the model contract is not enforced.

### Proposed fix
Proposed fix here:
https://github.com/dbt-labs/dbt-core/issues/12400#issuecomment-3997901606

### System information
**The output of `dbt --version`:**

```markdown
- dbt-databricks (working): 1.11.3post1+cd5d71f50cd9f003e2f879dc6729a3aa93e1cdaf
- dbt-databricks (regression): 1.11.4post1+ec229f85e80de8737ba308b58059fef56cf9e3fd
```

### Additional context

Originally reported in https://github.com/dbt-labs/dbt-core/issues/12400. Root cause explained in https://github.com/dbt-labs/dbt-core/issues/12400#issuecomment-3997901606.

PR #1470, opened 2026-05-20
fix: gate column-level constraints on contract.enforced (#1381)

### LINKED ISSUE #1381: Setting Constraints (PK/FK) not working properly
### Describe the bug
We have two Problems when we want to apply Constraints (Primary Key + Foreign Keys) on our models with the dbt-databricks Adapter.

#### 1. Multi-Column Primary Key
When we are defiing  multi-column primary Constraint based on the [dbt Documentation](https://docs.getdbt.com/reference/resource-properties/constraints?version=2.0) the compiled SQL doesnt contain a Statement for the Primary Key Generation. Only Primary Keys where we are defining a single column PK are included in the SQL.

Example SQL (based on the YAML-Definition below) where we applied a Multi-Column PK and a Foreign Key:
```
/* {"app": "dbt", "dbt_version": "1.11.6", "dbt_databricks_version": "1.11.6", "databricks_sql_connector_version": "4.1.3", "profile_name": "project", "target_name": "dev-local", "node_id": "model.project.table_one"} */
create or replace table `catalog`.`schema`.`table_one` (
  `hashkey` binary
    NOT NULL
    COMMENT 'Hashkey',
  `record_source` string COMMENT 'Record source',
  `load_timestamp` timestamp NOT NULL COMMENT 'Load-Timestamp',
  FOREIGN KEY (hashkey) REFERENCES `catalog`.`schema`.`table_two` (
      hashkey
    )
)
  using delta
  CLUSTER BY AUTO
  comment 'Table One'
```
As you can see the Table only creates the FK and a follow-up ALTER-Table Statement is not executed in the Query History
#### 2.  Primary Keys and Foreign Keys are removed on Re-Run the Model
When we are re-run models, with existing Constraints (PK,FK), after a successful run all Models loosing their Constraints due to the following Query:
```
ALTER TABLE
  `catalog`.`schema`.`table_one`
DROP CONSTRAINT IF EXISTS
  table_one_fk
```
No Follow-Up Query is executed which re-applies the Constraints:

Image

### Steps To Reproduce
- Our Models are materialized as Incremental models
- In our case we want to apply a Multi-Column PK to the Columns Hashkey + Load_timestamp and in parallely Hashkey is also a Foreign Key of table_two

dbt-project.yml:

```
...
flags:
  # set default adapter options for running this project
  use_materialization_v2: true
  warn_error_options:
    silence: [ConstraintNotEnforced]
...
models:
  +auto_liquid_cluster: true
  +persist_docs:
    relation: true
    columns: true
...
```

Create a Model with the following YAML-Definition (anonymized):
- Preconsumption: table_two exists with Primary Key.
```
models:
  - name: table_one
    description: "Table one data"
    constraints:
      - type: primary_key
        columns: ["hashkey", "load_timestamp"]
        warn_unsupported: False
    columns:
      - name: hashkey
        description: "Hashkey of Table"
        constraints:
          - type: not_null
          - type: foreign_key
            to: ref('table_two')
            to_columns: [hashkey]
      - name: load_timestamp
        description: "Load-Timestamp"
        constraints:
          - type: not_null
      - name: record_source
        description: "Record Source"
```

### System information
Core:
  - installed: 1.11.6
  - latest:    1.11.7 - Update available!

Plugins:
  - databricks: 1.11.6 - Up to date!

Windows

Python 3.12.3

sd-db:
## Summary

Closes #1381. Completes the contract-enforcement work from #1342 / #1343 for the column-level subpath.

Under `use_materialization_v2: true` without `contract.enforced: true`, the macro-layer gate stripped *model-level* constraints but left *column-level* ones untouched — they flowed through `parse_columns_and_constraints` into `CREATE TABLE`

**Breaking** for anyone relying on the leak — column-level NOT NULL / PK / FK on V2 now requires `contract.enforced: true`. Called out in CHANGELOG.

## Test plan

- [x] Unit: `TestParseColumnsAndConstraintsGate` in `tests/unit/test_constraints.py` (4 cases)
- [x] Functional: `tests/functional/adapter/constraints/test_column_constraint_gate.py` — gated and ungated paths against a live cluster
- [x] Full constraint regression sweep — 43 passed, 0 failed
- [x] Pre-commit clean


jprakash-db (approved):
LGTM. Thanks for making the changes

PR #1472, opened 2026-05-20
fix: skip DESCRIBE TABLE EXTENDED AS JSON for foreign/federated tables

willweld:
Resolves #

### Description

For federated/foreign tables, `DESCRIBE TABLE EXTENDED ... AS JSON` is not supported. The adapter was attempting it on every `get_columns_in_relation` call before catching the error and falling back to plain `DESCRIBE TABLE`, causing repeated failures and extra latency on affected code paths (e.g. incremental merge strategy, table create).

The fix adds `relation.is_foreign_table` to the `use_legacy_logic` guard in `get_columns_in_relation`, so the JSON path is never attempted for foreign tables. This is consistent with the existing check in `is_describe_as_json_supported()` which already excludes foreign tables — the column-fetch path was simply missing the same guard.

No new infrastructure is needed: `DatabricksRelationType.Foreign` and `relation.is_foreign_table` already exist.

### Checklist

**Note:** I do not have access to a Databricks workspace with federated tables to run functional tests against. Per CONTRIBUTING.md, I'm flagging this for a maintainer `/integration-test` run.


sd-db (approved):
Changes look good, I am looking to add a functional test + run all functional tests and will merge. Thanks for the PR !

PR #1477, opened 2026-05-23
fix: raise DbtRuntimeError when job run terminates with non-success result state

Dev-X25874:
Resolves #

### Description

While reading through `JobRunsApi` I noticed that `_handle_terminal_state`
has a gap in its error handling. When `poll_for_completion` sees a terminal
lifecycle state and calls into this method, two cases are covered: `CANCELED`
result state gets a `DbtRuntimeError`, and any non-`TERMINATED` lifecycle state
(e.g. `SKIPPED`, `INTERNAL_ERROR`) goes through `_handle_non_terminated_failure`.
What is missing is the case where `life_cycle_state` is `TERMINATED` but
`result_state` is something like `FAILED`, `TIMEDOUT`, `UPSTREAM_FAILED`, or
`UPSTREAM_CANCELED` — the method just falls off the end and returns `None`,
so dbt treats the Python model run as successful even though the notebook
actually failed on Databricks.

This is the most common production failure path: a notebook raises an
unhandled exception, Databricks marks the run `TERMINATED / FAILED`, and the
dbt run finishes green with no indication that anything went wrong.

The fix adds an `elif` after the existing `life_cycle_state != "TERMINATED"`
block. If we are in `TERMINATED` and `result_state` is anything other than
`None`, `SUCCESS`, or `SUCCESS_WITH_FAILURES`, we raise `DbtRuntimeError`
with the result state and Databricks state message so the failure is visible.

Added `test_poll_for_completion__failed` to the existing
`tests/unit/api_client/test_job_runs_api.py` test class, following the same
pattern as `test_poll_for_completion__cancelled` directly above it. The test
sets `life_cycle_state = TERMINATED` and `result_state = FAILED`, and asserts
that a `DbtRuntimeError` containing `result_state FAILED` is raised. On the
original code the call returns silently and the test fails; after the fix it
passes.

### Checklist


sd-db (approved):
Changes look good, thx!

PR #1474, opened 2026-05-21
chore(deps): bump databricks-sdk and databricks-sql-connector ceilings

sd-db:
Bump ceilings on `databricks-sdk` and `databricks-sql-connector` to admit newer versions. Floors unchanged.

- `databricks-sdk`: `<0.78.0` → `<0.105.0`
- `databricks-sql-connector[pyarrow]`: `<4.1.6` → `<4.3.0`
- `packaging>=21.0, <26.0` declared explicitly (was transitive via `dbt-core`).
- `uv.lock` refreshed to the latest within the new ranges.

Issue #767, opened 2024-08-14
Retry failed models due to connection or server error


septponf:
If a model fails due to intermittent failures not related to the model itself, it would be nice to have an auto retry.
For example, during the summer we have had some scheduled model failures due to "Remote end closed connection without response" or "Query could not be scheduled: HTTP Response code: 503. Please try again later. SQLSTATE: XX000".

For context we are executing DBT as a Databricks job using the DBT Task and SQL Serverless for compute.

For reference, I believe the bigquery adapter has such features.

benc-db:
In order to get that message, it generally has already retried for 15 minutes.  What did you have in mind?

septponf:
> In order to get that message, it generally has already retried for 15 minutes. What did you have in mind?

Ok, Well looking at the model execution timings it does not look like any retries where attempted.
See example log below.

```

...
03:35:29  Running with dbt=1.7.17
03:35:31  Registered adapter: databricks=1.7.10
...
03:42:45  134 of 158 START sql table model xx  [RUN]
03:42:46  134 of 158 ERROR creating sql table model xx  [ERROR in 1.09s]
...
03:43:47  Finished running 99 view models, 59 table models, 1 hook in 0 hours 6 minutes and 47.45 seconds (407.45s).
03:43:47  
03:43:47  Completed with 1 error and 0 warnings:
03:43:47  
03:43:47    Runtime Error in model xx (models/path/to/xx.sql)
  Runtime Error
    ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
03:43:47 
```

benc-db:
Hmm, this is very strange, as it indicates the connection was actively broken; I don't think we have retry in that circumstance, but we file a bug against databricks-sql-connector to remedy that.  Does this happen often?  If so, I would file a ticket with Databricks to understand why you're getting disconnected.

septponf:
Happened a couple of times in July, and tonight actually.
There is no entry in the query history, so I am assuming the query does not even get provisioned in the SQL engine.

It has been the same model each time which is strange. It has very simple logic so I am thinking it could be due to execution timing.
Maybe it is due to parallelism. I see in the dbt log that 7 models are starting at the same time (same second).  We run dbt in 7 threads on a 6 DBU serverless cluster.

Could be that the request sometimes is not queued properly or that we are exceeding some API rate limit?
I will file a support ticket to investigate this further.

Nevertheless, it would be nice with an auto retry :-).
[...]

PR #1479, opened 2026-05-25
feat: SPOG (Single Point of Gateway) host support

sd-db:
## Summary

Adds SPOG (Single Point of Gateway) support — account-level vanity hosts (e.g. `xyz.azuredatabricks.net`) where workspaces are disambiguated by `?o=<workspace-id>` on `http_path`. Matches the contract in `databricks-sql-python` ([#767](https://github.com/databricks/databricks-sql-python/pull/767)), `databricks-sql-go`, `databricks-jdbc`, and the ADBC Rust driver.

Opt-in via the dep-ceiling bumps in [#1474](https://github.com/databricks/dbt-databricks/pull/1474): activates only when `databricks-sql-connector ≥ 4.2.6` and `databricks-sdk ≥ 0.76.0` are installed (`Config.workspace_id` was introduced in SDK 0.76.0; verified end-to-end against a SPOG vanity URL). Legacy hosts and older deps are unaffected.

### Pre-flight checks

| Host type | `?o=` in `http_path` | Outcome |
|---|---|---|
| SPOG (`unified`) | present | proceed — workspace id passed to SDK |
| SPOG (`unified`) | missing | **warning** |
| non-SPOG | present | **warning** |
| non-SPOG | missing | proceed |
| host probe failed | — | proceed (probe is non-fatal) |

## Test plan

- [x] Unit (`hatch run unit tests/unit -q`) — 1174 passed
- [x] `pre-commit run --all-files`
- [x] `dbt debug` SPOG status block verified on both
- [x] New SPOG tests added (unit + functional)
- [x] Integration test run succeeds (🆗 [link](https://github.com/databricks/dbt-databricks/pull/1479?timeline_per_page=5#issuecomment-4564489088))
- [x] Integration test run suceeds at `min-deps` (🆗 [link](https://github.com/databricks/dbt-databricks/pull/1479?timeline_per_page=5#issuecomment-4593452041))
- [x] New workflow to run integration tests on spog urls weekly for sanity check (`spog-integration.yml`) succeeds
- [ ] Follow up docs PR


sd-db:
All SPOG Integration Tests are passing 🆗 

~~Note:- [SPOG Integration Tests / uc-sql-endpoint (0) (push)](https://github.com/databricks/dbt-databricks/actions/runs/26598817182/job/78394191525?pr=1479) failing is a regression in main that is due to a server side change. Fix raised in #1489.~~

jprakash-db:
---
🔴 High-Confidence / High-Severity

F1 — requests is imported at module load but not a declared runtime dependency [HIGH CONF · 2 reviewers]
dbt/adapters/databricks/spog/probe.py:13 does a top-level import requests. probe.py is imported transitively by decision.py → connections.py and by impl.py, so requests loads on every adapter import — not lazily. But pyproject.toml [project].dependencies (lines 24–27) lists only databricks-sdk and databricks-sql-connector[pyarrow]; requests appears only as the dev stub types-requests (line 65). It works today by transitive luck (the SDK/connector pull requests in), but that isn't contractually guaranteed across the allowed version ranges. If a future SDK drops transitive requests, the entire adapter fails to import for everyone, SPOG or not.
Fix: add requests to runtime dependencies, or import it lazily inside probe_host. Verified against pyproject.toml @ head.

F2 — The network probe runs on every open() for all capable-deps users — no host/?o= gate [HIGH CONF · 4 reviewers]
connections.py:493 calls check_spog_preconditions unconditionally in open(). The only short-circuit (decision.py:30) is the dep-version gate (connector_supports_spog() and sdk_supports_workspace_id()). There is no early-out for "no ?o= present anywhere." So any user whose resolver picks connector ≥ 4.2.6 + sdk ≥ 0.76.0 (the pyproject ceiling <4.3.0/<0.105.0 permits this) now pays a blocking HTTPS GET to https://{host}/.well-known/databricks-config on first connect to every host — including legacy, non-SPOG, no-?o= setups that get zero benefit. In a locked-down/air-gapped egress environment this is a guaranteed retry-budget stall on first open.
Fix: skip the probe entirely when no http_path contains ?o= (i.e. gate on extracted being non-empty), so legacy users aren't probed. Verified at decision.py:30/46 + connections.py:493.

▎ ⚠️ The PR description attributes the "opt-in via dep-ceiling bumps" to sibling PR #1474 — this PR's own pyproject.toml floors are unchanged (sdk>=0.68.0, connector>=4.1.1). Even if #1474 raises the floor, the gate remains dep-version, not host-shape — so "legacy hosts unaffected" is not true for the latency/network path regardless of where the floor lives.

F3 — Contract mismatch with the connector: cluster-path SPOG without ?o= splits data-plane vs control-plane routing [HIGH CONF · 2 reviewers · verified at the branching point]
dbt's extract_workspace_id (extract.py:7–12) reads only ?o=. The connector's _extract_spog_headers (databricks-sql-python session.py:212–226) inspects two sources in priority order: ?o= then the cluster-path segment /sql/protocolv1/o/<id>/<cluster>. Consequence: on a SPOG host using an all-purpose cluster http_path without ?o=, the connector routes SQL (data-plane) correctly via the path segment, but dbt never extracts workspace_id, so the SDK Config(workspace_id=...) (control-plane: REST/Jobs/Workspace) is not plumbed — a silent split-brain, the exact thing the PR says it fixes. Worse: check_spog_preconditions will emit the "no ?o=" warning for a config the connector already handles, and the dbt test docstrings asserting "the connector only reads ?o=" are factually wrong — they'll mislead future maintainers.
Fix: either (a) make extract_workspace_id fall back to the /o/<id>/ path segment to match the connector, or (b) if cluster-path SPOG is out of scope, say so explicitly in the warning + CHANGELOG and correct the test docstrings. Verified at session.py:184–226.

---
🟡 Medium-Severity

F4 — ?o= value is never validated; asymmetric with connector's ^[0-9]+$ [MODERATE CONF · 2 reviewers]
The connector only injects the org-id header if the value matches _ORG_ID_RE = ^[0-9]+$ (session.py:218), silently dropping a non-numeric value. dbt's extract_workspace_id returns the raw string (?o=abc → "abc"), which then flows into Config(workspace_id="abc"). Result: dbt plumbs a bad value into the control plane while the connector drops it on the data plane — another silent asymmetry with no diagnostic. Fix: mirror the connector's numeric guard and warn on mismatch. Verified.

F5 — Probe-failure logged at INFO while config smells log at WARNING [MODERATE CONF · 2 reviewers]
probe.py:50 logs the probe-exhaustion message (the case most likely to precede real routing failures) at logger.info, while the benign "looks like SPOG, no ?o=" smell logs at logger.warning (decision.py:49/60). The signal an operator most needs is the quietest. Fix: raise the probe-failure message to WARNING. Verified.

F6 — Existing connection-manager test now makes a live network call [MODERATE CONF · 2 reviewers]
tests/unit/test_connection_manager.py::test_open_calls_is_cluster_http_path_for_warehouse was modified to add compute = None and host = "example.cloud.databricks.com" but does not patch check_spog_preconditions / probe_host. With capable deps installed, open() reaches the real probe → live requests.get to example.cloud.databricks.com with 3 retries + real backoff. The _guard_unmocked_requests_get fixture only covers tests/unit/spog/, not this file. Fix: patch check_spog_preconditions (or stub probe_host) here, like TestOpenSpogIntegration does. Verified in diff.

F8 — Worst-case ~16s blocking connect latency on a hanging probe endpoint [MODERATE CONF · 3 reviewers]
requests.get(url, timeout=5) × 3 attempts + ~1.75s backoff ≈ ~16s of serial blocking added to first open() if /.well-known/databricks-config accepts the TCP connection but hangs. Mitigated by @cache (one probe per host per process) and non-fatal fallback, but for short-lived dbt run invocations or per-shard CI workers it's paid every run. Fix: tighten to a connect/read tuple e.g. timeout=(2, 3). (Performance scoped this LOW given caching; ops scoped it HIGH — reported at the conservative middle.) Verified.

jprakash-db (approved):

sd-db:
- F1 - Unrelated to the current task and mostly fine
- F2 - not required
- F3 - Not true, this was fixed
- F4 - Not required, we don't do the validations at our end. Better to bubble out whatever error server returns here
- F5 - Maybe we can move everything to INFO, let me check again. EDIT - Current code is fine, not required
- F6 - This is expected, not an issue
- F8 - Not an issue

PR #1488, opened 2026-05-29
test: stabilize TestChangingSchema* python-model tests under min-deps (dbt-core 1.11.2)

sd-db:
## Problem

The nightly **Integration Tests (Min-Deps)** workflow fails on the SQL-warehouse shard ([run 26536340109](https://github.com/databricks/dbt-databricks/actions/runs/26536340109)):

```
FAILED tests/functional/adapter/python_model/test_python_model.py::TestChangingSchemaV2::test_changing_unique_tmp_table_suffix
FAILED tests/functional/adapter/python_model/test_python_model.py::TestChangingSchemaIncrementalV2::test_changing_unique_tmp_table_suffix
  dbt.exceptions.ParsingError: Failed to render models/schema.yml ...
    dbt.exceptions.EnvVarMissingError: Env var required but not provided: 'DBT_TEST_SCHEMA_NAME_VARIABLE'
```

The same `main` commit **passes** these tests on latest-deps (dbt-core 1.11.8) the same night — this is specific to **dbt-core 1.11.2**, newly exercised since min-deps became first-class CI (#1480, #1476).

## Root cause

Under `--dist=loadfile`, every `test_python_model.py` class runs in one xdist worker. On dbt-core 1.11.2, a sibling `BasePythonModelTests` class's `test_source` schema.yml — whose `schema` renders `{{ var(env_var('DBT_TEST_SCHEMA_NAME_VARIABLE')) }}` — leaks into the parse of a *following* class. The `TestChangingSchema*` classes don't inherit `BasePythonModelTests`, so they supply neither thing that leaked source needs:

- `env_var('DBT_TEST_SCHEMA_NAME_VARIABLE')` resolves from the per-class invocation-context snapshot → needs the env var set.
- `var('test_run_schema')` is resolved by `ConfiguredVar` during schema rendering, which reads **only `config.cli_vars`** (`dbt/context/configured.py`) → needs `--vars`, not project-level `vars:`.

It is position-dependent: `TestChangingSchema` (V1) passes only because its predecessor has no source; the V2 classes run right after `TestServerlessCluster*`. dbt-core 1.11.8 isolates parses correctly, so latest-deps is unaffected.

## Fix

Test-only. A `SchemaNameVarMixin` that (a) sets `DBT_TEST_SCHEMA_NAME_VARIABLE` via an autouse fixture and (b) passes `test_run_schema` via `--vars` on `run_dbt` — exactly mirroring dbt-core's own `BasePythonModelTests`. Applied to all four `TestChangingSchema*` classes (the two failing V2 classes plus the two V1 classes, which pass only by ordering luck today). Inert on dbt-core 1.11.8.

## Verification

Against the live SQL warehouse under min-deps (dbt-core 1.11.2), full file with `-n 2 --dist=loadfile --reruns 1` (faithful to CI):

- Before: **2 failed** (`EnvVarMissingError`) — reproduces CI.
- After: **11 passed, 9 skipped, 0 failed** (both V2 classes pass).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - Defer SDK `Config` construction to connection-open time so offline paths (`dbt parse`/`list`/`compile`) don't trigger the host-metadata probe introduced in `databricks-sdk>=0.103`; as a side effect, auth errors now surface at first connection rather than during profile parsing. ([#1474](https://github.com/databricks/dbt-databricks/pull/1474))
 - Bump ceilings on `databricks-sdk` (now `<0.105.0`) and `databricks-sql-connector[pyarrow]` (now `<4.3.0`) to admit newer releases; floors unchanged. ([#1474](https://github.com/databricks/dbt-databricks/pull/1474))
+- Stabilize the `TestChangingSchema*` Python-model functional tests under min-deps (dbt-core 1.11.2), where a sibling class's source schema.yml could leak into their parse and fail with `EnvVarMissingError`. ([#1488](https://github.com/databricks/dbt-databricks/pull/1488))
 
 ## dbt-databricks 1.12.0 (May 18, 2026)
 
--- a/tests/functional/adapter/python_model/test_python_model.py
+++ b/tests/functional/adapter/python_model/test_python_model.py
@@ -1,3 +1,4 @@
+import json
 import os
 
 import pytest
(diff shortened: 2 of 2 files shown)


jprakash-db (approved):

sd-db:
Integration test failures are due to a separate issue which are already fixed in #1489

PR #1489, opened 2026-05-29
fix: ignore server-set delta.parquet.compression.codec tblproperty

sd-db:
### Summary

A recent server-side change began stamping the non-user-configurable property `delta.parquet.compression.codec=zstd` onto streaming tables and materialized views. Because it was absent from `TblPropertiesConfig.ignore_list`, it leaked into the existing-relation config but never the desired (model) config — so `get_changeset` reported a **phantom configuration change on every run of an otherwise-unchanged relation**.

Impact by `on_configuration_change`:
- `fail` → idempotent `dbt run`s now **error out** (regression)
- `apply` → spurious `ALTER … SET TBLPROPERTIES` every run
- `continue` → spurious "configuration changes were identified" warning every run

This surfaced fleet-wide on 2026-05-28 (both `main` nightly and feature-branch CI); the same tests passed on 2026-05-27 with no intervening code change, pinning the cause to the backend.

### Fix

Add `delta.parquet.compression.codec` to `TblPropertiesConfig.ignore_list` so it is dropped during config comparison, matching the existing pattern for other Databricks-set properties (e.g. `delta.enableDeletionVectors`, `delta.columnMapping.maxColumnId`).

### Testing

- **Unit:** added `test_from_results__drops_ignored_properties` to lock the filter behavior; full unit suite passes (1116 passed).
- **Functional:** `TestStreamingTableChangesApply::test_change_is_applied_via_alter` passes against a live SQL warehouse (was failing in CI). The other 6 failures — including `test_idempotent_run_does_not_fail`, which exercises the regression directly — route through the same comparison and are covered by the same filter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixes
 
+- Ignore the server-set `delta.parquet.compression.codec` tblproperty when diffing relation configs, so streaming tables and materialized views are not flagged as changed after the backend started stamping it automatically ([#1489](https://github.com/databricks/dbt-databricks/pull/1489))
 - Fix missing f-string prefix in `JobRunsApi.submit` debug log ([#1471](https://github.com/databricks/dbt-databricks/pull/1471))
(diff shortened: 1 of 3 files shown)


jprakash-db (approved):

sd-db:
Integration run looks okay, the issue at hand is fixed. There is a single failing test `TestPersistDocsColumnMissing::test_missing_column` on all three profiles which is due to a breaking change in a test dependency `dbt-tests-adapters`. Fix raised in #1490.

PR #1546, opened 2026-06-19
fix: stop metric views re-altering on every run under view_update_via_alter

sd-db:
### Description

Metric views configured with `view_update_via_alter=true` re-issued a redundant `ALTER VIEW ... AS` on every `dbt run`, even when the definition was unchanged.

`MetricViewQueryConfig.get_diff` compared the model's compiled YAML against the server's stored `View Text` with a whitespace-only text comparison. Databricks re-renders the stored definition on read-back — it requotes `source` from double to single quotes and rewrites flow-style lists (e.g. `synonyms: [a, b]`) to block style — so the normalized strings never matched and dbt always saw a change.

The fix compares the **parsed YAML structurally** (`yaml.safe_load`), so quote style and flow/block list style no longer register as changes. It falls back to the previous whitespace-normalized text comparison when a definition isn't valid YAML. Genuine definition changes are still detected, so real updates still apply via `ALTER`/replace.

Scope is specific to metric views: regular views / materialized views / streaming tables diff via `QueryConfig` over `information_schema.views.view_definition`, which for plain SQL is returned as submitted — verified that a regular view with `view_update_via_alter` does not perpetually diff.

### Testing

- Unit (`tests/unit/relation_configs/test_metric_view_config.py`): new `get_diff` cases for double-vs-single-quoted `source` and flow-vs-block lists (both failed before the fix), plus semantic-difference-detected and unparseable-fallback cases.
- Functional (`tests/functional/adapter/metric_views/test_metric_view_noop.py`): a metric view with `view_update_via_alter` is unchanged across an identical re-run (`information_schema.tables.last_altered` does not advance).
- Full `tests/functional/adapter/metric_views/` suite (19) and the unit suite pass — no regressions.

### Checklist


jprakash-db (approved):

sd-db:
Single integration test failure is for a known flaky test which can be ignored

PR #1561, opened 2026-06-26
fix: deterministically name unnamed PK/FK so incremental constraints don't churn or collide

### LINKED ISSUE #1333: TABLE_OR_VIEW_NOT_FOUND on constraint DROP during V2 incremental materialization

### LINKED ISSUE #1344: 1.11.5 fails with DELTA_CONSTRAINT_ALREADY_EXISTS in incremental models with two or more foreign keys to same relation
### Describe the bug
Running dbt build or run with dbt-databricks 1.11.6 fails if two or more foreign_key constraints for the same model relation are defined in the YAML of an incremental model. Running an incremental run or the first run raises:

```
> Database Error in model incremental_model (models/incremental_model.sql)
>   [DELTA_CONSTRAINT_ALREADY_EXISTS] Constraint 'incremental_model_relation_model_fk' already
 exists. Please delete the old constraint first.
  Old constraint:
  incremental_model_relation_model_fk FOREIGN KEY (`column_1`) REFERENCES `catalog`.`schema`.`relation_model` (`column_pk`)
```

After adding names to the foreign constraints, the error is not raised anymore, but this condition was not necessary in previous versions of dbt-databricks (previously using 1.10.9).
This only happens for incremental models. 

### Steps To Reproduce
models.yml
```
models:
    - name: incremental_model
        config:
          contract:
            enforced: true
          meta:
            owner: '@owner'
        columns:
          - name: column_1
                  data_type: string
                  description: Foreign Key.
                  data_tests:
                    - relationships:
                        arguments:
                          to: ref('relation_model')
                          field: column_pk
                  constraints:
                    - type: foreign_key
                      to: ref('relation_model')
                      to_columns: [column_pk]
          
          - name: column_2
                  data_type: string
                  description: Foreign Key.
                  data_tests:
                    - relationships:
                        arguments:
                          to: ref('relation_model')
                          field: column_pk
                  constraints:
                    - type: foreign_key
                      to: ref('relation_model')
                      to_columns: [column_pk]
    - name: relation_model
      config:
        contract:
          enforced: true
        meta:
          owner: '@owner'
      columns:
        - name: column_pk
          data_type: string
          data_tests:
            - not_null
            - unique
          constraints:
            - type: primary_key
            - type: not_null
```

incremental_model.sql:
```
{{
    config(
        materialized="incremental",
        on_schema_change="fail",
        unique_key="accounting_category_id",
    )
}}
with
import_cte as (
    select
        -- Primary key
        incremental_pk,

        -- Foreign keys
        column_1,
        column_2,

        -- Timestamp fields
        updated_at

        from {{ ref('stg_model') }}
    {% if is_incremental() %}
    where updated_at >= (select max(updated_at) from {{ this }})
    {% endif %}
)

select * from import_cte
```

The target/run incremental_model.sql file that causes the error:

```
ALTER TABLE `catalog`.`schema`.`incremental_model` ADD FOREIGN KEY (column_2) REFERENCES `catalog`.`schema`.`relation_model` (column_pk);
```
### Expected behavior
Both foreign keys are correctly added to the incremental model even when no name is specified, as it happens for non incremental models.

Terminal output:
```
> Database Error in model incremental_model (models/incremental_model.sql)
>   [DELTA_CONSTRAINT_ALREADY_EXISTS] Constraint 'incremental_model_relation_model_fk' already
 exists. Please delete the old constraint first.
  Old constraint:
  incremental_model_relation_model_fk FOREIGN KEY (`column_1`) REFERENCES `catalog`.`schema`.`relation_model` (`column_pk`)
```

### System information
```
Core:
  - installed: 1.11.6
  - latest:    1.11.7 - Update available!

Plugins:
  - databricks: 1.11.5 - Up to date!
```

MacOS Tahoe 26.3

Python 3.10.11

### Additional context
This problem first happens after upgrading from 
dbt-databricks 1.10.9 . Running tests in local before upgrading in production.

sd-db:
## What

Unnamed primary/foreign keys churned on every incremental run, and two or more unnamed foreign keys to the same parent failed with `DELTA_CONSTRAINT_ALREADY_EXISTS`. Databricks auto-names an unnamed key, so the model's `name=None` never matched the catalog read; the diff issued `DROP CONSTRAINT ... CASCADE` + re-add each run (silently dropping dependent FKs), and the unnamed FK re-adds collided on the server-derived `<table>_<parent>_fk` name.

## Fix

dbt now assigns an unnamed PK/FK the same deterministic name its create-time macro already writes (`synthesize_constraint_name`), generated identically on the create and incremental paths, so the model and catalog agree and the diff is a no-op. A foreign key's name folds in its referenced columns (its full identity), so distinct FKs never collide — and a config that currently fails at create (two unnamed FKs from one column to different parent columns) becomes possible.

- No change to the diff key or `get_diff` — once names are deterministic, the existing `(name, columns)` comparison + repoint reconciliation just work.
- Macro↔Python name parity is pinned by `TestConstraintNameParity` so the two implementations can't drift while both exist.

## Behavior

Existing unnamed foreign keys are renamed once on the next incremental run (a no-op `DROP`/re-add, no cascade), then stable — consistent with #1551 folding `expression` into the PK name in this same release. An unnamed key's `expression`/target edit is now reconciled (its name encodes it), unlike a named key (#1552 — still needs `--full-refresh`); the #1552 changelog note is scoped to named keys accordingly.

## Tests

- **Unit:** name generator reproduces a live-captured macro hash; `from_relation_config` names unnamed PK/FK; `get_diff` no-ops for matched and two-FK-to-one-parent cases (with a negative assertion); macro↔Python parity for FK (single/multi/expression) and PK (incl. RELY); distinct `to_columns` → distinct names.
- **Functional** (UC SQL warehouse, `use_materialization_v2: false`): two unnamed FKs to one parent survive an incremental run (#1344); an unnamed `RELY` PK's dependent FK survives an incremental re-run (#1333 + #1513). Both verified RED→GREEN on a live warehouse.

Supersedes #1554 (its unnamed-PK fix is reimplemented here more generally; its RELY-PK regression test is carried forward as an unnamed-PK variant).

Closes #1333
Closes #1344


jprakash-db (approved):

lorogno:
@sd-db Planned for Release 1.12.3?

sd-db:
@lorogno this is more involved change and would require a minor version bump. This PR is also incomplete as it misses the v2 path currently

PR #1567, opened 2026-07-01
test: pin refresh-schedule interval equality for MV and ST

sd-db:
## Description

Databricks reads back a `15 MINUTES` refresh interval as `900 SECOND`. The existing schedule-lifecycle tests only asserted `"900" in refresh.at_most_every`, which doesn't guard against the unit normalizing the two forms differently and producing a spurious schedule diff on an unchanged re-run.

Adds an equality assertion — `refresh == RefreshConfig(on_update=True, at_most_every="15 MINUTES")` — to both `TestMaterializedViewScheduleLifecycle` and `TestStreamingTableScheduleLifecycle`, pinning that the parsed config round-trips to the configured interval.

Server-observable (parsed `RefreshConfig` from `describe extended`). Verified green on `databricks_uc_sql_endpoint` (8 passed).

Test-only; no changelog entry required.

### DIFF
--- a/tests/functional/adapter/materialized_view_tests/test_schedule_modes.py
+++ b/tests/functional/adapter/materialized_view_tests/test_schedule_modes.py
@@ -121,6 +121,9 @@ class TestMaterializedViewScheduleLifecycle:
         assert refresh.mode.value == "on_update"
         assert refresh.at_most_every is not None
         assert "900" in refresh.at_most_every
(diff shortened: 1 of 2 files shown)


sd-db:
Closing — the existing MV/ST schedule-lifecycle tests already provide adequate coverage of the refresh-schedule interval round-trip; the added equality assertion is redundant.

PR #1572, opened 2026-07-01
perf: apply only new or changed databricks_tags on table re-runs

### LINKED ISSUE #1308: Databricks_tags are set every run
### Describe the feature
The current situation is that all databricks-tags are applied every dbt run. This adds significantly to the time of a run. In our situation that means setting one tag takes around 300 milliseconds. We want to tag each column in our dbt project: setting the tags for this much columns doesn't scale very well. The production workload takes more time and it doesn't work well for developers who want to test their code. 

We would like to know if it's possible that only new or changed databricks-tags will be applied, instead of all. As can be seen in the adapter, [every time a table is created](https://github.com/databricks/dbt-databricks/blob/main/dbt/include/databricks/macros/materializations/table.sql#L63), all tags are set. In the apply_tags macro there is no option in which tags that have already been set are skipped (even though the tags retain). 

We do see that the tags are [being fetched already](https://github.com/databricks/dbt-databricks/blob/bfcb5c7c7714e97e67023119f674d2938b04acb0/dbt/include/databricks/macros/relations/components/column_tags.sql#L11), but we do not see this used yet. We are wondering if this functionality, so that not all tags are being reset with every run, is on the roadmap of the maintainers of this adapter. 

#### The output of `dbt --version`:
```
Core:
  - installed: 1.10.13
  - latest:    1.11.2  - Update available!

Plugins:
  - databricks: 1.11.0 - Update available!
  - spark:      1.9.3  - Update available!

  At least one plugin is out of date with dbt-core.
```

#### The operating system you're using:
MacOS

#### The output of `python --version`:
Python 3.12.12

#### The dbt-databricks version we're using: 
dbt-databricks==1.10.4

sd-db:
## Problem

Table rebuilds and incremental full-refresh replacements reapply configured table and column tags even when those tags already match the server. This adds unnecessary `ALTER … SET TAGS` statements, particularly for models with many tagged columns.

## Root cause

Table rebuilds and incremental full-refresh replacements bypass the normal incremental ALTER changeset flow. Their creation paths therefore did not reuse the existing `TagsConfig.get_diff` and `ColumnTagsConfig.get_diff` comparisons.

## Fix

- Share `reconcile_tags` across the V1 and V2 Table and Incremental full-refresh paths.
- After an in-place table replacement, compare configured tags with the tags retained on the server and skip application when they match.
- Expose one tag-only adapter helper, `get_table_replacement_tag_changes`, so Jinja can reuse the Python processors and component diffs without fetching unrelated relation configuration.
- Fetch only configured tag categories; models with no tags skip tag metadata reads entirely.
- Apply full configured tags for fresh creates, drop-and-recreates, shallow-clone rebuilds, and safe staging replacements. An existing relation alone is not sufficient to enable the diff.
- Leave normal incremental updates on their existing changeset path, including `incremental_apply_config_changes`.

This PR wires replacement paths into the existing component diffs; it does not introduce per-key filtering. At this head, a changed table tag can reapply the desired table-tag map, and a changed column can reapply that column's desired tags. Finer-grained table and column tag diffs are handled separately in #1667 and #1668.

Only tags are in scope; reconciliation of other metadata after replacement is unchanged.

## Validation

- Full unit suite: 1,356 passed, 4 skipped.
- Code-quality and all-files pre-commit checks passed.
- Adapter and macro unit tests cover conditional metadata reads and full-versus-diff tag application.
- Eight live Unity Catalog SQL warehouse cases passed: Table and Incremental full-refresh, V1 and V2, with table and column tags. Each case queries `information_schema` after creation, unchanged full-refresh, and changed full-refresh to verify the stored tag values.
- Additional tag lifecycle, shallow-clone, and normal incremental column-tag tests passed during development.

Live replacement validation used Delta tables; managed Iceberg behavior was not verified.

Resolves #1308


jprakash-db (approved):

sd-db:
Closing this for now. The extra `information_schema` reads could cost more than reapplying the tags, and we haven’t measured a clear win. I’d rather keep Table rebuilds and incremental full-refresh simple.

#1667 and #1668 are separate—they improve tag diffs where we already use changesets.

---

_This comment was generated with [GitHub MCP](http://go/mcps)._

PR #1576, opened 2026-07-06
feat: add kernel integration test lane + use_kernel auth support

sd-db:
## Description

Adds a weekly integration-test lane that exercises the dbt-databricks functional suite through the **databricks-sql-connector Rust kernel backend** (`use_kernel=True`, connector 4.3.0), plus the adapter change required for connections to authenticate on that path.

### Why an adapter change is needed

The kernel routes over the SEA (Statement Execution API) HTTP transport and owns the OAuth token lifecycle itself, so its auth bridge **rejects the connector `credentials_provider`** that dbt passes on the default Thrift path. Without a change, every connection under `use_kernel=True` fails at open. `SqlUtils.prepare_connection_arguments` now, when `use_kernel` is set, forwards the raw credential the bridge understands instead:

| dbt auth | Kernel handling |
| --- | --- |
| PAT (`token`) | forward `access_token` |
| Databricks OAuth M2M (`client_id`/`client_secret`) | forward `oauth_client_id`/`oauth_client_secret` + the credential manager's resolved `oauth_scopes` |
| OAuth U2M (`auth_type: oauth`) | translate to kernel `auth_type=databricks-oauth`, forwarding the credential manager's resolved `oauth_client_id` + `oauth_scopes` (dbt's defaults, not the kernel's) |
| Native Azure SP (`azure_client_id`/`azure_client_secret`) | raise a clear `DbtConfigError` — the kernel has no Azure-AD flow, so fail loud rather than misroute |

The default (Thrift) path is unchanged.

### Workflow

`.github/workflows/integration-kernel.yml` mirrors the SPOG scheduled smoke but runs a single job: it installs the optional `databricks-sql-connector[kernel]` extra and runs the `databricks_uc_sql_endpoint` profile with `DBT_DATABRICKS_USE_KERNEL=1`. Scheduled **Saturday 09:17 UTC**, clear of the daily nightly and Sunday SPOG slots that share the peco warehouse.

SEA is **warehouse-only** (the connector rejects an all-purpose-cluster http_path), so only the SQL-warehouse profile runs, and the handful of tests that pin a model to a cluster path/compute are marked `@pytest.mark.skip_kernel` (a new marker enforced by an autouse fixture, gated on `DBT_DATABRICKS_USE_KERNEL`).

## Verification

- New unit tests in `tests/unit/test_handle.py` pin each auth shape's forwarded kwargs (PAT, M2M with/without `dose` prefix, `oauth_scopes`, U2M with/without explicit `client_id`, native-Azure guard).
- Auth routing independently validated against the real kernel wheel via dummy connections: PAT / M2M / U2M reach the network (bridge accepts); native Azure SP guards before connect.
- Live end-to-end on a SQL warehouse: `use_kernel=True` opens a `KernelDatabricksClient` and a basic table materialization passes (6/6); serverless Python models pass (their SQL connection stays on the warehouse).
- Full unit suite green (1250 passed); `skip_kernel`-marked tests skip under the flag and run normally without it.


dbt/adapters/databricks/handle.py
```diff
@@ -407,16 +407,64 @@ def prepare_connection_arguments(
...
+        """
+        if creds.token:
+            args["access_token"] = creds.token
+        elif creds.azure_client_id and creds.azure_client_secret:
+            raise DbtConfigError(
+                "use_kernel=True does not support Azure service-principal auth "
+                "(azure_client_id/azure_client_secret). Use a personal access token "
+                "or Databricks OAuth (client_id/client_secret) instead, or drop "
+                "use_kernel to use the default backend."
+            )
```
vikrantpuppala: can we let this error be raised at the kernel level?

vikrantpuppala: umm maybe that's not possible

vikrantpuppala (approved):

jprakash-db (approved):

PR #1587, opened 2026-07-13
test: add weekly test timings refresh workflow

saishreeeee:
Resolves #

### Description

Automates the weekly refresh of .github/test_timings.json so integration shard balancing doesn't drift stale. Adds a scheduled Refresh Test Timings workflow that regenerates timings from the last few distinct-SHA green integration runs and opens a PR when timings change substantially.

### Checklist

- [x] I have run this code in development and it appears to resolve the stated issue -> Ran it locally end-to-end against the live repo
- [ ] I have updated the `CHANGELOG.md` and added information about my change to the "dbt-databricks next" section - NA
- [ ] [Optional] I have run `/dbt-databricks-pr-ready` (AI agent skill in `.claude/skills/`) and addressed its merge-readiness feedback


.github/ISSUE_TEMPLATE/release.md
```diff
@@ -8,3 +8,13 @@ assignees: ''
 ---
 
 ### TBD
+
```
sd-db: we can remove this.

.github/workflows/integration.yml
```diff
@@ -325,7 +325,7 @@ jobs:
         with:
           name: uc-cluster-test-logs-${{ matrix.target.pr }}-shard-${{ matrix.shard }}
           path: logs/
-          retention-days: 5
+          retention-days: 14
```
sd-db: I think we can start with 7 days for now and maybe later look to extend if required

scripts/shard_assign.py
```diff
@@ -40,17 +40,24 @@ def lpt_test_count_assign(file_to_tests: dict[str, list[str]], num_shards: int)
     return _greedy_lpt(file_to_tests, num_shards, lambda fp: len(file_to_tests[fp]))
 
 
+def make_historical_weight(timings: dict[str, float]) -> Callable[[str], float]:
```
sd-db: Let us not change the shard assign logic. We can add this enhancement later in a separate PR if required

scripts/regenerate_timings.py
```diff
@@ -181,6 +181,74 @@ def median_across_runs(
     return merged
 
 
+def total_wall_by_profile(doc: dict[str, dict[str, float]]) -> dict[str, float]:
+    """Total per-profile wall seconds (sum of every file's time).
+
+    This is the drift signal the weekly refresh workflow gates on. Crucially it
```
sd-db: nit: make the comment terse

.github/workflows/refresh-test-timings.yml
```diff
@@ -0,0 +1,183 @@
...
+# stale — new test files land absent from it and get mean-filled, the shards
+# drift out of balance, and the matrix wall-clock (set by its slowest shard)
+# grows. This job refreshes it automatically and opens a human-reviewed PR when
+# balance has drifted meaningfully.
+#
+# What it does:
+#  1. Runs scripts/regenerate_timings.py to rebuild timings from the last few
+#     distinct-SHA green integration.yml runs (median per-file wall time), and
+#     in the same tested script computes per-profile total-wall drift vs the
+#     current file. Total wall time is independent of shard count, so the drift
```
sd-db: nit: 

```Total wall time is independent of shard count, so the drift
#     decision never duplicates integration.yml's (dynamically computed) shard
#     counts — a manual regen gets the identical numbers.
```

No need add detailed info here in the comments

scripts/refresh_timings.py
```diff
@@ -26,6 +26,12 @@
     ...
   }
 
+This script aggregates **one profile from one run**. To refresh the whole file
```
sd-db: I think we can remove `refresh_timings` file all-together now as we longer require to build the test timings from a single run. I think the only important things in this file are some utility functions like `classname_to_file`, `aggregate_per_file` which you can move

scripts/regenerate_timings.py
sd-db: We have repeated the same info multiple times on the timings being independent of shard count. We should follow DRY and keep this in a single place

scripts/regenerate_timings.py
```diff
@@ -181,6 +181,74 @@ def median_across_runs(
...
+    return {profile: round(sum(by_file.values()), 3) for profile, by_file in doc.items()}
+
+
+def drift_by_profile(
+    old_doc: dict[str, dict[str, float]],
+    new_doc: dict[str, dict[str, float]],
+) -> dict[str, float]:
+    """Per-profile percent change in total wall time from old to new timings.
+
+    Measured as an absolute percentage (direction-agnostic — a human reviews
```
sd-db: nit: no need for the added explanation here. Also adding a new profile is a very very rare case

.github/workflows/refresh-test-timings.yml
```diff
@@ -0,0 +1,183 @@
...
+            Automated weekly refresh of `.github/test_timings.json` from the last
+            ${{ steps.regen.outputs.runs_used }} distinct-SHA green integration runs (median per-file wall time).
+
+            **Per-profile total-wall drift** (old vs new; independent of shard count):
+
+            ${{ steps.regen.outputs.summary }}
+
+            Threshold to open this PR: `${{ env.DRIFT_THRESHOLD_PCT }}`%.
+
+            Please sanity-check the numbers before merging — this PR is **not**
```
sd-db: Cleanup JIRA link, +overall look to make the commit message terse

.github/workflows/refresh-test-timings.yml
```diff
@@ -0,0 +1,183 @@
...
+  schedule:
+    - cron: "0 6 * * 1"   # 06:00 UTC every Monday
+  workflow_dispatch:
+    inputs:
+      num_runs:
+        description: "How many distinct-SHA green integration runs to aggregate"
+        required: false
+        default: "3"
+        type: string
+      force:
```
sd-db: I think we can remove this flag, if we want to force a PR we can do this manually

.github/workflows/refresh-test-timings.yml
```diff
@@ -0,0 +1,183 @@
...
+#     current file. Total wall time is independent of shard count, so the drift
+#     decision never duplicates integration.yml's (dynamically computed) shard
+#     counts — a manual regen gets the identical numbers.
+#  2. Gates on drift: opens a PR only when the worst per-profile total-wall
+#     change exceeds DRIFT_THRESHOLD_PCT (avoids noisy weekly PRs from run-to-run
+#     variance). If drift is implausibly large it flags for manual review
+#     instead of opening a PR with potentially bad data.
+#  3. PRs are never auto-merged — bad timings degrade balancing silently, so a
+#     human always reviews the before/after numbers in the PR body.
+name: Refresh Test Timings
```
sd-db, line 3: We need to ensure that this can only run on `main`. We should not open PRs/refresh timings on branch code

scripts/regenerate_timings.py
sd-db, line 11: Let's look to add tests

sd-db (commented):
Have added some comments. Additionally please look to run the workflow locally to confirm it is working as expected. One issue I feel in the overall design is that the 10% drift check should probably be at a test file level and not a profile level. Ideally I feel 1.) for new tests files add the new value 2.) for existing test files only update if delta > 10%

Agree on the fact that if delta is ever > 60% we should raise for manual review

scripts/regenerate_timings.py
sd-db, line 141: Can we filter out just the runs on `main`. So integration tests can also run vs a branch and that can skew data.

scripts/regenerate_timings.py
```diff
@@ -54,6 +62,38 @@
...
+    if last_test_idx == -1:
+        return "/".join(parts) + ".py"
+    return "/".join(parts[: last_test_idx + 1]) + ".py"
+
+
+def aggregate_per_file(junit_paths: list[Path]) -> dict[str, float]:
+    """Sum `time` attributes across all testcases per file."""
+    by_file: dict[str, float] = defaultdict(float)
+    for jp in junit_paths:
+        for tc in ET.parse(jp).getroot().findall(".//testcase"):
```
sd-db, line 96: Could we validate parsed testcase timings before aggregating them? float() accepts NaN, infinity, and negative values, which can currently reach the generated timings file. A minimal check using `math.isfinite(value)` and `value >= 0`, together with `allow_nan=False` when serializing the JSON, should prevent invalid shard weights. Could we also add a small test covering this?

scripts/regenerate_timings.py
```diff
@@ -89,19 +149,22 @@ def discover_green_run_ids(repo: str, num_runs: int) -> list[str]:
...
+        if not run_has_live_artifacts(repo, run_id):
+            continue
         seen_sha.add(sha)
-        picked.append(str(r["databaseId"]))
+        picked.append(run_id)
         if len(picked) == num_runs:
             break
     if not picked:
-        sys.exit("ERROR: no green integration runs found")
+        sys.exit("ERROR: no green integration runs with downloadable test-log artifacts found")
```
sd-db: Could we avoid silently proceeding with fewer runs than num_runs? The local validation requested three runs but found only one, so the resulting value was not actually a three-run median. It should be fine to skip a run on a given week as this will again run the next week.

.github/workflows/refresh-test-timings.yml
```diff
@@ -0,0 +1,116 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: refresh-test-timings
+  cancel-in-progress: true
+
+jobs:
```
sd-db, line 24: Since regeneration and dependency setup do not need write access, could we avoid exposing the write token job-wide? Please set `persist-credentials: false` on checkout and provide `GH_TOKEN` only to the PR-creation step. This keeps the write capability limited to the one operation that needs it.

saishreeeee: Split into two jobs: read-only regenerate + write-only open-pr, both with persist-credentials: false.

scripts/regenerate_timings.py
```diff
@@ -62,11 +111,32 @@ def _profile_for_artifact_dir(name: str) -> str | None:
...
+        if not a.get("expired", True) and "-test-logs-" in a.get("name", ""):
+            return True
+    return False
+
+
+def discover_green_run_ids(repo: str, num_runs: int, branch: str = "main") -> list[str]:
+    """Return the newest `num_runs` green integration runs on `branch`, one per
+    distinct SHA, keeping only runs that still have downloadable test-log artifacts.
+
+    Restricted to `branch` (main) so branch/PR runs — which may run a different
```
sd-db: I don’t think --branch main fully guarantees that the artifacts tested main. Integration workflow_dispatch runs themselves execute from main, but their inputs can still target PRs or an arbitrary git_ref, so they also appear with head_branch=main. Could we additionally restrict discovery to scheduled runs, for example with --event schedule, or otherwise validate that the integration target was actually main?

sd-db (approved):
Changes look good, have added a minor comment

PR #1590, opened 2026-07-15
feat: add catalog_database support for Unity catalogs (catalogs.yml v2)

aahel:
Resolves #

### Description

Adds `catalog_database` support for Unity catalogs in catalogs.yml v2, so a model can be routed to an explicit **physical Unity catalog** independent of the dbt catalog label (`name`). This mirrors the Fusion implementation (dbt-labs/fs#11436): `databricks__generate_database_name` now returns `catalog_relation.catalog_database` first, taking precedence over the model `database` config and `catalog_name`.

Precedence (matches Fusion): `catalog_database` > model `database` config > `catalog_name` > `target.database`.

**Config** (v2 catalogs.yml):
```yaml
catalogs:
  - name: my_catalog          # dbt label (models reference this via catalog_name)
    type: unity
    table_format: iceberg
    config:
      databricks:
        catalog_database: prod_catalog   # physical Unity catalog
```

**Changes**
- `DatabricksCatalogRelation`: add `catalog_database` field.
- `UnityCatalogIntegration`: read the first-class `catalog_database` field (surfaced on the config by the base `CatalogIntegration` from the v2 catalogs.yml config block) and carry it onto the relation. Hive-metastore is unaffected (no catalog layer).
- `databricks__generate_database_name`: prefer `catalog_database`, mirroring fs.
- No dbt-adapters pin bump: `catalog_database` is read via `getattr(config, "catalog_database", None)` and the relation always carries the field, so on dbt-adapters `<1.24.4` (which has no `catalog_database` on the catalog config) it resolves to `None` — a graceful no-op, no crash. It's honored on `dbt-adapters>=1.24.4`.
- Unit tests (integration + macro precedence) + CHANGELOG.

No behavior change for existing projects that don't set `catalog_database`.

**Requirements:** `catalog_database` takes effect only under v2 catalogs — needs **dbt-core ≥ 1.12** (`use_catalogs_v2`, which triggers `bridge_v2_catalog`) and **dbt-adapters ≥ 1.24.4** (bridge surfaces `catalog_database` onto the catalog config). Below that it resolves to `None` (graceful no-op). Note: dbt-core's own `CatalogWriteIntegrationConfig` does not carry `catalog_database` even on 1.13.0a1, so the value flows only through the dbt-adapters v2 bridge — the v1 `write_integrations` path can't carry it.

⚠️ **dbt-core pin:** `pyproject.toml` currently pins `dbt-core <1.11.13`, which **excludes 1.12** — so the feature can't be exercised with the adapter's current pin. Merging as-is lands it dormant/forward-looking; activating it needs a dbt-core pin bump to allow ≥1.12 (separate/maintainer decision).

**Testing:** unit-tested — the `catalog_database` config→integration→relation flow (`test_catalogs_v2.py`, incl. a graceful-degradation case for adapters without the field) and the macro precedence via a rendered `databricks__generate_database_name` (`test_generate_database_name.py`: `catalog_database` > model `database` > `catalog_name` > `target.database`). End-to-end `dbt parse` on a v2-capable dbt-core is not run here (the pin caps below 1.12).

### Checklist

- [ ] [Optional] I have run `/dbt-databricks-pr-ready` (AI agent skill in `.claude/skills/`) and addressed its merge-readiness feedback


dbt/adapters/databricks/catalogs/_unity.py
```diff
@@ -58,4 +63,7 @@ def build_relation(self, model: RelationConfig) -> DatabricksCatalogRelation:
             file_format=parse_model.file_format(model) or self.file_format,
             external_volume=parse_model.location_root(model) or self.external_volume,
             location_path=parse_model.location_path(model),
+            # Set from the catalogs.yml `catalog_database` by the base CatalogIntegration;
```
sd-db: nit: this comment seems redundant (already explained in code)

tests/unit/test_catalogs_v2.py
```diff
@@ -19,9 +19,20 @@ class _Config:
     table_format: Optional[str] = "iceberg"
```
sd-db, line 19: Note for later to add functional tests once support for `dbt-core v1.12` is added in the adapter

dbt/include/databricks/macros/get_custom_name/get_custom_database.sql
```diff
@@ -4,16 +4,18 @@
...
-        {%- endif -%}
+    {%- set default_database = target.database -%}
+    {%- if node is not none and node|attr('database') -%}
+        {%- set catalog_relation = adapter.build_catalog_relation(node) -%}
+    {%- elif 'config' in target -%}
+        {%- set catalog_relation = adapter.build_catalog_relation(target) -%}
+    {%- else -%}
+        {%- set catalog_relation = none -%}
+    {%- endif -%}
+    {%- if catalog_relation is not none and catalog_relation|attr('catalog_database') -%}
```
sd-db, line 15: I think `custom_database_name` should always supersede both `catalog_database`/`catalog_name` as otherwise this might be a surprise for folks we are actually using this to see a silent change

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## dbt-databricks 1.12.3 (TBD)
 
+### Features
+
+- Support `catalog_database` in v2 catalogs.yml for Unity catalogs: routes models to the given physical Unity catalog via `generate_database_name`, taking precedence over the model `database` config and `catalog_name`. Decouples the dbt catalog label from the physical catalog name. Takes effect under v2 catalogs only — requires `dbt-core>=1.12` (adds `use_catalogs_v2`) and `dbt-adapters>=1.24.4` (surfaces `catalog_database` on the catalog config); a no-op otherwise. ([#1590](https://github.com/databricks/dbt-databricks/pull/1590))
```
sd-db: nit: make the changelog entry terse

dbt/include/databricks/macros/get_custom_name/get_custom_database.sql
aahel: This is the expected behaviour according to the v2 spec(docs will be live soon). 
All adapters using catalog_database in v2 catalogs will have this precedence order. So if you have a model having both a database and catalog_name configured it always follow the catalog's database if its configured.
catalog_database is a new field so users adopting v2 catalogs and configuring this will know that its the highest priority.
custom_database_name will still have greater priority than catalog_name so order of existing fields stay same.

Prev order : custom db > catalog name > target db

New order: catalog_db > custom db > catalog name > target db

sd-db: Makes sense, thx for the detailed explanation.

sd-db (approved):

PR #1595, opened 2026-07-17
Fix: Ensure that in incremental strategies columns are backquoted appropriately

### LINKED ISSUE #1594: Incremental strategies generate unquoted column identifiers, breaking non-ASCII column names on the second run with INVALID_IDENTIFIER
### Describe the bug

Incremental models with non-ASCII (e.g. Japanese) column names succeed on the first `dbt build` but fail on every subsequent build when using the `insert_overwrite` strategy (REPLACE ON path). The same applies to the `append` (schema-mismatch path) and `delete+insert` strategies.

The first run works because it compiles to a CTAS that uses the user's model SQL verbatim, where the columns are backtick-quoted by the user (e.g. ``some_value as `あ` ``). On the second run, the adapter itself generates the incremental DML: `get_insert_replace_on_sql` builds the column list from `adapter.get_columns_in_relation(...) | map(attribute="name")` and joins the **raw, unquoted** names into the `INSERT INTO ... REPLACE ON (...) (SELECT col1, col2, ...)` statement. Non-ASCII identifiers are invalid when unquoted, so Databricks rejects the statement:

- Non-ASCII column names (e.g. `あ`) → `[INVALID_IDENTIFIER] The unquoted identifier あ is invalid and must be back quoted as: `あ``
- Column names containing full-width digits (e.g. `１`) → `[PARSE_SYNTAX_ERROR] Syntax error at or near '１'`

Other strategies in the same file already quote correctly (`get_delete_insert_sql` uses `map(attribute='quoted')`, the merge macros use `adapter.quote(...)`), so this is an inconsistency in `get_insert_replace_on_sql` / `insert_into_sql_impl`, plus the config-derived keys (`partition_by` / `liquid_clustered_by` / `unique_key`) interpolated unquoted in the REPLACE ON / DELETE conditions.

### Steps To Reproduce

1. Create an incremental model with backtick-quoted non-ASCII column aliases:

```sql
{{
    config(
        materialized = 'incremental',
        unique_key = ['year', 'period'],
        partition_by = ['year', 'period'],
        incremental_strategy = 'insert_overwrite',
    )
}}

select
    year
    , period
    , value_a as `あ`
    , value_b as `い`
from {{ ref('some_source') }}
```

2. Run `dbt build --select the_model` → succeeds (table is created via CTAS; Databricks/Unity Catalog accepts the quoted Japanese column names).
3. Run `dbt build --select the_model` again with no changes → fails.

Reproduces on a SQL warehouse with the (default-on) `use_replace_on_for_insert_overwrite` behavior flag, and on DBR 17.1+ clusters (which take the REPLACE ON path unconditionally).

### Expected behavior

The second and subsequent incremental runs should succeed.

Second run, generated SQL (note the unquoted identifiers in the SELECT list):

```
[INVALID_IDENTIFIER] The unquoted identifier あ is invalid and must be back quoted as: `あ`.
Unquoted identifiers can only contain ASCII letters ('a' - 'z', 'A' - 'Z'), digits ('0' - '9'), and underbar ('_').
Unquoted identifiers must also not start with a digit.
Different data sources and meta stores may impose additional restrictions on valid identifiers. SQLSTATE: 42602 (line 4, pos 74)

== SQL ==
/* {"app": "dbt", "dbt_version": "2026.7.7+5fba926", "dbt_databricks_version": "1.12.1post1+86d26c9d8f62d0b86c4016711594afe5eadbab13", "databricks_sql_connector_version": "4.1.3", "profile_name": "user", "target_name": "default", "invocation_id": "cbd8559e-8192-4a4d-a03f-57c5b91ff869", "node_id": "my_catalog.my_schema.my_model"} */
insert into table `my_catalog`.`my_schema`.`my_model` AS t
        replace on (t.year <=> s.year AND t.period <=> s.period)
        (select year, period, あ, い from `my_model__dbt_tmp`) AS s
```

### System information

```
dbt Cloud (versionless), core 2026.7.7+5fba926
dbt-databricks: 1.12.1
databricks-sql-connector: 4.1.3
```

**The operating system you're using:** dbt Cloud

**The output of `python --version`:** dbt Cloud managed

### Additional context

- Buggy sites, all in `dbt/include/databricks/macros/materializations/incremental/strategies.sql`:
  - `get_insert_replace_on_sql`: SELECT list built from `map(attribute="name")`, `NULL as <dest_col>` alias, and the `t.<col> <=> s.<col>` REPLACE ON conditions from `partition_by`/`liquid_clustered_by`
  - `insert_into_sql_impl` (append strategy): `common_columns` joined unquoted into `insert into ... (...) select ...`
  - `delete_insert_sql_impl` / `delete_insert_legacy_sql`: `unique_key` interpolated unquoted into the REPLACE ON / DELETE conditions

ash2shukla:
Resolves #1594 

### Description
- Ensures that in incremental strategies columns are backquoted appropriately
- Change the relevant macros to use `adapter.quote`

### Checklist

- [x] [Optional] I have run `/dbt-databricks-pr-ready` (AI agent skill in `.claude/skills/`) and addressed its merge-readiness feedback


dbt/include/databricks/macros/materializations/incremental/strategies.sql
```diff
@@ -182,7 +182,7 @@ replace on ({{ replace_on_expr }})
   {#-- Build WHERE clause for DELETE statement --#}
   {%- set delete_conditions = [] -%}
   {%- for key in unique_keys -%}
-    {%- do delete_conditions.append(target_relation ~ '.' ~ key ~ ' IN (SELECT ' ~ key ~ ' FROM ' ~ source_relation ~ ')') -%}
+    {%- do delete_conditions.append(target_relation ~ '.' ~ adapter.quote(key) ~ ' IN (SELECT ' ~ adapter.quote(key) ~ ' FROM ' ~ source_relation ~ ')') -%}
```
sd-db, line 185: please add a unit test that actually renders `delete_insert_legacy_sql` with a non-ASCII unique_key and asserts quoted DELETE predicates. Current "legacy" tests are hardcoded strings and don't exercise this path.

sd-db (approved):
Changes look goo, have commented on a minor test scope increase. Tests are good as is as well but this will make the coverage complete.

dbt/include/databricks/macros/materializations/incremental/strategies.sql
ash2shukla: Addressed.

PR #1610, opened 2026-07-27
fix: harden credential redaction in logged SQL

### LINKED ISSUE #1609: credential and encryption clauses are not fully redacted in logged SQL
### Describe the bug

`redact_credentials` in `dbt/adapters/databricks/utils.py` is applied to every logged statement
(`connections.py:244` and `connections.py:342`), but the `COPY INTO` pattern it uses misses several
shapes, so secrets can reach `logs/dbt.log` verbatim:

- The lookbehind is the lowercase literal `credential` and the pattern has no `re.IGNORECASE`, so an
  uppercase `CREDENTIAL (...)` clause is not redacted. Uppercase is the form used in the Databricks
  SQL reference.
- The `encryption (...)` clause is not matched at all. `databricks_copy_into` emits that clause from
  the caller supplied `source_encryption` mapping (`macros/adapters/copy_into.sql`), and for AWS
  SSE-C that mapping carries the master key.
- The key pattern is `'\w+'`, so a dotted key such as `'fs.azure.account.key'` is not redacted.
- The value pattern has no `re.DOTALL`, so a value containing a newline is not redacted.
- `search` plus a single slice replacement redacts only the first clause in a statement.

There are also two problems in the same function that are not leaks:

- Splitting the clause body on `,` raises `ValueError: not enough values to unpack (expected 2, got 1)`
  when a value contains a comma. `redact_credentials` is called on line 244, before the `try:` on
  line 246 of the `exception_handler` context manager, so the exception escapes during `__enter__`
  and the statement is never executed.
- `'.*?'` inside a repeated group backtracks exponentially. On `credential (` followed by N
  `'K' = 'V'` pairs with no closing parenthesis: 18 pairs 87 ms, 22 pairs 1,423 ms, 26 pairs
  23,760 ms.

### Steps To Reproduce

```python
from dbt.adapters.databricks.utils import redact_credentials

# 1. uppercase keyword is not redacted
redact_credentials("copy into t WITH (CREDENTIAL ('AWS_SECRET_KEY' = 'SECRET'))")
# -> "copy into t WITH (CREDENTIAL ('AWS_SECRET_KEY' = 'SECRET'))"

# 2. encryption clause is not redacted
redact_credentials("copy into t WITH (encryption ('MASTER_KEY' = 'SECRET'))")
# -> "copy into t WITH (encryption ('MASTER_KEY' = 'SECRET'))"

# 3. a comma in the value raises
redact_credentials("copy into t WITH (credential ('KEY' = 'A,B'))")
# -> ValueError: not enough values to unpack (expected 2, got 1)
```

### Expected behavior

`credential` and `encryption` clauses are redacted regardless of keyword case, every clause in a
statement is redacted, keys that are not `\w` only are handled, and a value containing a comma or a
newline neither leaks nor raises.

Rendering `databricks_copy_into` with `source_encryption` produces the following, and
`redact_credentials` leaves the master key untouched today:

```
copy into target_tbl
    from 's3://bucket/path'
      WITH (
        credential ('AWS_ACCESS_KEY' = 'AKIAEXAMPLE', 'AWS_SECRET_KEY' = 'SUPERSECRETVALUE')
        encryption ('TYPE' = 'AWS_SSE_C', 'MASTER_KEY' = 'MASTERSECRET123')
      )
    fileformat = parquet
```

### System information

```
dbt-databricks 1.12.2 (main @ 2d0fbe9e)
```

**The output of `python --version`:** Python 3.11.15

### Additional context

This is log hygiene rather than a remote exploit, so I am reporting it in the open, which matches how
#1469 handled the credential logging change. I have a fix ready and will open a PR for this issue.

SreeramaYeshwanthGowd:
Resolves #1609

### Description

`redact_credentials` is applied to every logged statement (`connections.py:244` and `:342`), but the
`COPY INTO` pattern behind it misses several shapes, so secrets can reach `logs/dbt.log` verbatim: an
uppercase `CREDENTIAL (...)` clause, the `encryption (...)` clause that `databricks_copy_into` emits
itself, a dotted key such as `'fs.azure.account.key'`, a value containing a newline, and every clause
after the first.

The same function has two problems that are not leaks. Splitting the clause body on `,` raises
`ValueError: not enough values to unpack` when a value contains a comma, and because the call sits on
line 244 before the `try:` on line 246, the exception escapes while the context manager is being
entered, so the statement never runs. Separately, `'.*?'` inside a repeated group backtracks
exponentially on an unterminated clause.

**Fix:** match `credential` and `encryption` case-insensitively, recover the keys with a regex instead
of splitting on `,`, and `sub` every clause. A quoted value is described as
`'(?:[^']|'(?!\s*[,)]))*'`, which treats a quote followed by a delimiter as the closing quote and any
other quote as part of the value. That keeps the match linear and still redacts values containing a
quote. A cheap substring test runs first so the common no-clause path does not pay for the
case-insensitive scan; a 36 KB statement with no clause ends up faster than before, and the
pathological input drops from roughly 24 s to sub-millisecond.

The keyword deliberately has no `\b` prefix. The current lookbehind has no boundary requirement, so
`storage_credential (...)` is redacted today, and adding one would silently narrow coverage. Key names
are preserved and every value in a matched clause is redacted, including a non-secret one such as
`'TYPE' = 'AWS_SSE_C'`, which matches the existing behaviour for `credential (...)`.

**Scope note:** this repairs the existing `COPY INTO` redaction only. Other syntax that can carry
secrets, such as the `OPTIONS (...)` clause, is out of scope here.

**Testing:** `1321 passed, 5 skipped` on the unit suite. The three existing redaction tests are
unchanged and still pass byte for byte, which confirms no behaviour regression. Ten cases are added,
one per shape above plus negative cases proving that `credential_id = 1` and a clause with no
`'key' = 'value'` options are left alone. No functional test is included because log redaction is not
observable against a warehouse.

### Checklist


SreeramaYeshwanthGowd:
Flagging that I have three small fixes open: #1613, #1610 and #1612. Could someone approve the fork workflow runs on #1610 and #1612 so CI can report? #1613 is already green.

If it helps with review order, #1613 is the smallest. #1612 is the one that would benefit most from an `/integration-test` run, ideally against a cluster below DBR 17.1, since that is the branch it changes.

SreeramaYeshwanthGowd:
@sd-db @jprakash-db Merged main and cleared the changelog conflict, so this is ready again. No rush, just flagging in case it helps.

sd-db:
Thanks for working on this. I found one remaining redaction edge case: an escaped apostrophe immediately before `,` or `)` can be mistaken for the end of the secret value. This affects both SQL doubled apostrophes (`''`) and backslash escapes (`\'`).

 I prepared a narrow follow-up in [[602987d](https://github.com/databricks/dbt-databricks/commit/602987d035969920af080cdc196616bcfd8a4e8d)](https://github.com/databricks/dbt-databricks/commit/602987d035969920af080cdc196616bcfd8a4e8d). It keeps redaction best-effort and fail-open.

Could you use or cherry-pick this commit into the PR?

SreeramaYeshwanthGowd:
@sd-db Thanks for the fix. Pulled in 602987d as a fast forward so your authorship stays intact. Full unit suite and pre commit hooks pass locally.

sd-db (approved):

PR #1612, opened 2026-07-27
fix: match the whole key tuple in legacy delete+insert

### LINKED ISSUE #1611: delete+insert with a composite unique_key deletes unmatched rows on DBR below 17.1
### Describe the bug

On the `delete+insert` incremental strategy with a composite `unique_key`, the legacy DBR < 17.1 path
deletes rows that are not key matches, and never re-inserts them. The rows are lost silently.

`dbt/include/databricks/macros/materializations/incremental/strategies.sql` builds one condition per
key column and ANDs them:

```jinja
{%- for key in unique_keys -%}
  {%- do delete_conditions.append(target_relation ~ '.' ~ adapter.quote(key) ~ ' IN (SELECT ' ~ adapter.quote(key) ~ ' FROM ' ~ source_relation ~ ')') -%}
{%- endfor -%}
```

so for `unique_key: ['a', 'b']` the DELETE predicate is the cross product of the two columns' value
sets rather than the set of key tuples:

```sql
delete from target
where target.`a` IN (SELECT `a` FROM source)
  and target.`b` IN (SELECT `b` FROM source)
```

The DBR 17.1+ branch of the same macro is row-wise
(`replace on (target.`a` <=> temp.`a` and target.`b` <=> temp.`b`)`), so the same model with the same
config produces different data depending on the runtime version. dbt-core's cross-adapter default is
also row-wise: `where (unique_key_str) in (select distinct unique_key_str from source)`.

### Steps To Reproduce

1. Model with `incremental_strategy: 'delete+insert'` and `unique_key: ['a', 'b']`, on a cluster
   running DBR earlier than 17.1 so that `replace_on` is unavailable.
2. Target contains `(1, 10)`, `(2, 20)`, `(1, 20)`, `(2, 10)`.
3. An incremental run whose source produces only `(1, 10)` and `(2, 20)`.
4. The DELETE predicate `a IN (1,2) AND b IN (10,20)` matches all four target rows. All four are
   deleted, only the two source rows are inserted, and `(1, 20)` and `(2, 10)` are gone.

### Expected behavior

Only rows whose whole key tuple appears in the source are deleted, matching both the DBR 17.1+ branch
of the same macro and dbt-core's `delete+insert` contract.

### System information

```
dbt-databricks 1.12.2 (main @ 2d0fbe9e)
```

**The output of `python --version`:** Python 3.11.15

### Additional context

Affects all-purpose and job clusters below DBR 17.1, which includes the 14.3, 15.4 and 16.4 LTS
releases. SQL warehouses are unaffected because `replace_on` reports as available there. Single-column
`unique_key` is unaffected. I have a fix ready and will open a PR for this issue.

SreeramaYeshwanthGowd:
Resolves #1611

### Description

On the `delete+insert` strategy with a composite `unique_key`, the legacy DBR < 17.1 path builds one
`IN (SELECT ...)` per key column and ANDs them, so the DELETE predicate is the cross product of the
key columns' value sets rather than the set of key tuples. Rows whose key tuple is absent from the
source are deleted and never re-inserted, silently.

With `unique_key: ['a', 'b']`, a target holding `(1,10) (2,20) (1,20) (2,10)` and a source producing
only `(1,10) (2,20)`, the predicate `a IN (1,2) AND b IN (10,20)` matches all four rows. `(1,20)` and
`(2,10)` are lost.

The DBR 17.1+ branch of the same macro is already row-wise, and dbt-core's cross-adapter default is
`where (unique_key_str) in (select distinct unique_key_str from source)`, so today the same model with
the same config produces different data depending on runtime version.

**Fix:** use a correlated `EXISTS` predicate that compares every key column against the same source
row with null-safe equality.

```sql
delete from target
where exists (
  select 1
  from source
  where target.`a` <=> source.`a`
    and target.`b` <=> source.`b`
)
```

The single-key branch remains byte-for-byte identical, so its behavior is unchanged and remains
covered by the existing single-key and non-ASCII tests.

**Note on the changed tests:** `test_delete_insert_legacy_sql__multiple_unique_keys` previously
asserted the per-column form, so it encoded the bug and had to be updated. Two neighbouring tests that
asserted against their own hardcoded strings are converted to real macro renders, following your
request on #1595.

**Testing:**

- 17 focused unit tests cover composite keys, incremental predicates, and the unchanged single-key
  path.
- A functional regression forces the legacy `DELETE` + `INSERT` branch on live Databricks and
  verifies exact replacements, crossed-tuple preservation, and absent-row preservation.
- The correlated `DELETE` statement was also run directly on DBR 16.4 LTS and produced the expected
  final rows.

A full dbt end-to-end run on compute below DBR 17.1 would add release-matrix confidence, but the
changed predicate and complete dbt execution path are covered by the complementary checks above.

### Checklist


SreeramaYeshwanthGowd:
@sd-db @jprakash-db Merged main and cleared the changelog conflict. Still glad to hear your thoughts on the open question about the DELETE syntax whenever you get a chance.

sd-db (changes requested):
Thanks for putting this together — the underlying bug is real. I reproduced the baseline behavior: the legacy predicate deletes the crossed, unmatched tuples.

I did hit one blocker with the current patch: Delta rejects the row-valued IN in a DELETE condition with [DELTA_UNSUPPORTED_MULTI_COL_IN_PREDICATE] Multi-column In predicates are not supported in the DELETE condition. The new forced-legacy functional test therefore fails on the second dbt run before reaching its result assertion. A correlated EXISTS with one null-safe <=> comparison per key should preserve tuple correlation while using a supported DELETE shape.

When updating the regression test, it would also be useful to include at least one source tuple that already exists in the target with a changed payload. That proves the corrected strategy still deletes/replaces exact matches, rather than only preserving the crossed unmatched rows.

Could you share whether you or another user are currently hitting this row-loss behavior in an active workload — and, if so, which DBR version and how blocking it is — or whether this was found proactively? That context would help gauge the urgency. If it is actively impacting workloads, I am happy to chip in on the macro and regression-test changes as well.

SreeramaYeshwanthGowd:
@sd-db Thanks for catching this. Swapped the row valued IN for a correlated EXISTS with null safe equality per key, matched on the full tuple. Tests updated with a real key match case, to confirm true matches still get replaced and not just the false positive cases.

Also ran the actual statement directly on a DBR 16.4 LTS single node cluster against a composite key table. The DELETE succeeded with no error, and only the true tuple match was removed, the unrelated row stayed. So the internal Delta flag does not block this on Databricks Runtime, at least on 16.4 LTS.

Raw output from that run:
```
DELETE_OK= True
DELETE_ERR= None
AFTER_DELETE_ROWS= [Row(id=2, color='red', msg='goodbye')]
FINAL_ROWS= [Row(id=1, color='blue', msg='replaced'), Row(id=1, color='red', msg='updated'), Row(id=2, color='blue', msg='updated'), Row(id=2, color='red', msg='goodbye')]
```

sd-db (approved):
Changes look good, thx !!

PR #1613, opened 2026-07-27
fix: escape single quotes in relation comments at create time

### LINKED ISSUE #1251: persist_docs fails if model description contains an apostrophe
The macro implementation at https://github.com/databricks/dbt-databricks/blob/8cda62ee19d01e0670e3156e652841e3ffd3ed41/dbt/include/databricks/macros/relations/components/comment.sql#L1-L5 which is used to generate some relation comment clauses when `persist_docs.relation` is true, does not properly escape single quotes.

SreeramaYeshwanthGowd:
Resolves #1251

### Description

`get_create_sql_comment` interpolates a model description straight into a SQL string literal without
escaping, so a description containing an apostrophe produces an unterminated literal and the
statement fails to parse:

```
COMMENT 'Bob's model'
```

Every other place the adapter emits a comment already escapes: `relations/comment.sql`,
`adapters/persist_docs.sql` (three sites), `relations/view/create.sql` and
`materializations/seeds/helpers.sql`. This one was missed, so it is the only raw interpolation left
in the macro tree.

The macro is reached from three call sites, all of which pass a relation-config value straight
through: `relations/materialized_view/create.sql`, `relations/streaming_table/create.sql` and
`relations/streaming_table/alter.sql`. So a materialized view or streaming table with
`persist_docs: {relation: true}` and an apostrophe in its description cannot be built.

**Fix:** apply the same `replace` filter the sibling macros already use. The escaping happens only
at render time, and `CommentProcessor` continues to compare the raw project value against the raw
server value, so this introduces no configuration drift on reruns.

**Scope note:** I kept the fix identical to the existing convention rather than centralising the
filter into a shared helper. Consolidating the six sites is a reasonable follow-up but would turn a
one-line correctness fix into a six-file refactor.

**Testing:** four macro tests cover a plain comment, one apostrophe, several apostrophes, and the
`none` case that keeps the existing `comment is string` guard covered. The two escaping tests fail
before the change and pass after. For server-observable proof, `materialized_view_schema` now carries
a model description containing an apostrophe and
`test_create_materialized_view_with_comment_and_constraints` asserts the comment round-trips through
`information_schema.tables`; that test errors before this fix.

### Checklist


SreeramaYeshwanthGowd:
@sd-db @jprakash-db Merged main so this is conflict free again. It was fully green on CI before, happy to wait for another run.

dbt/include/databricks/macros/relations/components/comment.sql
```diff
@@ -1,5 +1,5 @@
 {%- macro get_create_sql_comment(comment) -%}
 {% if comment is string -%}
-  COMMENT '{{ comment }}'
+  COMMENT '{{ comment | replace("'", "\\'") }}'
```
sd-db: Could we escape existing backslashes before escaping apostrophes here? Databricks treats backslashes as string escapes. I verified that the rendered `Bob\\'s view` fails to parse, while `C:\temp\'s view` stores a tab instead of the literal `\t`. Please add regression cases for both inputs.

sd-db (changes requested):
The reported apostrophe case is fixed and well tested. I’m only blocking on the backslash edge case above, which can cause either a parse failure or a corrupted stored comment.

SreeramaYeshwanthGowd:
@sd-db Good catch. Backslashes are now escaped before apostrophes, so an existing backslash cannot merge with the escape or be misread by Databricks. Added your two cases plus a bare backslash case, which failed the same way before. All pass.

sd-db (approved):

PR #1615, opened 2026-07-28
fix: warn on missing persist_docs columns for view/materialized_view/streaming_table create

### LINKED ISSUE #1399: Materialization v2 with update_via_alter for nonexistent columns
### Describe the bug
When you use materialization_v2 with views with update_via_alter it attempts to push all comments from yml to databricks regardless if they exist.

### Steps To Reproduce
1. Create a model (view). In relevant .yml add a column which does not exist in .sql
2. Run the model (creation)
3. Run the model (again)

### Expected behavior
During both run the nonexistent columns are skiped when transfering their comments to DBX.

### System information
We use dbt-databricks 1.10.14, dt-core 1.10.13 and tested it for dbt 1.11 as well.

### Additional context
Might be solved by https://github.com/databricks/dbt-databricks/pull/1358?

r-jais:
### Description

resolves #1399 

Follow-up to #1563. That PR added the "warn when a documented column is absent from the relation" check as a **post-build validation** (`validate_persist_doc_columns`, mirroring the shared `validate_doc_columns` behavior every other adapter uses) and wired it into the **table** and **incremental** materializations.

This PR completes the create-time coverage by calling the same gated, post-build validation on the materializations #1563 didn't touch:

- **view** create (V1 + V2) — `dbt/include/databricks/macros/materializations/view.sql`
- **materialized view** create/replace/refresh — `.../materialized_view.sql`
- **streaming table** create/replace/refresh — `.../streaming_table.sql`

Each is a single `validate_persist_doc_columns(target_relation, model)` call after the relation is built. The macro is gated on `config.persist_column_docs()` and applies no comments, so it never fires when column persistence is off and stays safe under `--warn-error` (per the caution in #1563 about not warning unconditionally in `parse_columns_and_constraints`). Previously these paths iterated only the query's output columns (`get_persist_docs_column_list`) or built inline comments via `parse_columns_and_constraints`, so a YAML-only documented column was silently dropped.

### Stacking

> **This PR is stacked on #1563** and currently shows its commits too. It targets `main`; once #1563 merges, this diff reduces to just the view/materialized-view/streaming-table changes. Review/merge #1563 first.

### Tests

Functional (`tests/functional/adapter/persist_docs/test_persist_docs.py`): create-time missing-column warning for **view**, **materialized view**, and **streaming table** (each: documented column absent from the relation → warns exactly once, gated on `persist_docs.columns`). New fixtures in `tests/functional/adapter/persist_docs/fixtures.py`. The core `validate_persist_doc_columns` logic is unit-tested in #1563.

### Checklist

- [x] This PR includes tests


sd-db:
Can we rebase on the latest code for #1563 ?

tests/functional/adapter/persist_docs/test_persist_docs.py
```diff
@@ -740,3 +745,90 @@ def test_new_documented_column_is_not_warned_before_alter_view(self, project, ad
...
+        }
+
+    def test_view_create_warns(self, project):
+        util.run_dbt(["seed"])
+        _, logs = util.run_dbt_and_capture(["run"])
+        assert "column_that_does_not_exist" in logs
+        assert logs.count(_MISSING_COLUMN_WARNING) == 1
+
+
+class TestPersistDocsColumnMissingWarnsMaterializedViewCreate:
```
sd-db, line 919: Could we add the standard DLT/profile markers to this class and the streaming-table class below?

```python
@pytest.mark.dlt
@pytest.mark.skip_profile("databricks_cluster", "databricks_uc_cluster")
```

Both tests are currently collected under the all-purpose-cluster profiles, and the materialized-view test begins executing instead of being skipped. The existing materialized-view and streaming-table suites use these markers to keep these cases on the supported SQL-warehouse profile.

tests/functional/adapter/persist_docs/test_persist_docs.py
```diff
@@ -740,3 +745,90 @@ def test_new_documented_column_is_not_warned_before_alter_view(self, project, ad
...
+        return {"schema.yml": override_fixtures.missing_column_view_schema}
+
+    @pytest.fixture(scope="class")
+    def project_config_update(self):
+        return {
+            "flags": {"use_materialization_v2": True},
+            "models": {"test": {"+persist_docs": {"relation": False, "columns": True}}},
+        }
+
+    def test_view_create_warns(self, project):
```
sd-db: Could we avoid using captured log substrings as the functional assertion here and in the MV/ST cases below?

The behavior to pin is that the warning participates in `--warn-error` while the relation is still built correctly. Please use the existing `--warn-error-options '{"error": ["JinjaLogWarning"]}'` with `expect_pass=False`, then inspect `DESCRIBE EXTENDED` or `information_schema` to verify the real columns and present-column comment. Exact warning text or count belongs in a unit/macro test.

tests/functional/adapter/persist_docs/test_persist_docs.py
```diff
@@ -740,3 +745,90 @@ def test_new_documented_column_is_not_warned_before_alter_view(self, project, ad
         comments = {column.column: column.comment for column in columns}
         assert comments["id"] == "updated id comment"
         assert comments["added_col"] == "added column comment"
+
+
+class TestPersistDocsColumnMissingWarnsViewCreate:
```
sd-db, line 745: Could we extend this coverage beyond the V2 create path?

In particular, #1399 reports the V2 `view_update_via_alter` rerun, while the production validation call also covers a separate V1 branch. Please add:

- a first-run/changed-query rerun with `view_update_via_alter=true` and a YAML-only missing column;
- a V1 view case;
- a `persist_docs.columns: false` gate case.

The rerun should use `--warn-error` and then verify the altered columns/comments through `DESCRIBE EXTENDED`. A representative subsequent MV/ST operation would also pin that their validation remains in the shared execute path rather than becoming create-only.

tests/functional/adapter/persist_docs/test_persist_docs.py
r-jais: Noted, added to both classes.

tests/functional/adapter/persist_docs/test_persist_docs.py
r-jais: Ack...switched to --warn-error-options + expect_pass=False, & verifying via DESCRIBE EXTENDED/information_schema.columns

tests/functional/adapter/persist_docs/test_persist_docs.py
r-jais: added the V1 view case, the persist_docs.columns: false gate case, the V2 alter-path rerun with a YAML-only missing column, and a second run on MV/ST to cover the REFRESH path too.

sd-db:
Overall looks good, will approve and merge once the tests complete, thx again for the PR !!

sd-db (approved):
Changes look good, thx for the PR!

PR #1620, opened 2026-07-29
feat: add wiring for telemetry v1

saishreeeee:
Description:
Adds an opt-in, default-off telemetry subsystem for the Databricks adapter. Enable with `enable_dbt_telemetry: true` in `connection_parameters`.

- Two events per eligible invocation (`build`, `run`, `test`, `seed`, and `snapshot` only): `POST_PARSE` contains invocation configuration, manifest resource counts, connection shape, and behavior flags; `POST_RUN` contains status, termination reason, node and hook result counts, and duration.
- Privacy: no credentials, hostnames, cluster IDs, or object names. Auth is reduced to an enum family, `http_path` to compute type, and only a non-PII workspace ID is sent.
- Best-effort and async: sends run on a background thread with a 10-second request timeout and never block a SQL connection. Run-end waits up to two request timeouts for queued events. Failures are ignored and never affect the dbt command.

Testing:
(will check whether telemetry logs land in the dbt telemetry table)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## dbt-databricks next
 
+### Features
+
+- Add opt-in invocation telemetry for eligible commands via `connection_parameters.enable_dbt_telemetry` ([#1620](https://github.com/databricks/dbt-databricks/pull/1620))
+
 ### Fixes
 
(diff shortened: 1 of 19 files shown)


saishreeeee:
split into #1647 and #1648

PR #1625, opened 2026-08-06
docs: gate new @available methods behind portable alternatives

sd-db:
## Summary

Gate new `@available` methods behind portable alternatives in the agent guide.

Each `@available` method is permanent public API — Jinja is late-bound, so nothing flags an unused one — and must be re-implemented in the Fusion engine's Rust adapter before behavior matches across engines. Shared-contract macros get that for free.

`### Adding New Adapter Method` now opens with that cost, then lists options to work down, stopping at the first that works: a shared-contract macro, plain Jinja, a relation config component, and only then a new `@available` method. Reaching the last step asks for a justification, portable types, and an acknowledgement that Fusion parity is owed.

Shorter entries added to `### Code Patterns to Follow` and `### Common Pitfalls for Agents`.

## Test plan

Docs only — no runtime change, no CHANGELOG entry, no tests. `pre-commit run --all-files` passes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -388,11 +388,34 @@ Models can be configured with Databricks-specific options:
 
 ### Adding New Adapter Method
 
-1. Add method to `DatabricksAdapter` class in `impl.py`
-2. Implement database interaction logic
-3. Add corresponding macro if SQL generation needed
-4. Write unit tests with mocked database calls
-5. Write functional tests with real database
+Every `@available` method is permanent public API. Jinja is late-bound, so nothing
+warns you a method is unused, and each one must be independently re-implemented in the
+Fusion engine's Rust adapter — both its dispatch and its semantic classification —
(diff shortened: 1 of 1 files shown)


sd-db:
@xuliangs FYI

jprakash-db (approved):

AGENTS.md
```diff
@@ -497,6 +524,7 @@ Models can be configured with Databricks-specific options:
 9. **Use capability system for version checks** - Never add new `compare_dbr_version()` calls
 10. **Remember per-compute caching** - Different clusters may have different capabilities in the same run
 11. **Multi-statement SQL**: Don't use semicolons to separate statements - return a list instead and let `execute_multiple_statements()` handle it
+12. **Don't add an `@available` method for logic Jinja can already do** - each one is permanent API and owes a matching Fusion implementation. Check the shared adapter contract first.
```
xuliangs, line 527: Nice!

PR #1630, opened 2026-08-11
chore: raise dbt-common upper bound to <1.39.0

itsnamangoyal:
## Summary
- Raises the `dbt-common` upper bound from `<1.38.0` to `<1.39.0`, unlocking `dbt-common==1.38.0`.
- `dbt-common==1.38.0` includes `UncoercedBoolSerialization`, which fixes a `mashumaro>=3.15` regression where the bool deserializer does a naive truthy cast (`bool("false") -> True`) instead of raising on invalid input. That regression currently causes several config-validation checks (invalid `enabled`, malformed exposure/metric/source config) to silently accept bad values instead of raising.
- Ran `uv lock --upgrade-package dbt-common` to bump the pinned lock version alongside the bound, per the contributing guide.

## Test plan
- [x] `uv lock --check` passes (lock and pyproject are consistent)
- [ ] CI unit/functional suites


sd-db (approved):

Issue #1607, opened 2026-07-23
Add rows_affected and other build metadata to run_results `adapter_response`


tweavers:
### Describe the feature
Some dbt adapters include `rows_affected` in the adapter response after building a model via [run_results](https://docs.getdbt.com/reference/artifacts/run-results-json?version=2.0&name=Fusion). This is incredibly helpful from a monitoring perspective and something that we leveraged up until recently coming from the dbt-supported Postgres adapter.

### Describe alternatives you've considered
Alternative would be to develop a custom post-hook to store this data separately. This would be nowhere near as performant as the cached query proposed in the earlier PR, and would also be very difficult to row counts for incremental models

### Additional context
This was work previously in-flight in late 2024 / early 2025 but was never merged after a design change. Closed PR linked below for reference
https://github.com/databricks/dbt-databricks/pull/883

### Who will this benefit?
* Users of dbt artifacts to monitor pipeline jobs

PR #1632, opened 2026-08-11
feat: Report rows affected in Databricks adapter responses

### LINKED ISSUE #351: rows_affected not returned by adapter

cjcdoomed:
## Summary
- populate Databricks adapter responses with `rows_affected` from cursor rowcount when available, and include the row count in the adapter response message (for example, `OK 3`)
- stop follow-up tag/constraint/column-mask/row-filter/column-comment statements (and `create_table_at`'s schema-only CREATE, and `safe_relation_replace`'s backup drop) from silently overwriting the real DML statement's `rows_affected` by naming them explicitly instead of `'main'`
- de-duplicate the "last statement is `main`" logic into a single `execute_multiple_statements` helper
- add unit tests for known/unknown rowcount behavior and response-ownership regression tests across the affected macros
- add a functional incremental test to validate second-run `rows_affected` reporting

## Testing
- `hatch run code-quality` (ruff, ruff format, mypy, uv-lock) passes
- `hatch run unit` passes: 1362 passed, 4 skipped
- functional suite (`cluster-e2e-dev`) has not been run against a live workspace from this environment

## Related
- Resolves #351 (rows_affected not returned by adapter)
- Related to #1607 (Add rows_affected and other build metadata to run_results adapter_response) — this PR covers `rows_affected` only, not the other build metadata requested there
- Prior closed PR: #1614 [DRAFT] feat: expose rows_affected in adapter_response
- Prior closed PR: #883 Add affected, inserted, updated, deleted row to DatabricksAdapterResponse


dbt/adapters/databricks/handle.py
```diff
@@ -90,12 +90,22 @@ class DatabricksAdapterResponse(AdapterResponse):
 
     @classmethod
     def from_cursor(cls, cursor: Any) -> "DatabricksAdapterResponse":
+        rows_affected = cls._get_rows_affected(cursor)
+        message = "OK" if rows_affected is None else f"OK {rows_affected}"
         return cls(
-            _message="OK",
+            _message=message,
+            rows_affected=rows_affected,
```
sd-db, line 97: Reading `rowcount` here works, but the response can be overwritten by a later statement. I reproduced this with V1 tag changes, V2 create/replacement, and tagged snapshots: the DML returned a count, but `run_results.json` kept the later DDL response with `rows_affected: null`.

Could we preserve the data-changing response as `main` while follow-up statements use separate names?

tests/functional/adapter/incremental/test_incremental_run_result.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_incremental_second_run_reports_rows_affected(self, project):
+        util.run_dbt(["run", "--select", "incremental_run_result"])
+
+        results = util.run_dbt(["run", "--select", "incremental_run_result"])
+        assert len(results) == 1
+
+        rows_affected, message = _rows_and_message(results[0].adapter_response)
+        if rows_affected is None:
+            pytest.skip("Connector did not report rowcount for this incremental run")
```
sd-db: This skip makes the test pass when the feature is missing. Reverting the implementation would produce a skip instead of a failure.

Since the test environment uses connector 4.4.0, could we assert `rows_affected == 2`, message `OK 2`, and the final three table rows? Please also cover a metadata follow-up, such as tags, to catch response overwrites.

tests/functional/adapter/incremental/test_incremental_run_result.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+from dbt.tests import util
+
+from tests.functional.adapter.fixtures import RerunSafeMixin
+
+incremental_run_result_sql = """
```
sd-db: Consider moving this model SQL into `fixtures.py` to match the functional-test layout used elsewhere.

tests/unit/test_handle.py
```diff
@@ -657,20 +657,36 @@ class TestDatabricksAdapterResponse:
...
+
+    @patch.dict(os.environ, {}, clear=True)
+    def test_from_cursor__unknown_rowcount(self):
+        cursor = Mock()
+        cursor.query_id = "q2"
+        cursor.rowcount = -1
+        resp = DatabricksAdapterResponse.from_cursor(cursor)
+        assert resp.query_id == "q2"
+        assert resp.rows_affected is None
+        assert str(resp) == "OK"
```
sd-db, line 692: Consider adding cases for `rowcount == 0`, a missing `rowcount`, and a `None` cursor value.

sd-db (changes requested):
I ran these on sample projects locally and seeing that follow-up statements still overwrite valid counts, and the functional test skips when the count is missing. Separately, please document whether multi-statement strategies should return the main DML count or an aggregate.

dbt/adapters/databricks/handle.py
cjcdoomed: Fixed in the latest push. Non-DML statements now get unique names: statements that loop (e.g. multi-statement incremental strategies) use `'main'` for the first and `'statement_N'` for subsequent iterations; the tag, tblproperties, and column_tags macros use `'apply_tags'`, `'apply_tblproperties'`, and `'apply_column_tags'` respectively. The DML cursor is always captured under `'main'` and is no longer overwritten by follow-up DDL.

tests/functional/adapter/incremental/test_incremental_run_result.py
cjcdoomed: Updated in the latest push — though we landed in a different place than you asked for, so flagging for discussion.

We restored `pytest.skip` when `rows_affected is None` and loosened to `assert rows_affected > 0`, because MERGE rowcount semantics differ across DBR versions (some report matched+inserted, others only updated rows). Happy to tighten back to `== 2` if you're comfortable pinning to connector 4.4.0 behaviour.

We removed the tag assertion: the overwrite regression is now covered at the source (follow-up statements no longer use `'main'`), and tag propagation has its own test in `test_incremental_tags.py`. If you'd prefer to keep the tag check here specifically as a regression sentinel for the overwrite fix, we can add it back — just note the test now also carries `@pytest.mark.skip_profile("databricks_cluster")` since `system.information_schema.table_tags` is UC-only.

sd-db: Tests env is pinned to 4.4.0 - https://github.com/databricks/dbt-databricks/blob/main/uv.lock#L488

We should look to make this strict

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Redact all `credential` and `encryption` clauses in logged SQL, regardless of keyword case (thanks @SreeramaYeshwanthGowd!) ([#1610](https://github.com/databricks/dbt-databricks/pull/1610) resolves [#1609](https://github.com/databricks/dbt-databricks/issues/1609))
 - Stop `delete+insert` with a composite `unique_key` from deleting unmatched rows on DBR below 17.1 (thanks @SreeramaYeshwanthGowd!) ([#1612](https://github.com/databricks/dbt-databricks/pull/1612) resolves [#1611](https://github.com/databricks/dbt-databricks/issues/1611))
 - Escape single quotes in relation comments so materialized views and streaming tables with an apostrophe in the description can be created (thanks @SreeramaYeshwanthGowd!) ([#1613](https://github.com/databricks/dbt-databricks/pull/1613) resolves [#1251](https://github.com/databricks/dbt-databricks/issues/1251))
+- Report `rows_affected` from Databricks cursor rowcount in adapter responses so batch incremental runs display updated row counts. ([#1632](https://github.com/databricks/dbt-databricks/pull/1632))
```
sd-db: nit: can you move the changelog entry to the correct release ?

sd-db: Also we should maybe add this as a feature instead of a fix

sd-db (commented):
Hi @cjcdoomed I am still reviewing the latest changes, but I would recommend to rebase on top of `1.13.latest`. This PR contains significant changes to the generated artifacts and better to have it as a minor release instead of a patch one

dbt/adapters/databricks/handle.py
sd-db: > use 'main' for the first and 'statement_N' for subsequent iterations

I still have my doubts on the naming scheme here as the first statement is not necessarily the `main` one and `statement_N` is a bit vague

> the tag, tblproperties, and column_tags macros use 'apply_tags', 'apply_tblproperties', and 'apply_column_tags' respectively

This looks good, thx !

sd-db:
@cjcdoomed I added focused functional repros for the remaining response-ownership cases here: [diff](https://github.com/databricks/dbt-databricks/compare/23443c9bebca4643a5ee466cdf16df767354dc63...6d23c9d34e1698ddc2824a9e9bc768b549d0bbff). All four tests confirm the warehouse state first, then fail because the result returns plain OK without rows_affected.

- V2 create: expected OK 2, received plain OK.
- V2 safe replacement: expected OK 2, received plain OK.
- V1 constraint change: expected OK 1, received plain OK.
- V1 row-filter change: expected OK 1, received plain OK.

dbt/include/databricks/macros/etc/statement.sql
```diff
@@ -23,7 +23,7 @@
     {% endcall %}
   {%- else %}
     {%- for sql in statements %}
-      {% call statement(name="main") %}
+      {% call statement(name="main" if loop.first else "statement_" ~ loop.index) %}
```
sd-db: `main` is not just a label here: dbt serializes it as the model’s adapter response. Making `loop.first` the `main` result means list order now defines the user-visible `rows_affected`. For `delete+insert`, DELETE wins and INSERT is stored as `statement_2`.

That is a behavior change, not only a naming choice. Please define which count is authoritative, or aggregate the counts, and use names that describe the statements rather than `main` / `statement_N`. The same positional rule is duplicated in both incremental paths.

sd-db (changes requested):
Let's fix the naming and the observed repro issues

cjcdoomed:
Rebased this branch onto `1.13.latest` per your suggestion, and pushed the review feedback fixes on top:

- Multi-statement strategies now tag the **last** statement as `'main'` (not the first), so `delete+insert` reports the INSERT's rowcount instead of the DELETE's.
- Tightened the functional test to `assert rows_affected == 2` and dropped the `pytest.skip` fallback, since the test env is pinned to connector 4.4.0.
- Moved the changelog entry under `## dbt-databricks next` and reclassified it as a Feature.

dbt/include/databricks/macros/etc/statement.sql
cjcdoomed: Traced this further using the four repro tests you posted — the same clobbering pattern (a follow-up DDL/ALTER statement reusing `'main'` after the real DML) turned out to exist in five more places, not just this loop:

- `apply_row_filter` and `apply_constraints` were both naming their ALTER statements `'main'`, directly overwriting the incremental merge's real count.
- `create_table_at` had it backwards: the schema-only `CREATE TABLE` was named `'main'` while the row-writing `INSERT ... BY NAME` was named `'merge into target'`.
- `safe_relation_replace`'s backup-table `DROP` was also named `'main'`, clobbering the count the `create_table_at` fix had just corrected.
- `apply_column_masks` and `alter_column_comments` had the identical bug, caught in a final holistic review rather than by a repro, since they share the exact same call site (`apply_config_changeset`) as `apply_row_filter`/`apply_constraints`.

On the specific question here: rather than leave the multi-statement selection purely positional-and-undocumented, I consolidated the three copies of the "last statement is \`main\`" logic (this macro plus two inline copies in `incremental.sql`) into this one macro, and added a comment defining the rule explicitly: multi-statement strategies build preparatory statements first and their real data-writing statement last, so the last statement is authoritative. That's now also written into `AGENTS.md` so it doesn't regress again.

I stopped short of inventing per-statement descriptive names for the positional case (e.g. delete+insert's two statements) since there's exactly one real multi-statement strategy today and a fuller naming contract felt speculative — happy to revisit if you'd rather have that now.

All four of your repro tests are merged as permanent regression coverage in `tests/functional/adapter/response_ownership/`.

cjcdoomed:
Pushed a follow-up series addressing your Aug 20 review (rebased on `1.13.latest`, no conflicts):

- `apply_row_filter` and `apply_constraints`: renamed their ALTER statements off `'main'` (matches the pattern you already approved for `apply_tags`/`apply_tblproperties`/`apply_column_tags`) — fixes your V1 constraint and V1 row-filter repros.
- `create_table_at`: swapped which statement owns `'main'` — it was the schema-only `CREATE TABLE`, now it's the row-writing `INSERT ... BY NAME` — fixes your V2 create repro.
- `safe_relation_replace`: renamed its backup-table `DROP` off `'main'`, which was clobbering the count the `create_table_at` fix had just corrected — fixes your V2 safe-replacement repro.
- Consolidated the three duplicated copies of the "last statement is \`main\`" positional logic into `execute_multiple_statements`, with a comment defining the rule (replied inline on that thread with more detail).
- Added your four repro tests verbatim as permanent regression coverage under `tests/functional/adapter/response_ownership/`.
- Found two more instances of the same bug in a final pass (`apply_column_masks`, `alter_column_comments` — same call site as the row-filter/constraints fixes) and fixed those too, not caught by your repros but the identical pattern.
- Documented the naming convention in `AGENTS.md` so this class of bug doesn't reappear.
- Broadened the changelog entry to reflect the fix now covers table materialization, not just incremental.

All commits are separated by concern if it's easier to review incrementally. Let me know if you'd rather I use descriptive per-statement names for the delete+insert two-statement case instead of the positional last-is-\`main\` rule — flagged my reasoning for not doing that inline on the thread.

PR #1638, opened 2026-08-13
fix: silence constraint not enforced

### LINKED ISSUE #1196: Can not globally ignore ConstraintNotEnforced
### Describe the bug
dbt-core has the option to silence warnings in the project flags. dbt-databricks seems to ignore these (or some).

### Steps To Reproduce

- Create a table with a PK
- set flags:
`flags:
  warn_error_options:
    silence: [ ConstraintNotEnforced ]`
- Warning is displayed: unenforced constraint type: primary_key

### Expected behavior
I did not expect to see this warning.

However settings `warn_unenforced: true` on the specific PK makes the warning go away. But by this method you need to set this for every PK-contraint.

`09:09:06  unenforced constraint type: primary_key`

### System information
```
Core:
  - installed: 1.10.11
  - latest:    1.10.13 - Update available!

Plugins:
  - databricks: 1.10.12 - Up to date!
  - spark:      1.9.3   - Up to date!
```

Windows 11

Python 3.12.10

saishreeeee:
Resolves #1196 

### Description
Emit the structured ConstraintNotEnforced event instead of a free-form exceptions.warn on the V1 constraint path, so warn_error_options (silence/warn/error) can target the unenforced primary_key/foreign_key warning — matching the V2 path. 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Fixes
 
-- Honor `warn_error_options` (silence/warn/error) for the unenforced `primary_key`/`foreign_key` constraint warning when `use_materialization_v2: false`, by emitting the structured `ConstraintNotEnforced` event instead of a free-form warning (thanks @frankivo!) ([#XXXX](https://github.com/databricks/dbt-databricks/pull/XXXX) resolves [#1196](https://github.com/databricks/dbt-databricks/issues/1196))
+- Make the unenforced `primary_key`/`foreign_key` constraint warning respect `warn_error_options` when `use_materialization_v2: false` ([#1638](https://github.com/databricks/dbt-databricks/pull/1638) resolves [#1196](https://github.com/databricks/dbt-databricks/issues/1196))
 
 ## dbt-databricks 1.12.4 (Aug 12, 2026)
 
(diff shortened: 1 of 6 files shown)


sd-db:
Can you check if we can simplify the jinja <-> python interactions here --> http://github.com/databricks/dbt-databricks/blob/main/AGENTS.md#adding-new-adapter-method

EDIT: I see from the comment that this was considered, can we check if we can make this similar to v2 flow instead of introducing a new method as the overall vision is to unify in the long run.

saishreeeee:
> Can you check if we can simplify the jinja <-> python interactions here --> [http://github.com/databricks/dbt-databricks/blob/main/AGENTS.md#adding-new-adapter-method](https://github.com/databricks/dbt-databricks/blob/main/AGENTS.md#adding-new-adapter-method)
> 
> EDIT: I see from the comment that this was considered, can we check if we can make this similar to v2 flow instead of introducing a new method as the overall vision is to unify in the long run.

Reusing the ConstraintNotEnforced event. Reusing V2's process_constraint isn't as easy: in V2 the warning is a side effect of that step, which also renders the SQL:
- V2: parse_columns_and_constraints → relation.enrich → process_constraint (validates+warns and produces the DDL)
- V1: builds the DDL itself in Jinja

So there's no way to borrow just the warning from process_constraint without also taking its SQL rendering which means changing V1's whole constraint flow.

PR #1645, opened 2026-08-21
feat: support persist_constraints in v2 materializations

saishreeeee:
### Description

Stacked on #1653.

- Allow v2 table creation and relation-config reconciliation to read legacy `meta.constraints` when `persist_constraints` is enabled.
- Match v1 persist behavior by skipping unsupported constraints and generating stable names for unnamed constraints.
- Keep contract-only constraint handling unchanged.

### Testing

- Unit tests for v2 relation-config parsing, stable naming, and string config values
- Functional `persist_constraints` test on a UC SQL warehouse


sd-db:
Closing this PR as we `persist_constraints` is legacy and we don't want to support this in v2. Later we can think on this post unification(when that time comes); but for now we should drop this approach.

PR #1646, opened 2026-08-22
perf: skip constraint catalog lookups when the model contract is not enforced

### LINKED ISSUE #1641: Excessive information_schema queries from constraint-existence checks on models with primary_key set but contract not enforced
### Describe the bug

For models that declare a model-level `primary_key` property but do **not** have `contract.enforced: true` and do **not** have an explicit `constraints:` block, dbt-databricks still runs a full "does this constraint already exist?" lookup against `information_schema` on every run, for every such model.

This generates 2 queries per model with a `primary_key` set (against `information_schema.key_column_usage` and `information_schema.table_constraints`), plus an additional query against `information_schema.referential_constraints` for any foreign keys. No constraint is actually being applied or enforced for these models — the checks appear to run purely because the property is set.

In one project we investigated: 255 of 303 models (84%) have `primary_key` set, with `contract.enforced: false` and no `constraints` block anywhere in the project, yet every run issues the constraint-existence check for all 255 of them. In a warehouse shared across many projects, this pattern (where present) contributes disproportionately to `information_schema` query volume — enough, in our case, to help trip Unity Catalog's per-metastore rate limit (`RESOURCE_EXHAUSTED.UC_INFORMATION_SCHEMA_PER_METASTORE_DB_RATE_LIMIT_EXCEEDED`) and fail otherwise-unrelated jobs sharing that metastore. (We checked a second project in the same account with no `primary_key` usage at all, confirming this is tied to modeling convention/project, not something universal — but it's still a real cost wherever the convention is used.)

### Steps To Reproduce

1. Create a model with a model-level `primary_key` property set, but do not enable `contract.enforced` and do not define an explicit `constraints:` block, e.g.:
```yaml
models:
  - name: my_model
    primary_key: [id]
    columns:
      - name: id
```
2. Run `dbt build` (or `dbt run`) with debug logging (`--debug`) against a Databricks/Unity Catalog target.
3. Observe in the debug log that dbt-databricks issues a query similar to:
```sql
SELECT kcu.constraint_name, kcu.column_name
FROM `<catalog>`.information_schema.key_column_usage kcu
WHERE kcu.table_catalog = '...'
  AND kcu.table_schema = '...'
  AND kcu.table_name = 'my_model'
  AND kcu.constraint_name = (
    SELECT constraint_name
    FROM `<catalog>`.information_schema.table_constraints
    WHERE table_catalog = '...'
      AND table_schema = '...'
      AND table_name = 'my_model'
      AND constraint_type = 'PRIMARY KEY'
  )
ORDER BY kcu.ordinal_position;
```
4. Repeat for every model with `primary_key` set — query count scales linearly with the number of such models, regardless of whether any constraint is actually meant to be applied.

### Expected behavior

Constraint-existence checks (and any subsequent DDL) should only run for models where a constraint is actually going to be applied/enforced — i.e., where `contract.enforced: true` and/or an explicit `constraints:` block is present. A model-level `primary_key` property used purely for documentation/lineage purposes should not trigger live `information_schema` verification queries against the warehouse on every run.

This looks related to #1342, which described constraints being applied without an enforced contract as a regression and was closed as fixed — but the behavior described here reproduces on `dbt-databricks==1.12.1`, suggesting either the fix didn't cover the `primary_key`-property-only case (no `constraints:` block at all), or this is a related but distinct code path.

Separately, #845 requests batching per-column constraint/comment operations into fewer queries — applying that kind of batching to the constraint-existence check described here would also directly mitigate the volume problem even if the "don't check at all when not enforced" fix isn't feasible short-term.

### System information

```
Plugins:
  - databricks: 1.12.1
```

Environment: dbt Cloud, Databricks/Unity Catalog target, multi-project account.

### Additional context

Observed via Databricks warehouse-side error `RESOURCE_EXHAUSTED.UC_INFORMATION_SCHEMA_PER_METASTORE_DB_RATE_LIMIT_EXCEEDED`, which surfaced the underlying query volume. Table/schema/catalog identifiers have been genericized here to avoid sharing customer-identifying information.

TangoEnSkai:
Resolves #1641

### Description

**Context**

`IncrementalTableAPI._describe_relation` unconditionally issues three
`information_schema` lookups per relation (`fetch_non_null_constraint_columns`,
`fetch_primary_key_constraints`, `fetch_foreign_key_constraints`) to build the
server side of the constraint diff. Both consumers of that diff already refuse
to act on it unless the contract is enforced:

- `materializations/incremental/incremental.sql:211` — `{% if constraints and contract_config and contract_config.enforced ... %}`
- `relations/table/alter.sql:33` — `{% if constraints and contract_config and contract_config.enforced %}`

and `ConstraintsProcessor.from_relation_config` already returns an empty config
when the contract is not enforced (#1342). So for a model without
`contract.enforced: true` those lookups are computed and then discarded. In the
reported project 255 of 303 models set `primary_key` for documentation only,
which made this a meaningful share of that account's `information_schema`
traffic — enough to contribute to
`UC_INFORMATION_SCHEMA_PER_METASTORE_DB_RATE_LIMIT_EXCEEDED` on a shared
metastore.

**What**

- `ConstraintsConfig` carries `contract_enforced` and implements
  `requires_server_metadata_for_diff()`, the same opt-out hook `TagsConfig` and
  `ColumnTagsConfig` already use.
- `ConstraintsProcessor.from_relation_config` sets it to `False` on the
  not-enforced path it already short-circuits.
- `IncrementalTableAPI._describe_relation` skips the three constraint fetches
  when the model config says server metadata cannot affect the diff, mirroring
  the existing tag-fetch gating (including `model_config is None` still
  fetching, for backward compatibility).

**Why this is behavior-preserving**

With the contract off, the model side of the diff is empty and the server side
is now empty too, so `get_diff` returns `None` instead of an
`unset_constraints` set that both materializations were already dropping on the
floor. Nothing that reaches a `statement()` call changes. The
`DESCRIBE AS JSON` path is untouched: it derives constraints from a single
metadata call that is needed for column masks and row filters anyway, so there
is nothing to save there.

The `contract_enforced=True` default means every other construction site —
including `from_relation_results` — keeps today's behavior. When the contract
*is* enforced, the lookups still run, so removing a constraint from an enforced
model still produces the `ALTER ... DROP CONSTRAINT`.

**Testing**

- `hatch run unit` equivalent: 1348 passed, 4 skipped.
- `pre-commit run --all-files`: all hooks pass.
- New unit tests in `tests/unit/test_adapter.py` cover the three
  `_describe_relation` cases (contract off / contract on / `model_config` is
  `None`); the "contract off" test fails against `main` and passes with this
  change.
- New unit tests in `tests/unit/relation_configs/test_constraint.py` cover
  `requires_server_metadata_for_diff()` from both the model and results sides.
- No functional test: the change is a query-planning optimization with no
  server-observable effect, and per `docs/testing.md` functional tests must
  assert server-observable state. The user-visible behavior it must not break —
  constraints not applied without a contract — is already covered by
  `TestPrimaryKeyNotAppliedWithoutContractV1/V2` in
  `tests/functional/adapter/incremental/test_incremental_constraints.py`
  (the #1342 regression tests), which exercise exactly this path. I do not have
  a Databricks workspace, so I have not run the functional suite.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - Replace an existing table or view with a metric view using backup-and-create instead of `CREATE OR REPLACE VIEW ... WITH METRICS` ([#1640](https://github.com/databricks/dbt-databricks/pull/1640) resolves [#1639](https://github.com/databricks/dbt-databricks/issues/1639))
 - Interpolate lazily-formatted `databricks.sql` log records when mirroring them into dbt logs ([#1642](https://github.com/databricks/dbt-databricks/pull/1642) resolves [#1637](https://github.com/databricks/dbt-databricks/issues/1637))
 
+### Under the Hood
+
+- Skip the `information_schema` constraint metadata fetches for models whose contract is not enforced, since the result cannot be applied. (thanks @TangoEnSkai!) ([#1646](https://github.com/databricks/dbt-databricks/pull/1646) resolves [#1641](https://github.com/databricks/dbt-databricks/issues/1641))
+
 ## dbt-databricks 1.12.4 (Aug 12, 2026)
 
 ### Fixes
--- a/dbt/adapters/databricks/impl.py
+++ b/dbt/adapters/databricks/impl.py
@@ -81,6 +81,7 @@ from dbt.adapters.databricks.relation_configs.column_tags import (
     ColumnTagsConfig,
     ColumnTagsProcessor,
 )
+from dbt.adapters.databricks.relation_configs.constraints import ConstraintsProcessor
 from dbt.adapters.databricks.relation_configs.incremental import IncrementalTableConfig
 from dbt.adapters.databricks.relation_configs.materialized_view import (
     MaterializedViewConfig,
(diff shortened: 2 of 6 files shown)


sd-db:
Hi @TangoEnSkai https://github.com/databricks/dbt-databricks/pull/1643 already solves this and is merged to `1.13.latest`, I am looking to close this PR as this is already done.

TangoEnSkai:
Thanks for the heads-up @sd-db — I checked #1643 and it indeed covers the same code path (and more thoroughly). Glad to see it shipping in 1.13.0. Closing side is all good from my end.

PR #1647, opened 2026-08-22
feat: add opt-in POST_PARSE adapter telemetry

saishreeeee:
## Description
- Adds default-off, opt-in `POST_PARSE` telemetry for `build`, `run`, `test`, `seed`, and `snapshot`.
-  Events contain aggregate invocation, manifest, compute/auth, and behavior-flag data.
- Sending is asynchronous and best-effort, requires a successful SQL connection, and never affects dbt execution.
- Stacked follow-up: **POST_RUN** (#1648).

## Testing
Ran dbt build with telemetry opted in and manually verified the POST_PARSE event landed in the dbt telemetry table

PR #1648, opened 2026-08-22
feat: add POST_RUN outcome telemetry

saishreeeee:
## Description
  - Adds opt-in POST_RUN outcome telemetry on top of #1647.
  - Captures invocation status, termination reason, duration, and node/hook counts.
  - Handles fail-fast and ephemeral-node result reconciliation.
  - Flushes pending telemetry safely during cleanup.

image

## Testing
Ran dbt build with telemetry opted in and manually verified the POST_PARSE and POST_RUN events landed in the dbt telemetry table


sd-db:
I realised one gap `POST_RUN` sends run on a daemon thread, so the process can exit before delivery finishes. `cleanup_connections()` does not wait for it.

I have added a proposed fix here that performs a bounded flush after `CommandCompleted`. A timeout only warns, preserving best-effort telemetry:

- [Branch](https://github.com/databricks/dbt-databricks/tree/sd-db/codex/pr-1648-terminal-flush)
- [Commit](https://github.com/databricks/dbt-databricks/commit/659785abd04813d4bf973e650e9ab2a2eafc39af)

dbt/adapters/databricks/telemetry/coordinator.py
```diff
@@ -89,25 +131,158 @@ def mark_start(self, invocation_id: str) -> None:
...
+            elif observed_ephemeral_count == len(state.ephemeral_ids):
+                partial_selected_count = state.expected_result_count + observed_ephemeral_count
+            else:
+                partial_selected_count = None
+            return (
+                top_level_results + state.hook_results,
+                partial_selected_count,
+                state.expected_result_count,
+                False,
+                False,
```
sd-db: Could we preserve observed partial aggregates here? When `EndRunResult` is missing due to an interrupt or exception, `node_results` and hook results may still contain valid observations, but this `False` causes the builder to omit them.

Please return `state.results_captured` while keeping `coverage_complete=False`. This produces:

- observed results: `result_aggregates_available=true`, `expected_result_coverage_complete=false`
- no observed results: `result_aggregates_available=false`, coverage unset

Partial counts are lower bounds; consumers can filter for complete runs using the coverage flag. Please add tests for both cases.

dbt/adapters/databricks/telemetry/coordinator.py
```diff
@@ -89,25 +131,158 @@ def mark_start(self, invocation_id: str) -> None:
...
+                return [], 0, 0, False, False
+            observed_ids = {str(uid) for uid, _ in state.node_results}
+            observed_ephemeral_count = len(observed_ids.intersection(state.ephemeral_ids))
+            top_level_results = [
+                result for result in state.node_results if str(result[0]) not in state.ephemeral_ids
+            ]
+            if state.end_run_statuses is not None:
+                # EndRunResult statuses have no unique IDs.
+                remaining = Counter(_status_key(status) for status in state.end_run_statuses)
+                typed_results = top_level_results + state.hook_results
```
sd-db: Fail-fast can change a provisional `NodeFinished(error)` into a terminal `EndRunResult(skipped)`. Because this reconciles by status, the provisional error remains while the unmatched skipped status may be synthesized or treated as an ephemeral selection, inflating both status and selected counts.

Please treat the terminal `EndRunResult` status population as authoritative and use earlier node events only for resource-type attribution. Add regressions for the `error → skipped` transition with both `threads=1` and repeated concurrent fail-fast runs.

sd-db (approved):

PR #1663, opened 2026-09-02
fix: replace misleading log message when skipping ephemeral clones

ajhlee-dbt:
This is a port of the v2 fix for this issue raised in https://github.com/dbt-labs/dbt-core/issues/14474.

### Description

`dbt clone` logged `No relation found in state manifest for <model>` for ephemeral models, which is misleading — ephemeral models are CTEs that are never materialized, so having no relation in the state manifest is expected and correct behavior.

This PR replaces that message with `Skipping clone for ephemeral model <model>`, making it clear the model is intentionally skipped rather than missing. 

### Checklist


sd-db (approved):
Changes look good, thx !

PR #1674, opened 2026-09-09
fix: replace instead of drop when rebuilding a managed Iceberg table

### LINKED ISSUE #1662: Incremental full refresh on UC managed Iceberg drops the table instead of create-or-replace (v1 materialization)
### Describe the bug

Running `dbt build --full-refresh` on an **incremental** model that materializes as a **Unity Catalog managed Iceberg table** (`use_managed_iceberg: true` behavior flag + `table_format: iceberg`, no `use_materialization_v2`) executes `drop table if exists` followed by `create or replace table ... as select`, instead of a single atomic `create or replace`.

Because the drop happens first, the table does not exist for the entire duration of the CTAS — several minutes on large tables. Concurrent queries fail with `TABLE_OR_VIEW_NOT_FOUND`, and the table's history/time travel is lost. Grants survive only when they are inherited from the schema.

This looks like the v1-materialization / managed-Iceberg sibling of #1205 (v2 + managed Delta, fixed by #1210): the same "is this relation replaceable?" check fails, just for a different reason.

### Root cause (from reading the shipped macros)

In `macros/materializations/incremental/incremental.sql` (v1 path):

```jinja
{% set is_delta = (catalog_relation.file_format == 'delta' and existing_relation.is_delta) %}
{% set is_iceberg = (catalog_relation.file_format == 'iceberg' and existing_relation.is_iceberg) %}
{% set is_replaceable_format = is_delta or is_iceberg %}
```

With managed Iceberg, `catalog_relation.file_format` is always `'delta'`: `DatabricksCatalogRelation` defaults `file_format` to delta, and `iceberg_table_properties` even raises `DbtConfigError` if the user sets anything else ("When table_format is 'iceberg', cannot set file_format to other than delta."). Meanwhile the existing relation's metadata reports `Provider = iceberg`, so `existing_relation.is_delta` is false.

Result: `is_delta` is false (existing table is not delta) **and** `is_iceberg` is false (`file_format` is delta) → `is_replaceable_format = false` → the macro takes the drop-then-create branch, even though `CREATE OR REPLACE TABLE` works fine on managed Iceberg (it is exactly the statement the macro runs right after the drop).

The v2 path has the same shape (`is_replaceable = existing_relation.can_be_replaced and is_replaceable_format`), so it should be affected too.

### Steps To Reproduce

`dbt_project.yml`:

```yaml
flags:
  use_managed_iceberg: true
models:
  my_project:
    +table_format: iceberg
```

Model:

```sql
{{ config(materialized='incremental', incremental_strategy='merge', unique_key='id') }}
select 1 as id
```

1. `dbt run` (creates the managed Iceberg table)
2. `dbt run --full-refresh --log-level debug`
3. The debug log shows `Applying DROP to: ...` followed by `drop table if exists ...`, then the `create or replace table ... using iceberg ... as select`.

### Expected behavior

A single `create or replace table ... as select`, with no drop — same as delta-on-delta full refreshes — so the table stays continuously queryable and keeps its identity.

Sanitized debug log from a production-sized run (dbt-core 1.11.11, dbt-databricks 1.12.1, serverless SQL warehouse, UC managed Iceberg table that was itself created by dbt with this same config):

```
16:24:00.449339 [debug] MATERIALIZING INCREMENTAL
16:24:00.526970 [debug] Applying DROP to: `catalog`.`gold`.`my_incremental_model`
16:24:00.5      [debug] On model.my_project.my_incremental_model:
                        drop table if exists `catalog`.`gold`.`my_incremental_model`
                        ... followed by:
                        create or replace table `catalog`.`gold`.`my_incremental_model`
                        using iceberg ... as select ...
16:27:40.321475 [info ] 1 of 16 OK created sql incremental model gold.my_incremental_model [OK in 220.08s]
```

The table did not exist between 16:24:00 and 16:27:40 (~3m40s); queries against it during that window failed with TABLE_OR_VIEW_NOT_FOUND.

### System information

```
Core:
  - installed: 1.11.11
Plugins:
  - databricks: 1.12.1
```
(Changelog for 1.12.2–1.12.5 does not mention this area, so latest appears affected as well.)

**The output of `python --version`:** Python 3.13.14

### Additional context

- Related: #1205 (same replaceability check failing for managed tables on the v2 path with Delta; fixed by #1210). This report is the v1 path with managed Iceberg.
- Workaround we are using for planned full refreshes: compile the model and run the generated `create or replace table ... as select` manually, skipping the drop.

cemsbr:
Resolves #1662

### Description

Stacked on #1669 — the first four commits are that PR's; this PR adds one commit on top and will be rebased onto `main` once #1669 merges.

#1662 reports the drop-then-create on `--full-refresh` of an **incremental** model on a Unity Catalog managed Iceberg table. The **`table`** materialization has the same gate and hits it on **every** run:

```jinja
not (existing_relation.can_be_replaced and adapter.resolve_file_format(config) in ('delta', 'iceberg'))
```

With `use_managed_iceberg` on and `table_format: iceberg`, `resolve_file_format` returns `parquet`, so the existing table is always dropped before a `create or replace table ... using iceberg` CTAS that would have been atomic on its own. The table is unavailable for the whole CTAS (minutes on large tables) and concurrent readers fail with `TABLE_OR_VIEW_NOT_FOUND`; table history/time travel restarts on every run.

This change reuses `format_allows_create_or_replace(catalog_relation, existing_relation)` from #1669 in both the v1 and v2 branches of `table.sql`. The `is_shallow_clone` and `type != 'table'` arms are unchanged, and Delta targets keep the same behavior as before (`catalog_relation.file_format == 'delta'` and `existing_relation.is_delta`).

Not included: `materializations/seeds/helpers.sql` (`databricks__reset_csv_table`) has the same gate; happy to fold it in here or in a follow-up, whichever you prefer.

### Behavior change

When the existing table's provider differs from the target (an existing managed Iceberg table with a Delta target, e.g. a project turning `use_managed_iceberg` back off), the old gate attempted `create or replace table` across providers; Databricks rejects that with `[MANAGED_ICEBERG_OPERATION_NOT_SUPPORTED] Managed Iceberg tables do not support REPLACE with different providers.` and the rebuild failed. With the shared predicate that case drops and recreates the table instead, matching #1669's `test_delta_target_on_iceberg_relation`. Probe results in the comments below. With the provider-aware predicate from #1669 (`1024e315`), the reverse case — a managed-Iceberg target over a table the project still has as Delta — also drops and recreates on the next `table` run instead of failing; covered by `TestManagedIcebergTableOverExistingDelta`.

### Checklist

- [x] I have run this code in development and it appears to resolve the stated issue — as a root-project override of the `table` materialization with the same predicate, on dbt-databricks 1.12.5 against a UC managed Iceberg table (second run no longer emits `Applying DROP`; `information_schema.tables.created` is preserved)
- [x] This PR includes tests, or tests are not required/relevant for this PR — `TestManagedIcebergTableRebuild` in `tests/functional/adapter/iceberg/test_iceberg_support.py`, mirroring #1669's `TestManagedIcebergFullRefresh`; needs a workspace, so it relies on `/integration-test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## dbt-databricks 1.12.6 (TBD)
 
+### Fixes
+
+- Use `create or replace table` instead of dropping the table first when a full refresh rebuilds an incremental model on a Unity Catalog managed Iceberg table ([#1669](https://github.com/databricks/dbt-databricks/pull/1669) resolves [#1662](https://github.com/databricks/dbt-databricks/issues/1662))
+
 ### Under the Hood
 
 - Emit only changed `databricks_tags` keys in `ALTER … SET TAGS` ([#1667](https://github.com/databricks/dbt-databricks/pull/1667))
--- a/dbt/include/databricks/macros/materializations/incremental/incremental.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/incremental.sql
@@ -11,9 +11,7 @@
   {% set partition_by = config.get('partition_by') %}
   {% set language = model['language'] %}
   {% set on_schema_change = incremental_validate_on_schema_change(config.get('on_schema_change'), default='ignore') %}
-  {% set is_delta = (catalog_relation.file_format == 'delta' and existing_relation.is_delta) %}
-  {% set is_iceberg = (catalog_relation.file_format == 'iceberg' and existing_relation.is_iceberg) %}
-  {% set is_replaceable_format = is_delta or is_iceberg %}
+  {% set is_replaceable_format = format_allows_create_or_replace(catalog_relation, existing_relation) %}
   {% set compiled_code = adapter.clean_sql(model['compiled_code']) %}
 
   {% if adapter.get_behavior_flag_no_warn('use_materialization_v2') %}
--- a/dbt/include/databricks/macros/materializations/incremental/replaceable_format.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/replaceable_format.sql
@@ -6,15 +6,24 @@
      could never be true. The target is Iceberg exactly when `table_format` is iceberg and the
      behavior flag is on -- the same condition `file_format_clause` uses to emit `using iceberg`.
      The flag is read without warning because this runs for every incremental model, including
-     projects that never opt in (issue #1266). --#}
+     projects that never opt in (issue #1266).
+
+     The two arms are mutually exclusive because `create or replace` cannot change a table's
+     provider: Databricks rejects it with `MANAGED_ICEBERG_OPERATION_NOT_SUPPORTED` and leaves the
+     table as it was. A managed-Iceberg target over a legacy Delta table -- a project that has just
+     switched the flag on -- must therefore drop and recreate, even though `file_format` still reads
+     delta for it. --#}
 {% macro format_allows_create_or_replace(catalog_relation, existing_relation) %}
   {%- set target_is_managed_iceberg = (
         catalog_relation.table_format == 'iceberg'
         and adapter.get_behavior_flag_no_warn('use_managed_iceberg')
       ) -%}
-  {%- set replaceable = (
-        (catalog_relation.file_format == 'delta' and existing_relation.is_delta is true)
-        or (target_is_managed_iceberg and existing_relation.is_iceberg is true)
-      ) -%}
+  {%- if target_is_managed_iceberg -%}
+    {%- set replaceable = existing_relation.is_iceberg is true -%}
+  {%- else -%}
+    {%- set replaceable = (
+          catalog_relation.file_format == 'delta' and existing_relation.is_delta is true
+        ) -%}
+  {%- endif -%}
   {{ return(replaceable) }}
 {% endmacro %}
--- a/dbt/include/databricks/macros/materializations/incremental/replaceable_format.sql
+++ b/dbt/include/databricks/macros/materializations/incremental/replaceable_format.sql
@@ -1,0 +1,20 @@
+{#-- True when `create or replace table` can stand in for drop-then-create on a full refresh.
+
+     Managed Iceberg needs its own arm rather than reusing `file_format`: an Iceberg model keeps
+     `file_format` at delta (`iceberg_table_properties` raises for anything else) and `iceberg` is
+     not an accepted `file_format` at all, so the `file_format == 'iceberg'` test this replaced
+     could never be true. The target is Iceberg exactly when `table_format` is iceberg and the
+     behavior flag is on -- the same condition `file_format_clause` uses to emit `using iceberg`.
+     The flag is read without warning because this runs for every incremental model, including
+     projects that never opt in (issue #1266). --#}
+{% macro format_allows_create_or_replace(catalog_relation, existing_relation) %}
+  {%- set target_is_managed_iceberg = (
+        catalog_relation.table_format == 'iceberg'
+        and adapter.get_behavior_flag_no_warn('use_managed_iceberg')
+      ) -%}
+  {%- set replaceable = (
+        (catalog_relation.file_format == 'delta' and existing_relation.is_delta is true)
+        or (target_is_managed_iceberg and existing_relation.is_iceberg is true)
+      ) -%}
+  {{ return(replaceable) }}
+{% endmacro %}
--- a/tests/functional/adapter/iceberg/test_iceberg_support.py
+++ b/tests/functional/adapter/iceberg/test_iceberg_support.py
@@ -206,3 +206,29 @@ class TestManagedIcebergFullRefresh(ManagedIcebergMixin):
         assert get_version_zero_timestamp(project, "iceberg_full_refresh") == created, (
             "history restarted, so the full refresh dropped and recreated the table"
         )
+
+
+@pytest.mark.skip_profile("databricks_cluster")
+class TestManagedIcebergOverExistingDelta(ManagedIcebergMixin):
+    """Switching `use_managed_iceberg` on over tables a project already has as Delta must drop and
+    recreate them. `create or replace` cannot change a table's provider, so replacing here fails
+    with MANAGED_ICEBERG_OPERATION_NOT_SUPPORTED and leaves the table Delta (issue #1662)."""
+
+    @pytest.fixture(scope="class")
+    def models(self):
+        return {"iceberg_over_delta.sql": fixtures.incremental_iceberg_base}
+
+    def test_full_refresh_converts_the_delta_table(self, project):
+        project.run_sql(
+            "create or replace table {database}.{schema}.iceberg_over_delta using delta "
+            "as select 1 as id, 'initial' as status"
+        )
+        assert get_provider(project, "iceberg_over_delta") == "delta"
+
+        util.run_dbt(["run", "--full-refresh"])
+
+        assert get_provider(project, "iceberg_over_delta") == "iceberg"
+        rows = project.run_sql(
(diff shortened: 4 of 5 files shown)


moomindani:
Thanks for picking this up, and for reusing the macro and test shape from #1669.

I ran your branch here against a serverless SQL warehouse, since fork PRs can't start CI on their own: `code-quality` clean, unit 1349 passed / 4 skipped, `tests/functional/adapter/iceberg/` 12 passed. Reverting only `table.sql` makes `TestManagedIcebergTableRebuild` fail, so the new test does catch it. I also checked the v1 path is fine to stop dropping — `databricks__create_table_as` already emits `create or replace` for a managed Iceberg model.

One question. The old gate looked only at the target format, and `format_allows_create_or_replace` also requires the existing relation's provider to match, so an existing Iceberg table with a Delta target — a project turning `use_managed_iceberg` back off — moves from replace to drop. Dropping seems the safer of the two, but was that intentional?

Since this is stacked on #1669, that one needs to land first; it's waiting on a maintainer to approve its workflows.

cemsbr:
Thanks for running it end to end, and for the question — it made me check rather than assume.

Yes, intentional. Reusing the predicate unchanged keeps `table` and `incremental` behaving the same on a rebuild, and the "provider changed → drop" arm is the one #1669 already encodes in `test_delta_target_on_iceberg_relation`. But I did not know what the old gate's REPLACE actually did across providers, so I probed it on a Unity Catalog serverless SQL warehouse (dbt-databricks 1.12.5):

| existing table | statement | result |
|---|---|---|
| `using iceberg` (managed) | `create or replace table t using delta as select ...` | `[MANAGED_ICEBERG_OPERATION_NOT_SUPPORTED] Managed Iceberg tables do not support REPLACE with different providers.` — table left intact, still `Provider = iceberg` |
| `using delta` | `create or replace table t using iceberg as select ...` | same error — table left intact, still `Provider = delta` |

So in the case you describe (existing Iceberg table, project turns `use_managed_iceberg` back off), the old gate did not replace: the rebuild failed with that error until someone dropped the table by hand. Dropping first is the only path that works, so I would keep the predicate as it is for this PR.

One thing the probe surfaced about the predicate itself, which is #1669's territory and your call: the Delta arm (`catalog_relation.file_format == 'delta' and existing_relation.is_delta`) is also true for a managed-Iceberg target over a legacy Delta table (`test_managed_iceberg_target_on_delta_relation` → replace). Per the second row above that REPLACE fails with the same error, so a project flipping the flag *on* over existing Delta tables would hit it on the first full refresh (and, with this PR, on the first `table` rebuild). Making that arm symmetric — `existing_relation.is_delta` only when the target is not managed Iceberg — would turn that into a drop-then-create as well. Happy to send that as a follow-up on top of whichever lands first, or to fold it here if you prefer.

I've added a short "Behavior change" note to the PR description.

moomindani:
Your read was right, and thank you for probing it rather than assuming — I had the same two rows before pushing, so we agree on the facts:

| existing | statement | result |
|---|---|---|
| `using delta` | `create or replace ... using iceberg` | `MANAGED_ICEBERG_OPERATION_NOT_SUPPORTED`, table left as delta |
| `using iceberg` | `create or replace ... using delta` | same, table left as iceberg |

On your two points:

**The "provider changed → drop" arm is intentional — agreed, keep it.** Your probe settles what I was actually unsure about: the old gate never replaced successfully in that case, it just failed until someone dropped the table by hand. So this is not a behavior change worth hedging over; it turns an error into a working rebuild.

**The asymmetric Delta arm was a real bug, and I've folded the fix into #1669** rather than take you up on the follow-up — the predicate lives there, and leaving a known-broken arm in the commit that introduces it didn't seem right. I'd written the symmetric version first and then talked myself out of it on "don't change behavior beyond the reported bug", which was the wrong call: the behavior I preserved was one I hadn't verified.

`1024e315` on #1669 makes the arms mutually exclusive, flips `test_managed_iceberg_target_on_delta_relation` to `False` with the reason in the docstring, and adds a functional test (`TestManagedIcebergOverExistingDelta`) that creates the table as Delta, turns the flag on, and runs `--full-refresh`. On the old predicate that test reproduces your second row as a dbt error; on the new one it passes with `Provider = iceberg`. Unit 1357 passed, `code-quality` clean, `tests/functional/adapter/iceberg/` 12 passed. The commit message credits you for the probe.

I also rebased #1669 onto main, so the stack needs redoing — but `table.sql` shouldn't need any edit, since the corrected predicate is the same shared macro your change already calls. Your `table` path picks up the flag-on-over-Delta case for free.

cemsbr:
Thanks for folding the symmetric arm into #1669 — agreed that's where it belongs.

Restacked on `1024e315`. As you expected, `table.sql` needed no change: the diff is still the `catalog_relation` set plus the two gates calling `format_allows_create_or_replace`. The only conflict was both PRs appending a test class at the end of `test_iceberg_support.py`; kept both, yours first.

I also added `TestManagedIcebergTableOverExistingDelta` (separate `test:` commit), the `table` counterpart of your `TestManagedIcebergOverExistingDelta`: create the table as Delta, flag on, plain `dbt run`, assert `Provider = iceberg` and one row. It makes the "picks it up for free" claim for the `table` path something CI checks rather than something we say.

Local: unit 1357 passed / 4 skipped, `pre-commit` clean on the touched files. The two functional tests need the workspace, so they rely on `/integration-test` like the rest.

moomindani:
Restack verified. I ran the branch against a serverless SQL warehouse: `tests/functional/adapter/iceberg/` 14 passed, unit 1357 passed / 4 skipped, `code-quality` clean. I also read `resolve_file_format` (`impl.py:1164`) to confirm the gate you replaced: it returns `parquet` for a managed Iceberg model and never returns `iceberg`, so the old `in ('delta', 'iceberg')` test was false for exactly the case this PR fixes, and its `iceberg` arm was dead code the same way the incremental one was. Adding `build_catalog_relation` to `table.sql` is no new risk either — `relations/table/create.sql` already calls it on the same models.

One correction on what your new test does, worth recording so nobody prunes it later. I checked both directions:

| tree | `TestManagedIcebergTableRebuild` | `TestManagedIcebergTableOverExistingDelta` |
|---|---|---|
| this PR | pass | pass |
| `table.sql` reverted to pre-#1674 | **fail** (history restarted) | pass |
| `table.sql` from this PR, predicate reverted to pre-`1024e315` | — | **fail** (`MANAGED_ICEBERG_OPERATION_NOT_SUPPORTED`) |

`TestManagedIcebergTableOverExistingDelta` passes on the pre-#1674 baseline: the old gate dropped in that case too, because `resolve_file_format` returned `parquet`, and dropping is what this case needs. So it isn't a reproduction of the original bug — it's a guard on the composition, and it earns its place by failing if the predicate ever loses the symmetric arm. Which is the more useful thing for it to guard, since that arm is the part someone could plausibly "simplify" later. Might be worth a line in the docstring saying so.

Both PRs are still waiting on a maintainer to approve workflows — neither has any check reported yet.

cemsbr:
Thanks for running both directions — your matrix is right, and my docstring wasn't. `TestManagedIcebergTableOverExistingDelta` passes on the pre-#1674 baseline because the old gate also dropped there (`resolve_file_format` → `parquet`), so it isn't a reproduction of the original bug; `TestManagedIcebergTableRebuild` is. What it guards is the composition: with this PR's `table.sql`, it fails with `MANAGED_ICEBERG_OPERATION_NOT_SUPPORTED` if the predicate ever loses the symmetric arm from `1024e315`.

Docstring updated to say exactly that (amended into the `test:` commit, head `29f63ff3`; `pre-commit` clean). Nothing else pending on my side until a maintainer approves the workflows on both PRs.

PR #1681, opened 2026-09-14
fix: short-circuit column doc probe query when model.columns is empty (#1680)

### LINKED ISSUE #1680: persist_docs.columns triggers a full-plan probe query on views even when the model has no column docs
**Describe the bug**

With `persist_docs: {columns: true}`, `databricks__create_view_as` always calls `get_columns_in_query(sql)` before creating a view, which executes `select * from (<full model sql>) as __dbt_sbq where false limit 0` against the warehouse. It does this even when the model has no `columns:` entries in any schema YAML, in which case the resulting column list contains no `comment` clauses and is semantically identical to emitting no column list at all.

On a wide view the probe hangs for a long time. `where false` is only folded after full analysis, so Spark still plans the whole query and does Delta metadata/file listing on every source.

**Expected behavior**

No probe query when the model has no column descriptions. dbt's own `macros/adapters/persist_docs.sql` already uses this guard:

```jinja
{% if for_columns and config.persist_column_docs() and model.columns %}
```

The Databricks view path omits the `and model.columns` short-circuit.

**Suggested fix**

In `dbt/include/databricks/macros/relations/view/create.sql`:

```jinja
{%- if config.persist_column_docs() and model.columns -%}
```

**Steps to reproduce**

1. Set `persist_docs: {relation: true, columns: true}` in `dbt_project.yml`
2. Create a `materialized='view'` model with no corresponding `schema.yml` entry
3. Run `dbt run -s that_model` and observe the `where false limit 0` query in the Databricks query history

**Notes**

- Workaround: `persist_docs={'relation': True, 'columns': False}` per model.

**System information**

- dbt-databricks: 1.11.6
- dbt-core: 1.11.6
- databricks-sql-connector: 4.1.3

DelorneN007:
fix: short-circuit column doc probe query when model.columns is empty (#1680)

### Problem
* When `persist_column_docs()` is enabled on a view without column documentation, `get_columns_in_query(sql)` triggers an unnecessary probe query (`select * from ... limit 0`).
* This can cause performance overhead or hangs on wide views that lack explicit column definitions.

### Solution
* Added an `and model.columns` guard to the `config.persist_column_docs()` condition inside `databricks__create_view_as`.
* Prevents invoking the probe query when `model.columns` is empty or undefined.

Resolves #1680


DelorneN007:
Hi @jprakash-db, @saishreeeee, and @sd-db,

I’ve opened this PR to resolve #1680 by adding an and model.columns guard to databricks__create_view_as so we skip the unnecessary select * from ... limit 0 probe query when a model has no column documentation. 

Whenever you have a chance, I'd appreciate your review and any guidance on triggering the CI workflows. Thank you!

PhilippLange:
as I mentioned this could also be explored (short circuit not necessary):
https://github.com/databricks/dbt-databricks/issues/1680#issuecomment-5663624625

DelorneN007:
Hi @PhilippLange, thanks for the feedback! Could you share a bit more context on the alternative approach you had in mind? I’m happy to pivot or refine this PR to align with your suggestion.

PhilippLange:
glad to elaborate @DelorneN007:

**1. Stop inlining the column list at creation.** In `databricks__create_view_as`, the `(col comment '...')` clause is the only reason `get_columns_in_query(sql)` runs. It's needed because the DDL has to name the columns before the view exists. Drop that clause and the probe disappears for *every* view - not just the empty-columns case the current guard covers.

**2. Reapply the comments after creation, from the catalog.** Same block `databricks__alter_view` already uses after `ALTER VIEW ... AS`:

```jinja
{%- set existing_columns = adapter.get_columns_in_relation(relation) -%}
{%- set columns_to_persist = adapter.get_persist_doc_columns(existing_columns, model.columns) -%}
{{ alter_column_comment(relation, columns_to_persist) }}
```

Wrap it in a new macro (e.g. `persist_view_column_comments(relation)`) next to the create macro, and call it from each place in `dbt/include/databricks/macros/materializations/view.sql` that creates or replaces the view.

**3. Mind the runtimes where this isn't available.** `alter_column_comment` resolves to `COMMENT ON COLUMN`, which is the only statement that sets a column comment on an existing view. Two cases can't use it:

- DBR < 16.1, where `comment_on_column_sql` falls back to `ALTER TABLE ... ALTER COLUMN`, invalid against a view. Gate on `adapter.has_dbr_capability('comment_on_column')`.
- Hive metastore — same fallback problem. Gate on `not relation.is_hive_metastore()`.

3) has yet to be solved.

